Issue #40, opened 2015-04-14
Title in panel doesn't update when closing single window in frame


The-Compiler:
When there's a frame with a single window and it's closed, the title of that window persists in the panel.

`window_title_changed` is not emitted, and the panel doesn't update the title on `tag_flags` because that'd be too much noise.

We probably should emit `window_title_changed` without arguments in that situation.

original mail:

```
Simplest case: a tag, one frame and an xterm in it. I am running panel.sh,
to display tags, window title, date and time. For xterm, the window title
is the shell prompt string and current directory. If you simply exit from
the shell, say with ctrl-d, the window title remains in the panel. Switch
to another tag and then back to the original tag, and the title gets
updated correctly. I note that panel.sh does look for the
window_title_changed event, so I'm guessing that the window manager itself
is failing to generate the event in this case, but I haven't researched
this carefully so it's just a guess.

Note that if you have two xterms in the frame with different current
directories and thus different window titles, if you exit from one, the
window title in the panel gets updated correctly.

This is not an xterm-specific problem -- I'm just using xterm as a simple
way to reproduce this. If you have just a browser in a frame and close it,
the same problem occurs.

I'm running herbstluftwm 0.6.2-2 on an up-to-date 64-bit Arch Linux system.
```

discussion:

```
21:56 <The-Compiler> thorsten`: what do you think about the bug on the ML? Should hlwm emit title_changed when a window gets closed and none is active? Should panel.sh update the title on tag_flags?
22:00 <thorsten`> The-Compiler: Either window_title_changed or a new window_closed (which is only emitted if the focused window is closed). panel.sh shouldn't update the title on tag_flags because then, panel.sh has to think too much
22:01 <ypnos> do you really want to introduce new hooks at this point
22:01 <ypnos> it is getting pretty random
22:01 <thorsten`> yep
22:02 <thorsten`> so you prefer re-emitting window_title_changed?
22:02 <ypnos> if the current window disappears because it is moved to another tag window_closed might be expected but is not the right hook
22:02 <thorsten`> oh right
22:02 <thorsten`> then window_closed isn't the right way
22:03 <The-Compiler> +1 for window_title_changed fwiw
22:04 <The-Compiler> damn, that has a WINID
22:04 <The-Compiler> just emit window_title_changed with WINID 0 or -1 or so?
22:04 <ypnos> 0 is the root window, right?
22:04 <thorsten`> or without any arguments
```

PR #41, opened 2015-04-14
Add missing window_title_changed hook.

matiaslina:
This patch fixs issue #40.

### DIFF
--- a/src/clientlist.cpp
+++ b/src/clientlist.cpp
@@ -399,6 +399,12 @@ void unmanage_client(Window win) {
     if (m) monitor_apply_layout(m);
     ewmh_remove_client(win);
     tag_set_flags_dirty();
+
+    // Get the current client and update the windows focus.
+    client = frame_focused_client(tag->frame);
+    if (!client) {
+        hook_emit_list("window_title_changed", NULL);
+    }
 }
 
 // destroys a special client


src/clientlist.cpp
```diff
@@ -399,6 +399,16 @@ void unmanage_client(Window win) {
     if (m) monitor_apply_layout(m);
     ewmh_remove_client(win);
     tag_set_flags_dirty();
+
+    // Get the current client and update the windows focus.
+    client = frame_focused_client(tag->frame);
+    if (client) {
+        client_window_focus(client);
```
The-Compiler: I'm a bit confused about what this is supposed to do - aren't you essentially focusing the already focused client here?

src/clientlist.cpp
```diff
@@ -399,6 +399,16 @@ void unmanage_client(Window win) {
...
+    } else {
+        char buf[STRING_BUF_SIZE];
+        snprintf(buf, STRING_BUF_SIZE, "0x%x", (unsigned int) 0);
+        hook_emit_list("window_title_changed", buf, "", NULL);
```
The-Compiler: The window with the ID 0 is the root window - I think it'd be better to emit `window_title_changed` with no arguments at all, i.e. `hook_emit_list("window_title_changed", NULL);` if I'm not mistaken.

src/clientlist.cpp
matiaslina: I had a doubt about this too. I've done this to manage the two cases, but obviously it can be removed.

src/clientlist.cpp
matiaslina: Yeah, `hook_emit_list("window_title_changed", NULL);` works fine. Let me commit this changes. Thanks for the feedback!

t-wissmann: Thanks for your work! IMO, the final diff looks very good. :-)

Just some non-technical/non-programming comments:

Could you please merge the two commits into one? I'd like that because
its not yet in master, and I think we should avoid adding code and
removing it directly afterwards :-) (i.e. in those cases, we can just
omit that code from our history) :-)

Oh and as The-Compiler just spotted, add curly braces around the single
statement in your then-clause:

  if (!client) {
     hook_emit_list("window_title_changed", NULL);
  }

t-wissmann:
Thanks for your work! IMO, the final diff looks very good. :-)

Just some non-technical/non-programming comments:

Could you please merge the two commits into one? I'd like that because
its not yet in master, and I think we should avoid adding code and
removing it directly afterwards :-) (i.e. in those cases, we can just
omit that code from our history) :-)

Oh and as The-Compiler just spotted, add curly braces around the single
statement in your then-clause:

```
  if (!client) {
     hook_emit_list("window_title_changed", NULL);
  }
```

matiaslina:
Done @t-wissmann :)

First time doing a rebase on a pull request, but I think I've done it fine.

The-Compiler:
Looks good to me and works! :+1:

One question I brought up in IRC but wasn't answered by @t-wissmann yet (I think): Maybe it'd actually make more sense to emit `focus_changed` (with no parameters), as panels should react on that as well, and what's happening is actually rather a changed focus than a changed window title? But before you change anything, let's see what @t-wissmann thinks.

matiaslina:
I've tested it with several windows and the `focus_changed` hook it's working, but only when there's another window open to focus.
For me it's fine to only emit `windows_title_changed`.

razzintown:
I agree that `focus_changed` makes more sense than `windows_title_changed`.

I noticed that moving focus to an empty frame currently emits `focus_changed   0x0`, and that clears my panel.sh window title (herbstluftwm 0.6.2). Similarly, when a window is closed, it disappears, thus the focus necessarily changes.

Also, existing behavior with multiple windows in a frame is to change focus when the focused window closes.

matiaslina:
> I noticed that moving focus to an empty frame currently emits `focus_changed 0x0`, and that clears my panel.sh window title

I didn't notice this, but seems fine :-)

Let's see what says @t-wissmann

PR #43, opened 2015-04-21
Use PKG_CONFIG variable during make.

flocke:
Allows the location of the pkg-config binary to be set as a variable
during make. This is needed to use prefixed pkg-config for cross
compilation.

Submitted with the permission of the original author (see comments in https://galileo.mailstation.de/gerrit/#/c/1249/).


The-Compiler:
Looks good to me :+1:

PR #47, opened 2015-09-11
use $TERMINAL environment variable for spawning a terminal

sedrubal:
i3 also searches for `$TERMINAL` and so I set the variable but I don't have xterm installed. And I wondered, why herbstluftwm can't start my terminal ;)

### DIFF
--- a/share/autostart
+++ b/share/autostart
@@ -21,7 +21,7 @@ Mod=Mod1    # Use alt as the main modifier
 hc keybind $Mod-Shift-q quit
 hc keybind $Mod-Shift-r reload
 hc keybind $Mod-Shift-c close
-hc keybind $Mod-Return spawn xterm
+hc keybind $Mod-Return spawn ${TERMINAL:-xterm} # use youre $TERMINAL with xterm as fallback
 
 # basic movement
 # focusing clients


share/autostart
```diff
@@ -6,6 +6,8 @@ hc() {
     herbstclient "$@"
 }
 
+if [ -z $TERMINAL ]; then export TERMINAL="xterm"; fi
```
ypnos: Only set a script-local variable here.

sedrubal: ok, why?

ypnos: Exporting a variable is a side effect on any commands called from autostart. I understand you want autostart to use your $TERMINAL env with a fallback. While I never saw $TERMINAL yet, your solution wouldn't hurt and maybe make the default script more readable. But it should not have side effects.

share/autostart
```diff
@@ -21,7 +23,7 @@ Mod=Mod1    # Use alt as the main modifier
 hc keybind $Mod-Shift-q quit
 hc keybind $Mod-Shift-r reload
 hc keybind $Mod-Shift-c close
-hc keybind $Mod-Return spawn xterm
+hc keybind $Mod-Return spawn $TERMINAL
```
The-Compiler: You could simply use `${TERMINAL:-xterm}` here and drop the other change - see [bash parameter expansion](http://wiki.bash-hackers.org/syntax/pe#use_a_default_value)

sedrubal: Ok, that's nice, thank you

share/autostart
sedrubal: ok. I thought if someone is wondering, why there spawns a xterm and if he realizes that the env var is set, he may try to set $TERMINAL to his fancy terminal emulator and then it's working as he thought it might work.

t-wissmann:
Nice Idea! I suggest melting the two commits to one single commit (in total only one word is changed from `xterm` to `${TERMINAL:-xterm}`) in order to (fast forward) merge that to master. I can do that next week if @The-Compiler isn't faster.

sedrubal:
cool :D I can rebase it, too. But then I have to force push and I don't know, if github pull requests suffer this... I'll try...

t-wissmann:
Github pull request survive this (and I don't know any other way of resubmitting commits). Can you also take care to start the commit message with an upper case letter, please? This makes your commit message look more like the other commit messages in hlwm, and matches the commit message style in linux/git/…, see e.g. https://github.com/torvalds/subsurface/blob/master/README#L87

sedrubal:
Ok, thanks for your help and sorry for offending against HACKING...

The-Compiler:
I fixed the typo and merged this - thanks!

PR #49, opened 2015-09-13
Add initial support for CMake

ideasman42:
- single file.
- supports out of source builds. (no files made in source dir)

Builds binary, docs, installs.
Example output: http://members.iinet.net.au/~ideasman42/herbstluftwm_cmake.html

Advantages with CMake are:

- loads as a project file in various IDE's.
- support multiple build-dirs (release & debug for eg).
- easier editing/configuring of build options, compilers... etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,330 @@
+
+cmake_minimum_required(VERSION 3.1)
+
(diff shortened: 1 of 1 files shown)


ideasman42:
there was an error with executable permissions, will make a new PR

PR #51, opened 2015-09-13
Make shift action give more predictable behavior

ideasman42:
When shifting a window into another container, it would add it after the active client.

This was a bit strange & not obvious where it would add,
as well as being different to the behavior of directional focus shifting.

This patch changes behavior to add to the start/end, based on the movement direction.


ideasman42:
I've got a better patch which uses spatial checks to find the best insert position. closing.

PR #52, opened 2015-09-13
Add initial support for CMake

ideasman42:
- single file.
- supports out of source builds. (no files made in source dir)

builds binary, docs, installs.
For build output see: http://members.iinet.net.au/~ideasman42/herbstluftwm_cmake.html

Advantages with CMake are:

- loads as a project file in various IDE's.
- support multiple build-dirs (release & debug for eg).
- easier editing/configuring of build options, compilers... etc.
- improved portability (between different compilers and platforms).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,335 @@
+# vim: set ts=4 sw=4 et:
+
+cmake_minimum_required(VERSION 3.1)
+
+set(CMAKE_BUILD_TYPE_INIT "Release")
+
+cmake_policy(SET CMP0005 NEW)  # Escape preprocessor strings
+cmake_policy(SET CMP0010 NEW)  # So syntax problems are errors
+
+set(CMAKE_C_STANDARD   99)
+set(CMAKE_CXX_STANDARD 11)
+
+# redirect output
+set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR}/bin CACHE INTERNAL "" FORCE)
+set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR}/lib CACHE INTERNAL "" FORCE)
+
+if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
+    message(FATAL_ERROR "In-source builds aren't supported. Remove the CMakeCache.txt and run from another directory.")
+endif()
+
+project(Herbstluftwm)
+
+
+# ----------------------------------------------------------------------------
+# Options
+
+option(WITH_DOCUMENTATION "Build with documentation" ON)
+option(WITH_XINERAMA "Use multi-monitor support" ON)
+
+set(DESTDIR "" CACHE PATH "Root directory, prefix for CMAKE_INSTALL_PREFIX and CMAKE_INSTALL_SYSCONF_PREFIX when set")
+set(CMAKE_INSTALL_SYSCONF_PREFIX "/etc" CACHE PATH "Directory to install configuration files")
+
+set(SYSCONFDIR "${DESTDIR}/etc")
+set(CONFIGDIR "${SYSCONFDIR}/xdg/herbstluftwm")
+
+
+if(CMAKE_COMPILER_IS_GNUCC)
+    set(CMAKE_C_FLAGS "-pedantic -Wall")
+    set(CMAKE_CXX_FLAGS "-pedantic -Wall -Wno-sign-compare -Wno-narrowing -Wno-deprecated-register")
+endif()
+
+
+# ----------------------------------------------------------------------------
+# Find Libraries
+
+include(FindPkgConfig)
+pkg_check_modules(GLIB2 REQUIRED glib-2.0)
+
+if(WITH_XINERAMA)
+    find_package(X11 REQUIRED)
+
+    if(NOT X11_Xinerama_FOUND)
+        set(WITH_XINERAMA OFF)
+    endif()
+endif()
+
+
+# ----------------------------------------------------------------------------
+# Find Vars
+
+# VERSION_GIT
+set(VERSION_GIT " (unknown)")
+if(EXISTS ${CMAKE_SOURCE_DIR}/.git)
+    find_package(Git)
+    if(GIT_FOUND)
+        execute_process(
+            COMMAND git rev-parse --short HEAD
+            WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+            OUTPUT_VARIABLE VERSION_GIT
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
+        set(VERSION_GIT " (${VERSION_GIT})")
+    endif()
+endif()
+
+# VERSION_* from 'version.mk'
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,362 @@
...
+set(SHORTVERSION "${VERSION_MAJOR}\.${VERSION_MINOR}\.${VERSION_PATCH}${VERSION_SUFFIX}")
+set(VERSION "${SHORTVERSION}${VERSION_GIT}")
+unset(_contents)
+
+
+# ----------------------------------------------------------------------------
+# Program: 'herbstluftwm'
+
+set(SRC
+    src/clientlist.cpp
```
ypnos: can you please use src/header pairs?
e.g.
````
    src/clientlist.cpp src/clientlist.h
    src/command.cpp src/command.h
````

File: CMakeLists.txt [hunk 2]
```diff
@@ -0,0 +1,362 @@
...
+    -DHERBSTLUFT_VERSION_PATCH=\"${VERSION_PATCH}\"
+)
+
+set(LIB
+    ${X11_X11_LIB}
+    ${GLIB2_LIBRARIES}
+)
+
+set(SRC
+    ipc-client/client-utils.c
```
ypnos: see above

CMakeLists.txt
```diff
@@ -0,0 +1,362 @@
...
+set(DOCDIR ${DATADIR}/doc/herbstluftwm)
+set(EXAMPLESDIR ${DOCDIR}/examples)
+set(LICENSEDIR ${DOCDIR})
+set(XSESSIONSDIR ${DATADIR}/xsessions)
+set(ZSHCOMPLETIONDIR ${DATADIR}/zsh/functions/Completion/X)
+set(BASHCOMPLETIONDIR ${SYSCONFDIR}/bash_completion.d)
+
+
+install(
+	TARGETS
```
ypnos: tab instead of spaces here

ideasman42:
Updated based on @ypnos feedback

ypnos:
Here would be respective documentation:
https://github.com/ypnos/herbstluftwm/commit/9b6c6856db51a793be2b244306e24065f29b0099
https://github.com/ypnos/herbstluftwm/commit/e74306a56fb34384711494c9473b943042891b4a

ideasman42:
@ypnos, thanks for looking into this, just a note that your current edits mix in the cmake/make docs... I'd prefer they be kept under separate headings, if new info gets added it makes instructions harder to follow.

Also, would rather handle docs separately if possible, first see if CMake is going to be accepted.

The-Compiler:
-0.5 from my side :smile:

I think the few 100 lines of plain Makefiles we have right now work very well, so I don't really see a reason to introduce additional complexity (and a dependency).

Also it's kind of worrying that nobody from the "core" people really knows CMake very well (or do you, @ypnos?).

ypnos:
The Makefile solution has drawbacks that may be relevant to some developers. The documentation itself states an important drawback outright. Also the current Makefile solution, which also spans several files, is more complex than it should be. So I understand the reasoning to offer an alternative.

I think the only drawback of adding this file is that it needs to be maintained.
I consider myself quite expertised in CMake. I really hate it with a passion because I had to work with it for over 5 years. I still volunteer to maintain this file as it is rather simple and it won't take much effort to keep it updated.

t-wissmann:
I personally hate the default workflow of CMake. (I'm always confused what to do when building weechat). Can we somehow put a minimalistic `Makefile` in the root dir, such that users unfamiliar with cmake can just type `make` (doing all the cmake-specific things)?

In times of `git`, I don't see the point in seperate "release" & "debug" directories.

The advantage of CMake I see is that we do not need take care of (the colors and of) the dependency files `src/*.d` anymore.

t-wissmann:
Does someone know how well CMake is working on other unixes like *bsd? Because right know they have to install GNU's make in order to build hlwm.

BTW I don't think that the current Makefile setup is too complex. Makefile contains the main build system, `version.mk` nothing but the current hlwm version, `config.mk` platform specific paths, and `colors.mk` does the coloring of the output, i.e. if you don't care about colored output, just ignore `colors.mk`. So I don't share @ypnos' opinion.

ideasman42:
Regarding having a wrapper `Makefile`, this will do a build in the current directory under `build/`

    BUILD_DIR:=./build
    all:
    	cmake -H. -B$(BUILD_DIR)
    	make -C $(BUILD_DIR)
    
    install:
    	make -C $(BUILD_DIR) install

ideasman42:
@t-wissmann, yes, the Makefiles work on BSD too (no gnu-make requirement)

ypnos:
My understanding was that we keep the Makefile and provide CMake as an option. It is not uncommon for free software projects to support several build systems. Compatibility with different platforms is a great reason for it (CMake does one thing right by supporting the native toolchain on each platform without further dependencies).

In any case I would not provide a pseudo-Makefile, which is then later overwritten by CMake, I think that is confusing and bad style. I would rather write a 'configure' script, as `./configure && make && make install` is pretty well-established.

I didn't say that the Makefile-solution is "too" complex. It is just unnecessarily complex in terms of file content and customization, as a CMake (or other higher level build system) based solution is clearer in expression and aids the user in customization (in CMake via ccmake tool, although it sucks in a few ways). This is just a counter-argument from my side to "Makefile is good enough" (next to the obvious defect with deps).

So for some people, CMake is unnecessarily complicated (they may have to install it, and familiarize themselves with it). For others it is the Makefile. That's why I would just offer both and again, I volunteer to maintain the CMakelists.txt along.

ideasman42:
@ypnos, right, the 2 can co-exist and one need not replace another.

Just one minor point, that a `pseudo-Makefile` can generate the build in a subdirectory, and need not be overwritten.
That'd be problematic since you'd always have to remember not to accidentally commit it... include in patches etc.

Another option could be to rename `Makefile` to `GNUMakefile`, to avoid naming collision... though I'd recommend just disallowing in-source-cmake-builds, so makefile collisions *never* happen. (as the current patch ensures).

t-wissmann:
Using the two build systems at the same time feels like combining the disadvantages of both. But if you're right that its common for open source projects, then adding this pull request just means: The non-official build system CMake is maintained in the official repos such that the CMake-people don't need a extra repo and all the others can still use the old-fashoned plain Makefile?

If that makes you happy, then why not? (explicit question at all :-) )

BTW, of course a Make-wrapper around cmake will *never ever* overwrite the Makefile-wrapper itself. I already asked in #cmake how to configure an alternate name for the `Makefile`, but did not get an answer yet.

ideasman42:
@t-wissmann, makefile's name is hard-coded in CMake, think its best to only support out-of-source builds.

The main drawbacks as far as I can see:

- A developer may commit, and forget to update one.
- Any bigger changes to the build process can become a hassle.
- Another technology to have to learn.
- More to document.

The-Compiler:
As for 1), we probably should set up Travis (which is a good idea regardless), and make it build both.

ideasman42:
Any updates needed for this PR to be accepted?

ideasman42:
Updated HACKING with instructions on using CMake.

ypnos:
Looks fine for me.

PR #60, opened 2015-11-19
Add .travis.yml for winterbreeze.

The-Compiler:
This currently compiles hlwm with GCC and clang, and does a simple smoke test
(running hlwm with Xvfb and trying a "hc version").

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+dist: trusty
+
+language: cpp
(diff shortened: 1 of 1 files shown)


The-Compiler:
Compiling and the autostart print some warnings, but I guess that's because stuff is WIP.

I'll push this to winterbreeze, as it's sure a good starting point.

PR #62, opened 2016-02-15
Change all anonymous github URLs to https scheme

siccegge:
Preventing all classical MITM schemes and at least provide as much
security as X509 PKI can provide


t-wissmann:
I think we discussed it in the irc that the `.git` suffix isn't necessary, so I dropped it and rebased your commit to the current `master`.

PR #63, opened 2016-02-15
Verbose

siccegge:
Allows turning `verbose` on during runtime of hlwm

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -91,7 +91,7 @@ SettingsPair g_settings[] = {
     SET_INT(    "auto_detect_monitors",            0,           NULL        ),
     SET_INT(    "pseudotile_center_threshold",    10,           RELAYOUT    ),
     SET_INT(    "update_dragged_clients",          0,           NULL        ),
-    SET_STRING( "tree_style",                      "*| +`--.",  FR_COLORS   ),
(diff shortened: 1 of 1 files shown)


t-wissmann:
Thanks! By the way, you've forgotten the documentation (man page and NEWS), but I've just added that in de85af40b19481a8b21aad7af077d8f19860d94b.

t-wissmann:
As it just turned out today, the fetch_frame_colors() updates the global variable `g_tree_style` if the setting `tree_style` changes. Of course, fetch_frame_colors() was the wrong place to do it, and indeed looked like a bug in the settings table. 0e370d5ea3ccd73a678398c7a7a4127eeeabac5d fixes this.

The-Compiler:
@t-wissmann Did you mean to answer to this issue? I can't see how this is related to verbosity off-hand :wink:

t-wissmann:
@The-Compiler The connection is commit ba31d3a48a80a0f00bcffe06f02c5149f036eb07 ("Updating tree_style doesn't need any update to framecolor") which was part of this pull request and revealed the issue that was finally fixed in 0e370d5ea3ccd73a678398c7a7a4127eeeabac5d (see also its commit message).

Issue #64, opened 2016-02-28
XFetchName doesn't always return a name.


prohunterdogkeeper:
Even though a window might have a WM_NAME property, XFetchName doesn't always return a name - for some reason. Googling around a bit I couldn't figure out why though. I noticed that herbstluftwm didn't name Steam's windows. Steam windows don't have a _NET_WM_NAME property, so herbstluftwm falls back to calling XFetchName to get the WM_NAME property.
Running herbstluftwm in verbose, it prints that "no title for window xxx found, using ''" when a Steam window is focused, where xxx is the ID of a Steam window. This means that XFetchName returned 0, even though using xprop shows the window to indeed have a WM_NAME property. 

I did a test of the XGetWMName function, using code from this stack overflow thread [1], which returned the correct name, where XFetchName returned no name.

I propose to change XFetchName to XGetWMName.

[1] https://stackoverflow.com/questions/8825661/xfetchname-always-returns-0

ypnos:
To me this looks like a very reasonable suggestion.
It needs to be noted that XGetWMName explicitely deals with encoding and provides a way to query for the encoding of the returned value, while XFetchName's result is "implementation dependent". To quote:

> f the data returned by the server is in the Latin Portable Character Encoding, then the returned string is in the Host Portable Character Encoding. Otherwise, the result is implementation dependent.

"Implementation dependent" might very well mean returning zero.

Could you maybe prepare a pull request for the change?

prohunterdogkeeper:
I'm not much of a programmer, and so I haven't set up a git workflow, neither do I feel very confident in my programming abilities, but wouldn't something as simple as this work? 
```
XTextProperty xtp_new_name;\n
/* fall back to WM_Name */
if (0 != XGetWMName(g_display, client->window, &xtp_new_name)) {
    new_name = g_string_new(xtp_new_name.value);
```
This replacing lines 871 through 875 in src/clientlist.cpp

I'm not familiar with xlib, and so I assume you don't need to XFree() the XTextProperty object, since it doesn't mention it in the manpage for XGetWMName, while it does mention it in the description of XFetchName?

ypnos:
Yes this looks o.k. however based on the documentation of XTextProperty I think one needs to not directly access `value` but use conversion methods instead. Maybe thorsten knows better how to do this?

https://tronche.com/gui/x/xlib/ICC/client-to-window-manager/converting-string-lists.html

prohunterdogkeeper:
[...]

PR #65, opened 2016-03-11
Changed XFetchName to XGetWMName, fixing issue #64

prohunterdogkeeper:
Changed XFetchName to XGetWMName as described in issue #64, first time contributing to any project and first time writing and not just reading C/C++ so please go ahead and reject / tell me if I did something wrong.

### DIFF
--- a/src/clientlist.cpp
+++ b/src/clientlist.cpp
@@ -868,15 +868,24 @@ void client_update_title(HSClient* client) {
     GString* new_name = window_property_to_g_string(g_display,
         client->window, g_netatom[NetWmName]);
     if (!new_name) {
-        char* ch_new_name = NULL;
-        /* if ewmh name isn't set, then fall back to WM_NAME */
-        if (0 != XFetchName(g_display, client->window, &ch_new_name)) {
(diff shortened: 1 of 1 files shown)


ypnos:
Looks pretty good to me. In two lines the indention seems to be off by one character.
Nice job!

prohunterdogkeeper:
Yeah, line 873 and 874?
I have no idea where that came from, it's not there in my fork? I haven't changed anything since the pull request.

ypnos:
The problem is that sometimes you have tabs in there, while the rest of the file indents with spaces.

Because Thorsten likes clean commits, I suggest you replace your tabs with spaces, then do a `git commit -a -v --amend` and then push fast-forward (`git push -f`), something that normally should **not** be done. Then it will stay one commit. No need to redo the pull request.

prohunterdogkeeper:
Alright, fixed. Thanks for the help.

The-Compiler:
> and then push fast-forward (`git push -f`)

You mean force-push - fast-forward is when you're pulling and there is no merge because it's the same base :wink:

PR #67, opened 2016-04-18
Fix list formatting in FAQ

dnnr:
Apparently, having a listing block in a list item fails if the item is
indented. Without the indentation, the listing block as well as the subsequent
text are correctly rendered as part of the list item.


t-wissmann:
Thanks for spotting this. However it suffices to connect the paragraphs by pluses. I prefered that because that makes the indentation of the source code more similar to that of the output.

PR #68, opened 2016-04-18
Apply rules to unmanaged popup windows

dnnr:


t-wissmann:
I think the changes are fine. The only problem is (or is it a problem?) that a matching rule can switch the tag using `switchtag` although the unmanaged window is not on any tag at all.

t-wissmann:
I think the changes are fine. The only problem is (or is it a problem?) that a matching rule can switch the tag using `switchtag` although the unmanaged window is not on any tag at all.

dnnr:
Good point. Would this be solved by moving up the if-block where `force_unmanaged` is checked to right before the evaluation of `changes.monitor_name` (and ultimately `changes.switchtag`)? Seems to me that the order of these sections isn't vital, but I'm not sure.

dnnr:
Is merged.

PR #74, opened 2016-06-03
Save previous tag when stealing from other monitor

hpdeifel:
When `use`ing a tag that was already shown on another monitor and
swap_monitors_to_get_tag was set, we did not save the current tag in
monitor->previous_tag which resulted in use_previous not jumping back to
the previous tag but the one before that.


t-wissmann:
Thanks for fixing that! Though I noticed that something strange is happening in those cases, I never spent time on investigating that....

PR #75, opened 2016-06-08
Add fish shell completion script

wilriker:
After switching to `fish` shell I was missing the completions for `herbstclient` I had when using `bash` or `zsh`. So I wrote a completion script to get the full completion also in `fish`.


t-wissmann:
Thanks! I tried your completion file by copying it to `/usr/share/fish/completions/herbstclient.fish` (Is this the supposed way of installing fish completion files?). However, when typing
```
herbstclient use <tab>
```
then the completion lists `herbstclient` and all commands starting with `use`, even though I typed a space before hitting the `<tab>`-key. I expected that all the tag names are listed instead.

Am I doing something wrong or is this a bug of your completion script?

ypnos-web:
Thanks for this awesome work! I wonder however if it might be more appropriate to try a pull request with the fish shell directly. They are hoarding completion files for all kind of commands, see:
https://github.com/fish-shell/fish-shell/tree/master/share/completions

I observe the same problem as @t-wissmann. Completions work right in the case of:
`herbstclient use a<tab>`
but not in the case:
`herbstclient use <tab>`

t-wissmann:
I prefer keeping it here, because in case of new command line flags for `herbstclient`, we can directly adjust the completion. Or in other words: the version of the completion file is more specific to the herbstluftwm version than to the fish version (assuming that they keep their completion api backwards compatible), so the completion file should be contained in a herbstluftwm-distro-package and not in the distro-package of fish.

wilriker:
> `/usr/share/fish/completions/herbstclient.fish` (Is this the supposed way of installing fish completion files?)

At least in latest `fish` 2.3.0 it is supposed to go to
`/usr/shar/fish/vendor_completions.d/herbstclient.fish`

I will have a look at the described issue. I was testing with
`herbstclient toggle <tab>` and this one is working as supposed. I will get back to you as soon as I have fixed this issue. I have an idea what is going wrong here.

wilriker:
So, I have refactored the code a lot. This is now working except for two things:

1. If the command contains an argument starting with a dash like in `herbstclient use_index -1` this will be ignored. As I am typing this a solution to this problem came to my mind. I will update the code again.
2. Going back in the commandline will not let you insert anything but `herbstclient` options like `-n` or `-0` but I think this is not a primary goal, is it?

wilriker:
So, issue with what I listed in 1) above is also fixed now.

share/herbstclient.fish
```diff
@@ -0,0 +1,37 @@
...
+        if string match -q -- "-*" $token
+            set -e tokens[1]
+        else
+            break
+        end
+    end
+
+    set -l position (count $tokens)
+
+    # If we have tokens but and cursor sits on a word (not space) decrease position due to 0-based in hlwm and 1-based in fish
```
ypnos: Wording in this comment can be improved. "but and" is probably a typo.

ypnos:
I tried your new version and it seems to work as it should.

wilriker:
Wording in comment improved. ;-)

ypnos:
So I guess the Makefile should be extended to install the file to `/usr/shar/fish/vendor_completions.d`?

wilriker:
You are right. I thought that would be something that packagers would only do (due to various different directory structures across Linux distros) but you are right that it should install the file to the default location that `fish` is expecting.

wilriker:
Just out of curiousity: is something missing for this PR to get actually pulled or do you have a policy of only integrating PRs at a certain point in time?

t-wissmann:
Thanks for reminding me of this. One tiny issue still left is that the completion should not add a space after these object paths like `clients.`:
```
herbstclient attr clients.fo<tab>
```
In this case, `<tab>` completes to `clients.focus.<space>`, although it should not print a space at the and. In order to implement this in bash, I have added the `completion_shell` command.

However, I don't see how this can be fixed for the [fish completion](https://fishshell.com/docs/current/commands.html#complete). So I'll merge your PR for the time being, since the space thing is just something _nice to have_.

If you come up with a solution, I'd be happy to merge that as well!

wilriker:
> However, I don't see how this can be fixed for the fish completion.

Sadly, neither do I.

PR #90, opened 2016-12-22
Make compatible with dmenu < 4.5

wilriker:
Also exit without creating a new rule if the user aborted dmenu selection

### DIFF
--- a/share/dmenu_run_hlwm
+++ b/share/dmenu_run_hlwm
@@ -8,23 +8,10 @@ fi
 # Get the currently active tag
 tag=$(herbstclient attr tags.focus.name)
 
-# Prints the path of of the binary in path selected by dmenu
-dmenuPrintPath() {
-	cachedir=${XDG_CACHE_HOME:-"$HOME/.cache"}
-	if [ -d "$cachedir" ]; then
-		cache=$cachedir/dmenu_run
-	else
-		cache=$HOME/.dmenu_cache # if no xdg dir, fall back to dotfile in ~
-	fi
-	IFS=:
-	if stest -dqr -n "$cache" $PATH; then
-		stest -flx $PATH | sort -u | tee "$cache" | dmenu "$@"
-	else
-		dmenu "$@" < "$cache"
-	fi
-}
+selectedPath=$(dmenu_path | dmenu "$@")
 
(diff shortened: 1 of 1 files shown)


t-wissmann:
What's the purpose of deleting the entire `dmenuPrintPath` function? Here with dmenu2 https://bitbucket.org/melek/dmenu2 I don't even have `dmenu_path` in my path. Or asking the other way around: what from the previous `dmenu_run_hlwm` has been incompatible to dmenu < 4.5?

wilriker:
The problem with dmenu < 4.5 is that there is no `stest`. And as `dmenuPrintPath` was nearly exactly the content of dmenu 4.6's `dmenu_path` I thought it would make sense to direct the decision on how to compile the list over to dmenu.

Sadly I am stuck with CentOS 6 at work (as a friend of mine dubbed it: CenturyOldSoftware) I either have to use ancient versions of software from the (very restricted) repos. In that case it would be dmenu 4.1. Or I have to compile everything on my own (including hlwm) and I try to do this as few as possible (by the way, hlwm 0.7.0 is not being compiled by CentOS's ancient `gcc` - I don't remember the version by heart but it is one before they introduced lambdas) that's how I ran into this problem in the first place.

Nevertheless I did not know anything of dmenu2 and its lack of `dmenu_path`. I will submit a new patchset tomorrow that will look for what is available and takes a decision from that.

Best regards
Manuel

t-wissmann:
All right, the new version seems to work for me. I have squashed your two commits into one, because the second commit re-introduces the dropped content of `dmenuPrintPath`. It's now in master.

Thanks for your contribution.

Regarding your `gcc` problem: from now on, you need a gcc/g++ or clang that supports `c++11`. Maybe you can statically compile link it on a newer OS and then use that binary on centos.

wilriker:
Thanks for accepting!

> Regarding your gcc problem: from now on, you need a gcc/g++ or clang that supports c++11. Maybe you can statically compile link it on a newer OS and then use that binary on centos.

Yeah, I could try that. But right now I have the hopes that some changes in the companies structure will allow me to switch to Arch Linux next year - I'd prefer that by far! ;-) If that does not come true, I will try your proposal.

PR #93, opened 2017-02-14
Feature/mirror

towa:
This is a mirror function based on the existing rotate function.
It mirrors the focused tag on the vertical axis.
It's particularly useful on dual monitor setups when you want the important frames to remain at the center even if they are moved from one monitor to the other.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -494,6 +494,9 @@ rotate::
     Rotates the layout on the focused tag counterclockwise by 90 degrees. This
     only manipulates the alignment of frames, not the content of them.
 
+mirror::
+    Similar to rotate. Mirrors the layout on the focused tag at the vertical axis.
+
 set 'NAME' 'VALUE'::
     Sets the specified setting 'NAME' to 'VALUE'. All <<SETTINGS,*SETTINGS*>>
     are listed in the <<SETTINGS, section below>>.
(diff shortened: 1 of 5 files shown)


t-wissmann:
I feel uncomfortable with implementing these kind of very special features, especially when I don't really understand the motivation.

After swapping the contents of the two monitors you would like to mirror the layout in each monitor again such that the same windows are again at the central edge between the monitors?

t-wissmann:
Are you still interested in this feature? If you are, I can migrate your PR to the new code base if you like!

towa:
Hi Thorsten,
That would be great! Right i stopped using hlwm and X11 altogether.
Mainly because I'm not able to rotate Monitors connected over DisplayLink :/
If only there was hlwm for wayland :)

t-wissmann:
All right, then I'll do this in a separate PR. I'm sorry to hear that you can't use X11 anymore.
I haven't even started porting hlwm to wayland, so I'm not sure whether this will happen at all :-(

syslino:
> I'm not sure whether this will happen at all :-(

I really hope that it will be ported! I'm using it on X11 currently, but if I'll be forced to use wayland I **_really_** hope that this software will be ported.

t-wissmann:
I've adapted the 'mirror' command to the new codebase in #1046.
Unfortunately, it probably does not help you much if you are still not on X11 anymore. I'm still trying to make hlwm more and more independent of X11 in order to add a wayland backend at some point but it is still a lot of work and I can't make any promises when this ever will be finished.

PR #100, opened 2017-08-04
Cleanups for panel.sh

johnLate:
panel.sh used a mix of `"${herbstclient_command[@]}"` and `herbstclient`. Also, if herbstclient_command was set and was an array, unquoted use would not work with dzen's `^ca()`.

I have tested it by setting `herbstclient_command=( bash -c 'exec herbstclient "$@"' arg0 )` just before `HC_QUOTED`.

### DIFF
--- a/share/panel.sh
+++ b/share/panel.sh
@@ -82,7 +82,6 @@ hc pad $monitor $panel_height
         # This part prints dzen data based on the _previous_ data handling run,
         # and then waits for the next event to happen.
 
-        bordercolor="#26221C"
         separator="^bg()^fg($selbg)|"
         # draw tags
         for i in "${tags[@]}" ; do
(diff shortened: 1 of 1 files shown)


share/panel.sh
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env bash
 
+quote() {
+	local q="$(printf '%q ' "$@")"
+	printf '%s' "${q% }"
+}
+
+HC_QUOTED="$(quote "${herbstclient_command[@]:-herbstclient}")"
```
The-Compiler: This variable name should probably be lower-case.

johnLate:
I've changed `HC_QUOTED` to `hc_quoted`.

t-wissmann:
Seems to work fine, thanks!

PR #107, opened 2017-09-20
Make DESTDIR and PREFIX settable from environment

mgoral:
This follows GNU standards where variables like DESTDIR should be controllable via environment variables: https://www.gnu.org/prep/standards/html_node/DESTDIR.html


t-wissmann:
Sure! Sorry for ignoring this PR for a while...

PR #109, opened 2017-11-03
Fix error clamping first client when unmapping

ideasman42:
Comparison with size_t caused zero to wrap to last


t-wissmann:
Sorry didn't go through open PRs for a while...

Is `(int)(count -1)` really correct? Shouldn't it be '((int)count)-1'?

PR #110, opened 2017-12-09
Fix typo for client position in floating mode

johnLate:
If you have a hlwm monitor with an y-offset, the position of floating windows was calculated wrong. This also affected mouse resize behavior.


t-wissmann:
Thanks a lot!

PR #111, opened 2017-12-09
Fix mouse resize in tiled mode for multi-monitor

### LINKED ISSUE #87: Tiled mode mouse resize does not work for all frame borders on second monitor
Situation:

2 herbstluftwm monitors (left, right).
A Tag, horizontal split, right side vertical split, top right a client window:
(split horizontal:0.500000:1 (clients vertical:0) (split vertical:0.500000:0 (clients horizontal:0 0x180000a) (clients horizontal:0)))

```
+---+------+
|   |Client|
|   +------+
|   |      |
+---+------+
```

If the tag is on the left monitor, mouse resize in tiled mode works fine.

If the tag is on the right monitor, mouse resize works for the frame border of the horizontal split, but not for the vertical split of the right side, below the client.

johnLate:
rect = &frame->last_rect is absolute, so there's no need to convert g_button_drag_start.

Fixes #87.

### DIFF
--- a/src/mouse.cpp
+++ b/src/mouse.cpp
@@ -324,7 +324,6 @@ void mouse_function_resize_tiled(XMotionEvent* me) {
     if (g_drag_init_done == false) {
         g_drag_init_done = true;
         /* initialize local state */
-        HSMonitor* monitor = g_drag_monitor;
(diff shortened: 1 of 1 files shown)


t-wissmann:
Thanks again!

PR #117, opened 2018-07-24
Added herbstclient cwd and pwd

countingsort:
Since hlwm also handles process spawning (with `herbstclient spawn`) I thought it would be fair enough, if it would be able to change directories (on a tag basis).

In case somebody wants to script something theres also `herbstclient pwd`, that just prints the current working directory.


The-Compiler:
Why would the current working directory be on a per-tag basis? That seems quite unintuitive to me...

countingsort:
So that when you work on multiple things in different tags it won't disturb you.

I.e. some documents with documentation on one tag and the source (probably in another directory) on another one.

t-wissmann:
For which applications do you need it concretely? I have a script https://github.com/t-wissmann/dotfiles/blob/master/utils/with-same-pwd[with-same-pwd] that expects a command as a parameter, reads the PWD from the currently focused window and calls that command with the same PWD. E.g. `with-same-pwd thunar` launches a thunar file browser in the current directory of a shell running in the urxvt terminal.

countingsort:
@t-wissmann For pretty much everything I do, while working on a project.

The problems I have with your script, and similar are:

1. I don't have a general project dir (ie when I have a window in the src dir the next one will be spawned there instead of my project root)
2. Being in a project is bound to having windows left open. I can't just close what I finished and open a new window after that.

ypnos:
I never understood why herbstluftwm should have an unbuilt fully-fledged shell or any shell-like functionality to begin with. However, in the current framework, the existing methods all fulfill a somewhat generic purpose.

Handling of the working directory for me is a task so far off, on the other hand tying the working directory variable to the tag looks quite random as well (diminishing a pretty slim use-case further).

Workaround: You can have a script track the tag changes (using `herbstclient -i`) and spawn another script that uses this information to run a program in the associated working directory.

countingsort:
@ypnos What would you suggest instead of having it tied to a tag? A global one?

I've talked to a few people prior to making this PR and it seems that the general consent is one wd per tag is more intuitive (maybe I just hang out with weird people though).

Perhabs I should've added the possibility to do something like `herbstclient pwd [tagname]` to get the working dir of another tag, so that stuff like `herbstclient cwd $(herbstclient pwd OTHER_TAG)` becomes possible.

Maybe `herbstclient cwd [global|current] <path>` should be a thing so that you can do that, if you prefer it.

ypnos:
I realized my proposed workaround was overly complicated, actually you can just spawn a script that queries `herbstclient attr tags.focus.name` and then decides for a working directory and spawns the process.

The only benefit I see in your PR over this is the convenience of storing a user state variable (in this case, cwd) within herbstluftwm. I think allowing user variables as custom object properties (for all objects) would be a generally usable feature and therefore quite nifty. Obviously this is something that should wait for winterbreeze and is going beyond this PR.

I think a global CWD would be pretty much useless. I very well understand the reasoning behind your proposed functionality, but me personally I would argue for implementing this outside of hlwm, it is a bit more complicated but still rather simple to do. I offer to help with that as well.

How do you currently go about setting the cwd? What is the scenario?

countingsort:
@ypnos I'm using it mostly from inside ranger and with a keybinding to "reset" (i.e. go back to $HOME).

I'd stick to this until there's a reason not to (e.g. winterbreeze, or the wm I'm currently working on become usable). Thanks for the help though.

t-wissmann:
> The only benefit I see in your PR over this is the convenience of storing a user state variable (in this case, cwd) within herbstluftwm. I think allowing user variables as custom object properties (for all objects) would be a generally usable feature and therefore quite nifty.

this feature already exists. The command is called `new_attr`. In the scenario of cwd per tag, one can do the following:
```
herbstclient chain , new_attr string tags.focus.my_cwd , set_attr tags.focus.my_cwd ~/my/project/dir
```
So you can have the following alias in your shell to `cd` the focused tag to the cwd of your shell:
```
alias hlwm_cd='herbstclient chain , silent new_attr string tags.focus.my_cwd , set_attr tags.focus.my_cwd "$PWD"'
```

And instead of e.g.~`spawn thunar` do the following:
```
herbstclient mktemp string CWD_VAR chain , set_attr CWD_VAR . , silent substitute X tags.focus.my_cwd set_attr CWD_VAR X , substitute CWD CWD_VAR spawn env -C CWD -- thunar
```
What this does is to temporarily create a new attribute CWD_VAR with the value `.` and if the focused tag has `my_cwd` set, its value is written to CWD_VAR. Then the value of CWD_VAR (namely the desired CWD) is replaced in the command `spawn env -C CWD -- thunar`. so if the tag has `my_cwd` set, then thunar is spawned in this directory, and in `.` (i.e. hlwm's working directory) otherwise.

Please note that:
1. you only need to replace thunar in the above command, all the rest is generic.
2. this works very well with spaces in directories because all the replacements are performed on string arrays and not on strings.

t-wissmann:
Oh and as a follow up: write the following in your autostart:
```
spawn_cd=( mktemp string CWD_VAR chain , set_attr CWD_VAR . , silent substitute X tags.focus.my_cwd set_attr CWD_VAR X , substitute CWD CWD_VAR spawn env -C CWD -- )
```
and then you can use "${spawn_cd[@]}" instead of `spawn` in order to start applications in a tag's cwd:
```
herbstclient keybind Mod4-Return "${spawn_cd[@]}" urxvt
herbstclient keybind Mod4-Shift-Return "${spawn_cd[@]}" thunar
```

ypnos:
Right, the heat! Totally forgot about that while I thought about it yesterday.

t-wissmann:
Alright, if you have issues with it, feel free to ask!

PR #118, opened 2018-08-18
Fix spelling in manpage

siccegge:


The-Compiler:
Thanks!

The-Compiler:
Huh, FastHub (on Android) shows me a 404 when trying to merge this... Oh well, will do so later then unless someone else is faster :D

PR #120, opened 2018-09-13
Might wanna change this to 2018

MrMorkat:


The-Compiler:
Probably should do that everywhere while we're at it (search for "Copyright" in the sources)?

ypnos:
Maybe the best solution is to just get rid of dates in the copyright notices. They don't really serve any purpose.

t-wissmann:
I also prefer getting rid of the copyright notices. Shall we just drop the comments from all the files?

ypnos:
Let's do it, as I see it we have already enough votes for it.

PR #123, opened 2018-11-10
Rewrite HSFrameSplit::lookup()

dnnr:
This change is motivated by a compiler warning for the implicit
fallthrough in the switch/case block.

Extracting the special "@" case simplifies the remaining logic, making
it easier to follow, and also solving the warning along the way.

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -122,6 +122,7 @@ std::shared_ptr<HSFrame> HSFrameSplit::lookup(const char* index) {
         case '0': new_root = a; break;
         case '1': new_root = b; break;
         case '/': new_root = not_selected; break;
+        case '.':
         default: new_root = selected;
     }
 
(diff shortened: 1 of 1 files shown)


t-wissmann:
Seems fine to me. The reason for the fallthrough is that '.' is mentioned explicitly in the man page, whereas for all other characters the behaviour is undefined. (Section 'FRAME INDEX'). Maybe we can add something as a comment to not forget this when one looks at the code.

dnnr:
Good point, I've added an explicit case for `'.'`.

t-wissmann:
Doesn't this bring the fallthrough-warning back? What about adding `[[fallthrough]];`?

ypnos:
`[[fallthrough]];` is not part of C++11 and I refrained from using it somewhere else in the code to avoid old compilers warning about it (in C++11, the behavior when observing unknown attributes is undefined).

Another argument for switching to a newer standard :-P. I will open an issue about the topic of C++ standard when I get to it so we can discuss it (again, I know, but I think the new reason I see for it warrants to revisit this topic).

The-Compiler:
FWIW it's also possible to add a fallthrough comment instead, see https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

t-wissmann:
```
    case '.':
    /* FALLTHRU */
```
Sounds reasonable.

PR #126, opened 2018-11-10
Fix shadowing of root pointer in Settings ctor

dnnr:


ypnos:
Good spotting. The shadowing is something Thorsten and me talked about during SHC.

Actually, your fix is converse of a policy that we want to implement per the example of the Qt project.
Member variables have a '_' postfix, allowing a clean API regarding constructor arguments as well as accessors: `variable()` / `setVariable()`.

dnnr:
Went making this change, I let myself guide by code elsewhere, that uses the `_` suffix for the ctor parameter.

Later, I also ran into code that did it exactly the opposite way.

It's pretty subjective (and not my call either), but I'd actually prefer not having `_` suffixes at private members, and instead find individual workarounds. Often times, you can simply choose a more appropriate name for the local variable, like `new_value`.

My argument for this is that (unlike in languages like Python), the compiler is able to warn you about these kinds of shadowing. So, this kind of "in-band signalling" via variable suffixes is not actually needed, if you enable that kind of warning.

But if you're already set on doing it with a suffix and setter/getters, I'm fine with changing this PR accordingly.

ypnos:
Yes, that's the plan Thorsten and me had agreed on.

I also believe that it is the job of the compiler+IDE to distinguish local variables from member variables. On the other hand it is often very hard to fend off the `m_` crowd and the example of Qt convinced me that this is a good compromise retaining some advantages as mentioned.

t-wissmann:
> Alright! So just to be perfectly clear: In the long run, you want _every_ private member to have a `_` suffix, right?

Yes, except for `Attribute_<…>`-members about which I don't want to make any statement yet. (Maybe they are private members some where at the moment; but I definitely don't want them to have a `_` suffix)

PR #129, opened 2018-11-10
Treat CXX warnings as errors

dnnr:


dnnr:
The build for this should succeed as soon as #124, #125, #126, #127, and #128 are merged.

PR #131, opened 2018-11-11
Rework .travis.yml: Build matrix, integrate tox

dnnr:
This introduces a matrix of builds using combinations of Ubuntu versions and compilers (and their versions).

Additionally, one of the matrix entries is used to run tox instead of the usual build-install-execute script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,42 @@
-dist: trusty
-
 language: cpp
-compiler:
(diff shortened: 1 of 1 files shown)


t-wissmann:
Great thanks! :-)
Additionally, we should shut down the xvfb more gently, but that's a different issue.

PR #132, opened 2018-11-11
Use right type in ArgList::shift instead of size_t

dnnr:
This eliminates an old-style cast that was needed because size_t and
difference_type are not exactly the same thing.


t-wissmann:
Merged via https://github.com/herbstluftwm/herbstluftwm/commit/0c775a0f70b7bfca1d90c0e7799324541e0d2612

PR #133, opened 2018-11-11
Rename local variables to avoid shadowing members

dnnr:


t-wissmann:
Yes, makes sense in order to distinguish them from the non-boolean active/urgent members.

PR #134, opened 2018-11-11
Add error handling for XOpenDisplay()

dnnr:
This can be triggered by not having a DISPLAY environment variable set.

### DIFF
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -1,5 +1,6 @@
 #include "xconnection.h"
 #include "globals.h"
+#include "utils.h"
 
 #include <X11/Xlib.h>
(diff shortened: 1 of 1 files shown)


dnnr:
I haven't looked at the test suite yet, but I suppose this is a prime candidate for adding a simple test case.

t-wissmann:
So far, the test-suite checks the behaviour (and stability) of a running hlwm instance. I'm not absolutely familiar with test-suite-slang, but I'll try: We have a fixture that sets a running hlwm up in order to run test-cases which use herbstclient. If we did include something like this in the test-suite we need to tell tox which test cases need the hlwm fixture and which don't.

Apart from that, the patch looks fine.

PR #135, opened 2018-11-11
Remove die() from utils

dnnr:
The current implementation of die() triggers a compiler warning
about the format string not being a literal.

Most of its points of usage are not really needed (malloc never fails[tm]),
or can be replaced with cerr/exit.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -899,10 +899,12 @@ int main(int argc, char* argv[]) {
     XConnection X = XConnection::connect();
     g_display = X.display();
     if (!g_display) {
-        die("herbstluftwm: cannot open display\n");
+        std::cerr << "herbstluftwm: cannot open display" << std::endl;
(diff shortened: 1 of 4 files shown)


dnnr:
Down the road, we should probably just throw exceptions and catch them in main().

But for now, I'd say this is fine.

ypnos:
I only see two occasion where this method is used, one is some malloc bullshit we will get rid of anyway (and until then, malloc never fails[tm]). The other one can simply stderr and exit itself.

I suggest removal of die(). Additionally, I suggest to remove die() from ipc-utils as well.

ypnos:
Note that fcb49946b0ba4c80b4f988fa9c70cc741074d2c6 introduced another point of usage.

t-wissmann:
Right... right now, the "no display"-handling is twice anyway. Both in `XConnection` and in the `main()` function.

PR #138, opened 2018-11-12
Drop glib.h from ipc-client and further clean ups

t-wissmann:
This PR cleans up the ipc-client subdirectory aka herbstclient.

@dnnr: Can you take a look at it? I'm unsure about the changes in the Makefile and the config.mk. It's better than before, but I'm not very familiar how to properly setup the dependencies between the Makefile variables.

### DIFF
--- a/ipc-client/client-utils.c
+++ b/ipc-client/client-utils.c
@@ -2,13 +2,15 @@
 #include "client-utils.h"
 #include <stdlib.h>
 #include <stdio.h>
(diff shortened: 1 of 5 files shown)


t-wissmann:
As noticed by @dnnr I completely messed up `CPPFLAGS` and `CXXFLAGS` so I've dropped that respective commit from the PR. So for the time being, herbstclient unnecessarily links against `glib`.

PR #139, opened 2018-11-12
Underscore suffix for privates in MonitorManager

dnnr:


t-wissmann:
> It looks like the global isn't intended to stay anyway, but how would you feel about quickly renaming it so that shadowing is less likely? I've already prepared such a change, and there are about 24 points of usage.

Fine for me, and e.g. `g_tags` is a better name anyway.

PR #143, opened 2018-11-12
Use underscore suffix for privates in TagManager

dnnr:


t-wissmann:
I think this one change `tags` vs `global_tags` should not be in this PR.

dnnr:
Oops, my bad! Fixed.

PR #145, opened 2018-11-12
Gently shut down hlwm in travis

t-wissmann:
Wait for hlwm to shut down in travis in order to avoid the fatal IO error message complaining about the lost display.


dnnr:
Looking good!

PR #150, opened 2018-11-12
Suppress unused root pointer warning in Settings

dnnr:


dnnr:
Just in case anybody wonders: Turns out that Travis still doesn't quite to what it's supposed to, so we currently don't actually have a clang build with this kind of warning being treated as an error. (My bad.)

If we did, this unused member would have broken the build.

I'm already working on fixing the Travis configuration, so bear with me...

PR #151, opened 2018-11-14
Fix compiler selection .travis.yml; general rework

dnnr:
Setting CC and CXX as environment variables gets overwritten by Travis
before the "script" is executed. The previous configuration basically
always used the default compiler.

To fix this, the variables are now set in a single-line "before_install"
script.

This change prints the CC/CXX values right before calling make, to
allow for proper manual verification.

As it turned out, the compilers we intended to use were not even
available. Installation of the right compiler is fixed here, too. Along
the way, the range of compiler is extended.

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -623,7 +623,7 @@ TilingResult HSFrameLeaf::layoutMax(Rectangle rect) {
 }
 
 void frame_layout_grid_get_size(size_t count, int* res_rows, int* res_cols) {
-    int cols = 0;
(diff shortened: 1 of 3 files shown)


dnnr:
Beware that the build for this will fail until #150 is merged.

Testing seven combinations is probably overkill, so I'm open to suggestions on which jobs to remove.

I'd tend toward dropping all older clang versions. Basically, we just want to make sure that there is *one* way to build it on an ancient Ubuntu.

PR #152, opened 2018-11-14
Fix cumbersome API for parsing directions

ypnos:
Have a small helper struct that encodes if a valid direction was parsed
or not. Apart from that, treatable in the same fashion as a C++ enum class.

Please comment whether you like this solution to the awkward workaround introduced in e3318f536.

### DIFF
--- a/src/floating.cpp
+++ b/src/floating.cpp
@@ -14,30 +14,35 @@
 
 using namespace std;
 
+Direction::Direction(char ch)
+{
+    std::map<char, V> mapping = {
+        {'u', Up}, {'r', Right}, {'d', Down}, {'l', Left},
+    };
+    auto it = mapping.find(ch);
+    if (it != mapping.end()) {
+        value = it->second;
+        valid = true;
+    }
+}
+
+bool operator==(const Direction &lhs, const Direction::V &rhs) {
+    return lhs.valid && lhs.value == rhs;
+}
+
 void floating_init() {
 }
 
 void floating_destroy() {
 }
 
-int char_to_direction(char ch) {
-    switch (ch) {
-        case 'u': return DirUp;
-        case 'r': return DirRight;
-        case 'l': return DirLeft;
-        case 'd': return DirDown;
-        default:  return -1;
-    }
-}
-
 // rectlist_rotate rotates the list of given rectangles, s.t. the direction dir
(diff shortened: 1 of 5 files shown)


src/floating.cpp
```diff
@@ -14,30 +14,35 @@
...
-void floating_init() {
+Direction::Direction(char ch)
+{
+    std::map<char, V> mapping = {
+        {'u', Up}, {'r', Right}, {'d', Down}, {'l', Left},
+    };
+    auto it = mapping.find(ch);
+    if (it != mapping.end()) {
+        value = it->second;
+        valid = true;
```
dnnr: Is indicating a parsing failure within the ctor the only purpose of the `valid` member?
There is no default ctor being used anywhere, right?
If so, I'd suggest throwing an exception here instead, and get rid of the valid flag.

(If there is any need for having "undefined, invalid" instances of `Direction`, I would recommend using `std::optional`)

ypnos:
Thank you for the review. Thorsten tries to avoid exception usage in the code. I have no strong opinion either way so I respect that decision and avoid them. Note that `Attribute_::parse()` throws, something that also needs to be documented.

I also thought about two alternative designs:
1. Make it a `enum class` and have a parse() method (which would flow around freely… something I dislike) that returns a tuple (success, value). This would be nice to use only if we level up to C++17 and structured bindings.

2. Strip the static parse()/str() functionality pair out of `Attribute_<T>` into something like `Primitive<T>` and have `Attribute_<T>` inherit from that (multiple inheritance). So we can deal with these kinds of primitives that are communicated with the user in a coherent way. The `Direction` might probably not be used as an `Attribute_` later, but e.g. layout algorithms will (when Frames become part of the object tree).

This is by part also an answer to your other question: Yes, the pollution of types left and right is homegrown. I think it would be a great opportunity for cleanup, especially when looking at the ones that are not strictly internal.

ypnos:
Just FYI, I plan to prepare some alternative code so we can discuss that one too as well. I envision to have a central place and handling for/of types that are exposed/written by the user through strings.

t-wissmann:
I think for parsing, throwing exceptions is fine. `std::optional` is only c++14, right?

dnnr:
> `std::optional` is only c++14, right?

It appears to be C++17: https://en.cppreference.com/w/cpp/utility/optional

PR #154, opened 2018-11-18
Fix missing overrides

blarz:
Found by clang-6.0


blarz:
Together with #150 this makes it possible to compile the project with a modern clang.

PR #155, opened 2018-11-18
Integrate clang-tidy into Makefile

dnnr:


dnnr:
For now, the purpose would be running it manually in order to identify potential problems to be fixed.

In the long term, I'd recommend integrating it into the CI build so that any (reasonable) failing checks turn the build red.

I originally planned on proposing both steps at once but now that Simon has shown interest in fixing issues, I wanted to enable everyone to see the list of pending issues for themselves.

I hope that as soon as some of the "low-hanging fruit" type checks pass, we could run it in Travis using that subset of checks. From there, we could gradually activate more checks as the issues get fixed. So it's basically the same approach as for additional compiler warnings.

ypnos:
Just a note on this, there is a pitfall in starting to fix all kind of generic low-level issues/warnings now while we still need to rewrite huge chunks of code. If you do it to a certain extreme, you will find yourself wasting a lot of time.

I think right now, a top-down approach is more helpful even for new contributors to winterbreeze like you and Simon. Even if you want to start "small" (no architectural stuff), it is probably more pressing right now to get rid of argc/argv method signatures, char* string manipulation, and Glib usage. There is some horror stuff there in the code that will just be completely gone with ArgList vs argc/argv. Like using memset on strings etc. So I would rather start from there.

I also wonder, Thorsten, should we convert the stuve pad into issues?

ypnos:
Btw. I love (a) warnings-free code, (b) this clang-tidy integration. I'm just saying we are getting at a point where we might want to actively manage ressources a bit.

dnnr:
@ypnos 
I see your point, and I think I've already stumbled across the issues that you're referring to. I've put them aside at first, because it wasn't immediately clear to me how to fix them.

I agree that we shouldn't make the mistake of spending time on polishing code parts that are already slated for removal/major rework. But I should also stress that most of my recent PRs have been quite mechanical. So sparing code parts that will be reworked soon wouldn't have made those PRs much easier, I think.

One main reason for me to prioritize those changes I'm currently doing, instead of working on the more impactful stuff is that, in my experience, there is a certain invariant: *It'll never be any easier than it is right now.* Unless, of course, you're about to throw away half you code base anyway.

And aside from all that... I really enjoy cleaning up. I just can't help it. Even though there is no feature progress, it doesn't feel like time wasted to me. As long as it doesn't inconvenience the reviewers, of course.

dnnr:
> As soon as you like to get your hands a bit more dirty while cleaning up, maybe we should have a chat and lay out a plan about these things I mentioned so we can get some progress there.

Sure thing! I would certainly appreciate seeing your existing plans and known problems laid out publicly as issues in GitHub.

t-wissmann:
All right, then let's merge this and see that we convert the pad to issues. :-)

PR #156, opened 2018-11-18
Use bool literals instead of casting integers

dnnr:


ypnos:
You are setting the initial value of several attributes to `true` while they were 0 before. In error?

dnnr:
Oh boy, I've messed up the search&replace here, I'm terribly sorry. Fix is coming up!

dnnr:
Should be correct now.

t-wissmann:
Great, thanks!

PR #160, opened 2018-11-18
Remove older compilers from .travis.yml

dnnr:


dnnr:
This PR is an attempt at opening the discussion that didn't happen in #151.

In case nobody feels strongly enough about it to discuss it, *this* is what I would suggest and prefer over the current state, which is a rather lengthy build (~14 min).

t-wissmann:
I don't have strong opinions about this. I'm fine with having 'an old compiler on a old system' and 'a new compiler on a new system'. So I'm fine with your suggestion. Maybe I'm even fine with dropping the gcc-5 on 16.04.
Other opinions?

dnnr:
Alright then! I've remove that one as well. :+1:

PR #161, opened 2018-11-18
Add missing member inits, replace some ctors

dnnr:


t-wissmann:
Can you change the initializers to the `=` notation? we use it already in src/client.h except for constructors with more than one argument.
Or is there any reason to stick to `member {default}`?

t-wissmann:
I agree with @ypnos, and I don't see how `=` vs `{}` affects our case of class member initialization. I've experimented with src/client.h, and the following initializations are accepted:
```
    bool        visible_ = 234232343;
    bool        visible_ {234232343};
    bool        visible_ = 234232343.1;
    bool        visible_ {234232343.1};
```
So I'd vote for `=` because I find the syntax slightly more intuitive. But I don't have strong feelings here, as long as it's constent throughout the code.

ypnos:
This is what I get:

```
bool int_ = 234232343;
bool inta_ {234232343}; // error, cannot be narrowed
bool double_ = 234232343.1; // warning, will evaluate to true
bool doublea_ {234232343.1}; // error, cannot be narrowed
char doubleb_ = 234232343.1; // warning, out-of-range conversion is undefined
char doublec_ {234232343.1}; // error, cannot be narrowed
```

As we compile with -Werror, the only difference I see is in the case of `int` → `bool`. But indeed there is a difference.

t-wissmann:
It seems to fail with clang but not with gcc: https://travis-ci.org/herbstluftwm/herbstluftwm/builds/458409975

ypnos:
Yes. Odd behavior by GCC.

t-wissmann:
@ypnos: Do you prefer `=` or brace-syntax in hlwm's source?

The-Compiler:
My $0.02 even though I'm not very active currently: I prefer using brace-syntax everywhere when writing C++, because it usually comes with less surprises.

ypnos:
In the case of member initializations at declaration time I prefer = for readability. Everywhere else I would use {}.

t-wissmann:
Thank you for your comments, @ypnos and @The-Compiler.
I'm fine with both options. But in order to make some progress, I'd say we stay with `=` for member initialization, because the truncation issues are of minor relevance for us (we have nearly no floating point numbers in the code).

PR #162, opened 2018-11-19
Remove obsolete suppression for -Winvalid-offsetof

dnnr:
There are no warnings for this in the codebase.


t-wissmann:
I don't get why it's obsolete. If we don't want to use offsetof than, then we should keep the warning. Is this implication wrong? Or is the premise wrong?

dnnr:
Maybe I'm mistaken, but as I understand it, the flag *disables* a warning related to offsetof.

t-wissmann:
You're right! I missed the `no` in the actual patch after reading the commit message.

PR #163, opened 2018-11-24
Use $(CXX) as a default value for LDXX

dnnr:
Using g++ circumvents explicit compiler selection via the CXX
environment variable. Using $(CXX) as fallback ensures that the selected
compiler is used for linking.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -16,7 +16,7 @@ endif
 # FLAGS
 CC ?= gcc
 LD = $(CC)
-LDXX ?= g++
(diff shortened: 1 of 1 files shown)


ypnos:
While we're at it, I wonder why there is a default setting for CC but no default setting for CXX given.

dnnr:
Good question. I'm not even sure if the line I added isn't simply the default for GNU make. For now, my priority was to fix a clear bug.

PR #164, opened 2018-11-24
Add some tests for tag handling

dnnr:


tests/test_tags.py
```diff
@@ -0,0 +1,23 @@
...
+    assert hlwm.get_attr('tags.1.name') == 'foobar'
+
+
+def test_move_focused_client_to_new_tag(hlwm):
+    hlwm.callstr('add foobar')
+    (proc, winid) = hlwm.create_client()
+
+    hlwm.callstr('move foobar')
+
+    # TODO: Assert that foobar now has 1 client
```
dnnr: @t-wissmann How would I do these assertions? The tag's `client_count` attribute and the client's `tag` attribute don't seem to exist yet/anymore. Am I missing something?

ypnos: These should come back (tags.*.clients.count and clients.*.tag) as symlinks, right?

dnnr: So there is no way to do these assertions right now? Would you be fine with merging this like it is then? Better than nothing, I suppose.

t-wissmann: Yes, they don't exist yet, and I'm fine with merging this for the time being.

PR #166, opened 2018-11-24
Replace global function w/ TagManager::removeTag()

dnnr:


dnnr:
This includes changes from other PRs (#163, #164, #165), which should be merged first!

src/tag.cpp
```diff
@@ -89,62 +89,6 @@ HSTag* find_unused_tag() {
...
-        buf.push_back(client);
-    });
-    for (auto client : buf) {
-        stack_remove_slice(client->tag()->stack, client->slice);
-        client->setTag(target);
-        stack_insert_slice(client->tag()->stack, client->slice);
-        ewmh_window_update_tag(client->window_, client->tag());
-        target->frame->insertClient(client);
-    }
-    tag->frame = shared_ptr<HSFrame>();
```
dnnr: I've dropped this step. At first it just seemed pointless to me now that I'm `delete`ing the tag, but this also caused a segfault in `applyLayout()` when (a couple of calls down) the frame got accessed once more.

dnnr: I meant to say that I didn't carry the assignment of `tag->frame` to the new implementation because it seemed superfluous to me, given the (now) explicit deletion of the tag object.

Besides that, I discovered that the existing code even segfaulted *due to* this assignment. So maybe this was a bug anyway, but that's not what motivated me to drop the assignment.

That said, I'm not 100% sure whether there was some importance to that assignment that I might have missed, and that's why I wanted to draw your attention to it.

t-wissmann: I see, thanks! I guess the double assignment simply came from an inconsistent / unfinished migration to the `HSTag` class.

dnnr:
I've included #173 into this change, so that one should be merged first

src/tagmanager.cpp
```diff
@@ -65,6 +65,73 @@ int TagManager::tag_add_command(Input input, Output output) {
...
+    }
+    input.shift();
+    auto tagNameToRemove = input.front();
+    input.shift();
+    auto targetTagName = input.empty() ? get_current_monitor()->tag->name : input.front();
+
+    auto targetTag = find(targetTagName);
+    auto tagToRemove = find(tagNameToRemove);
+
+    if (tagToRemove == nullptr) {
```
ypnos: Here I would prefer `if (!tagToRemove)`.

t-wissmann:
I don't care about `if (!p)` vs `if (p == nullptr)`, so we can merge :)

src/tagmanager.cpp
dnnr: I was under the impression that it's actually not recommended (anymore) to use pointers as logical expressions. I fail to find any evidence for that, so I'll gladly go with your suggestion.

PR #168, opened 2018-11-24
Common interface and behavior for communicating basic types with the user

ypnos:
This is a follow-up to the discussion in #152.

Ideas behind this approach:

- Strip logic of type conversions out of Attributes, put it into types.h instead
- Have a common interface for said conversions such as the types we have conversions for can be easily made into attributes (e.g. Rectangle, which can be used to have the geometry of a client as attr)
 - Converting to/from user is always done in a consistent manner (parsing throws std::exceptions)
 - Conversions are defined where the types are defined, e.g. Color conversions are defined in x11-types.h where Color is defined

Open issue: Color constructor that takes a string currently fails silently (as it did previously) and falls back to black. I would suggest to let the exception fall through instead.

### DIFF
--- a/src/arglist.cpp
+++ b/src/arglist.cpp
@@ -8,8 +8,7 @@ ArgList::ArgList(const std::initializer_list<std::string> &l)
 ArgList::ArgList(const ArgList::Container &c) : c_(std::make_shared<Container>(c)) { reset(); }
 
 ArgList::ArgList(const std::string &s, char delim) {
-    c_ = std::make_shared<Container>();
-    split(*c_, s, delim);
+    c_ = std::make_shared<Container>(split(s, delim));
     reset();
 }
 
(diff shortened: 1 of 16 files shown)


t-wissmann:
Looks great, thanks!

Regarding the constructor of Color: I'm not sure whether it's good design to have a constructor that throws an exception. But that's something for the future.

ypnos:
I looked it up and the consensus seems to be that throwing from a constructor is fine.
I take it you prefer the silent fail (fallback to black) of the constructor then?

t-wissmann:
If we allow the constructor to fail than this means that every object with a color attribute has a constructor that possibly throws an exception. I think I prefer to change Color in such a way that the constructor can't fail (regardless whether failing is silent or an exception).

ypnos:
@dnnr Do you understand why the test fails? https://travis-ci.org/herbstluftwm/herbstluftwm/jobs/460294393

PR #170, opened 2018-11-25
Enforce 2 seconds timeout on hc processes in tests

dnnr:


ypnos:
Could you please shortly describe what's behind this so I can learn from it :)

dnnr:
> Will it fail the test instead of waiting indef. for a hanging herbstclient call?

Exactly. This happened when I ran into a segfault in hlwm, which led to hc waiting for a response forever. I considered introducing a timeout into hc instead, but ultimately dismissed that approach.

I suppose for the test execution it's fair to expect all hc invocations to conclude within 2 seconds.

Would you like me to explain this in a comment as well?

PR #172, opened 2018-11-26
Enable GCC sanitizers for test execution

dnnr:


dnnr:
The build for this will fail until the existing leaks are fixed.

ypnos:
It seems there is still a memory leak in IPC client, but we also got a segfault in Attribute.

t-wissmann:
I've found the segfault in Attribute, which also causes the memory leak in ipc client. But I still need to understand why this causes a memory leak in ipc-client. (there is no leak if hlwm does not crash).

t-wissmann:
The crash is fixed in ecd881ff0290ddd3cfc75adac8352f542e232b5d and the memory leak is fixed in b847740bdff79559fa6fd1d749575623ca72ad0b, both directly in the winterbreeze branch.

dnnr:
> Can we also add an option for the sanity checks in the config.mk? We can also keep them activated per default before winterbreeze is merged into master.

You mean so that you can easily used them on your local machine, not just in Travis, right?

That certainly makes sense. I didn't do it like that because I didn't want to pollute the Makefile(s) unnecessarily.

t-wissmann:
Yes, to have the checks already in local tests. We can also add this once we have a cleaner Makefile setup. Didn't you (@dnnr or @ypnos) have an idea how to shorten the CMakeLists.txt? (the CMakeLists.txt is only in master at the moment).

ypnos:
Idk, you know that I advocated for using a more high level build system than what we have. On the other hand, I really hate CMake. It might be unavoidable in the end though, so I suggest these resources on how to use CMake in a more sane way (not the insane 2.x way):

* https://www.slideshare.net/DanielPfeifer1/cmake-48475415
* https://cliutils.gitlab.io/modern-cmake/

PR #174, opened 2018-11-27
Fixes in the split command

t-wissmann:


t-wissmann:
This does not make the split command perfect, but I think it's a significant improvmenet of the code and behaviour.

src/layout.cpp
```diff
@@ -960,23 +961,19 @@ bool HSFrameLeaf::split(int alignment, int fraction, int childrenLeaving) {
 }
 
 
-int frame_split_command(int argc, char** argv, Output output) {
+int frame_split_command(Input input, Output output) {
     // usage: split t|b|l|r|h|v FRACTION
-    if (argc < 2) {
+    std::string cmd, splitType, strFraction;
+    if (!input.read({ &cmd, &splitType})) {
```
dnnr: Inconsistent whitespace (within braces)

t-wissmann:
I've fixed the whitespace. I also propose to merge this by a merge commit without squashing the commits, because they cover different issues around the split-command.

PR #175, opened 2018-11-27
Clarify method signature

ypnos:
Make clear that this does not alter the object

### DIFF
--- a/src/arglist.cpp
+++ b/src/arglist.cpp
@@ -66,7 +66,7 @@ bool ArgList::read(std::initializer_list<std::string*> targets) {
     return true;
 }
 
-ArgList ArgList::replace(const std::string& from, const std::string& to) {
+ArgList ArgList::replaced(const std::string& from, const std::string& to) const {
     int i = 0;
     vector<std::string> new_list(size());
     for (auto v : *this) {
--- a/src/arglist.cpp
+++ b/src/arglist.cpp
@@ -18,10 +18,6 @@ ArgList ArgList::operator+(Container::difference_type shift_amount) {
     return ret;
 }
 
-std::string ArgList::operator[](size_t idx) {
-    return c_->operator[](idx);
-}
-
 ArgList::Container ArgList::split(const std::string &s, char delim) {
     Container ret;
     std::stringstream tmp(s);
--- a/src/arglist.h
+++ b/src/arglist.h
@@ -47,7 +47,7 @@ struct ArgList {
     bool read(std::initializer_list<std::string*> targets);
     /** construct a new ArgList with every occurence of 'from' replaced by 'to'
      */
-    ArgList replace(const std::string& from, const std::string& to);
+    ArgList replaced(const std::string& from, const std::string& to) const;
     // the first element without any shifts.
     std::string command() const {
         if (c_->begin() != c_->end()) {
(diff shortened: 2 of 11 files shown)


ypnos:
Did we change to a PR-only model on winterbreeze? So I cannot push even small fixes like this because no travis. Is it the intention?

What is the workflow then, should I do every snippet in a PR or maybe have aggregated PRs that are developed alongside the feature/patch one is actually working on?

ypnos:
Another issue I found in ArgList is the `operator[]` which ignores the internal shift. It is rarely used, to put it mildly. I am not sure but it looks like I will introduce the first usage right now (a command with variable number of arguments that should be processed one after another).
Please vote @dnnr, @t-wissmann 

1. Remove operator[], alternatives are to use begin()+end() or toVector()
2. Change behavior to start counting at `begin_`

t-wissmann:
I've changed the branch properties that one can only merge PRs for which the checks pass. But I did not intend to force a PR-only model.

t-wissmann:
I'm voting to remove `operator[]` because both its possible semantics (with or without shift) will cause confusion.

ypnos:
Yes I think using the iterators instead works pretty well:

```
-    RectangleVec buf(argc);
-    for (int i = 0; i < argc; i++) {
-        buf[i] = Rectangle::fromStr(argv[i]);
-    }
+
+    RectangleVec buf;
+    for (auto &i : input) {
+        buf.push_back(Rectangle::fromStr(i));
+    }
```

t-wissmann:
And I've changed your role, so maybe you're now allowed to push to winterbreeze directly. If this does not work either, I think we should drop the branch protection and just keep it as an informal rule to merge only PRs for which the checks pass.

dnnr:
> So I cannot push even small fixes like this because no travis. Is it the intention?

The repo configuration change was my suggestion, as a mitigation for what happened in #164. But as Thorsten already said, we did not intend to prohibit direct pushes by project members.

However, I'd strongly recommend a workflow where every change goes through a PR with enforced automated checks. That said, I don't see any point in trying to *make* that happen for hlwm, so I'm just saying this as a suggestion to consider it.

> What is the workflow then, should I do every snippet in a PR or maybe have aggregated PRs that are developed alongside the feature/patch one is actually working on?

From my experience, I'd avoid aggregate PRs. Instead, one should keep CI build durations short in order to enable and encourage small-scale PRs.

ypnos:
In general I am a huge fan of a stringent PR workflow due to the benefit of a four-eyes.

For *this* project I don't see a huge advantage of travis over running tox manually on my machine. Also *this* branch is still alpha quality and there are lots of small fixes that probably don't warrant the overhead.

So maybe we should agree to small fixes without PR to the winterbreeze branch at the discretion of the author, whereas somewhat bigger changes need a PR, and completely lock-down master?

At the same time, should we agree to a policy that one only accept other people's PRs, or introduce a LGTM system?

dnnr:
> For _this_ project I don't see a huge advantage of travis over running tox manually on my machine. 

Well, the advantage I see is that a Travis build does a considerably more than I am usually doing locally, and this difference will probably increase over time (multiple compilers, ASAN, clang-tidy, coverage, etc.). From my personal experience, I actually enjoy the workflow of doing the quick superficial verification locally (like "does it build with *my* compiler?" and "does my IDE indicate warnings?") and then rely on the automated build to be much more thorough than that.

> So maybe we should agree to small fixes without PR to the winterbreeze branch at the discretion of the author, whereas somewhat bigger changes need a PR, and completely lock-down master?

Full ACK to that.

ypnos:
Yes, Travis already showed its fruits since we have it here, and I look forward to enabling more features. When I currently push I don't go out for dinner before the tests are done :)

ypnos:
Btw I will further improve on my last commit here.

PR #176, opened 2018-11-28
Turn add_monitor_command into manager method

dnnr:


src/monitormanager.cpp
```diff
@@ -213,6 +214,71 @@ void MonitorManager::removeMonitor(HSMonitor* monitor)
...
+            output << input.command() <<
+                ": Tag \"" << input.front() << "\" is already being viewed on a monitor\n";
+            return HERBST_TAG_IN_USE;
+        }
+    }
+
+    input.shift();
+    std::string name;
+    if (!input.empty()) {
+        name = input.front();
```
t-wissmann: I think you can even improve this by writing:
```
if (input.read({ &name })) {
```
And similarly earlier where the rectangle and the tag name is read.

tests/test_monitor.py
```diff
@@ -20,6 +20,59 @@ def test_add_monitor(hlwm):
     assert hlwm.get_attr('monitors.1.name') == 'monitor2'
 
 
+def test_cannot_add_monitor_without_free_tag(hlwm):
+    call = hlwm.call('add_monitor', '800x600+40+40', check=False)
+
+    assert call.returncode != 0
+    assert call.stderr.endswith(' not enough free tags\n')
```
t-wissmann: I'm not sure how important the error message looks like, but it's important that the command fails (we also don't care about the returncode as long as it is non-zero here). If you agree, you can boil the test down to:
```
hlwm.fails('add_monitor', '800x600+40+40')
assert hlwm.get_attr('monitors.count') == '1'
```

src/monitormanager.cpp
ypnos: I really think this is a bad interface (from a C++ point of view) but it serves is purpose to extract several arguments at once. For the single argument case I already intended to propose a `pop()` method that combines front with shift.

tests/test_monitor.py
ypnos: If there is only one error case it doesn't matter, I assume the rationale behind this approach is to ensure the right error is given.

dnnr: For tests like these I strongly recommend verifying the failure cause. It's easy to mask errors in the code or simply mistakes in the test if you just expect it to fail. For example, if you have a typo in `'add_monitor'` in the test, it will be green.

That said, I would prefer checking the error message in a more precise fashion than string matching. I shortly look into checking the actual return code, but for that I'd have to evaluate `ipc-protocol.h`, which doesn't even hard-code them.

As an intermediate solution, we could duplicate the exit code definitions in the tests.

src/monitormanager.cpp
dnnr: I've changed it for the monitor name, but it wasn't immediately clear to me how to use `read()` for the rect and tag without complicating things with additional local variables. Am I missing something?

PR #177, opened 2018-11-29
Turn HSStack into a class

dnnr:


dnnr:
The common base for this branch is already outdated due to the merge of #166. I'll resolve the conflicts later, but I think you can already review it if you like.

src/stack.h
```diff
@@ -39,10 +39,31 @@ typedef struct HSSlice {
...
+    bool is_layer_empty(HSLayer layer);
+    void clear_layer(HSLayer layer);
+
+    // returns the number of windows in this stack
+    int window_count(bool real_clients);
+    void to_window_buf(Window* buf, int len, bool real_clients, int* remain_len);
+    void restack();
+    Window lowest_window();
+
     GList*  top[LAYER_COUNT];
```
ypnos: While we are at it, could we also get rid of this?
How is this used? Like an std::list? Or even std::forward_list?

src/stack.cpp
```diff
@@ -197,7 +191,7 @@ static struct HSTreeInterface stack_nth_child(HSTree root, size_t idx) {
     return intface;
 }
 
-static size_t stack_child_count(HSTree root) {
+static size_t stack_child_count(std::shared_ptr<HSStack> root) {
```
ypnos: What's the deal with this method? Should we get rid of it?

src/stack.cpp
```diff
@@ -294,7 +288,7 @@ static void slice_to_window_buf(HSSlice* s, struct s2wb* data) {
     }
 }
 
-void stack_to_window_buf(HSStack* stack, Window* buf, int len,
+void HSStack::to_window_buf(Window* buf, int len,
```
ypnos: I suggest fixing the signature of this method, replacing remain_len argument with a return value.

src/tag.h
```diff
@@ -22,7 +22,7 @@ class HSTag : public Object {
     Attribute_<bool>         floating = {"floating", false};
     Attribute_<std::string>  name = {"name", {}};   // name of this tag
     int             flags;
-    struct HSStack* stack;
+    std::shared_ptr<HSStack> stack;
```
ypnos: Why is stack a pointer and not a simple member object? For the tree interface?

dnnr:
As a general comment, there are a lot of potential improvements that occurred to me while working on this change, but I've purposely postponed them for later PRs to keep the scope of this one fixed.

The only exception I'd make here is the plain removal of obsolete functions, but I couldn't decide on that on my own.

So far, we have:
 * Fix typos in method names (at least I assume r`aise_slide` was supposed to be `raise_slice`)
 * Apply our method naming scheme (camelCase instead of snake_case)
 * Remove HS prefix (introduce namespace in same step or separately?)

src/stack.cpp
dnnr: Would you be okay with doing this separately?

src/stack.cpp
dnnr: I'm not sure if it was *this* method, but there were some that appeared entirely unused. As I said in my global comment, please advise what can be simply removed.

dnnr:
> I will not nag you with "btw" stuff that is not already part of the PR in the future.

Don't get me wrong, I'm happy to hear these suggestions! It's just that I'll probably prefer addressing them in separate changes unless they're clearly related to the same PR.

src/stack.cpp
t-wissmann: If `stack_child_count` is not explicitely used, we can remove it.

src/stack.cpp
t-wissmann: This will return a `vector<Window>` at some point, but I think the current version is fine for the time being.

src/stack.h
t-wissmann: It's simply a `vector<HSSlice*> top[LAYER_COUNT]`.

src/tag.h
t-wissmann: Yes, it can simply be a member of HSTag

src/stack.cpp
ypnos: What kind of windows would this vector be comprised of?

t-wissmann:
I think the current PR is fine as it is. The stack.h and .c were written quite strictly in the style of object oriented C, like it is done in the linux kernel or in glib. And the current PR sincerely translates this into object orientation in C++, which I find is a very reasonable step and which is not to be confused with redesigning the data structures. Thank you :-)

I fully agree with @ypnos's comment on HS-prefix and namespaces. The "btw"-comments are good to be kept in mind for future PRs. So IMO, this PR can be merged once the conflicts are resolved.

src/stack.cpp
t-wissmann: The windows in the stack. The monitor stack made things much more complicated than they need to be. Once the monitor stack is a simple vector (as a member of MonitorManager), the stack is a array (of length LAYER_COUNT) of HSSlice-vectors.

src/stack.cpp
dnnr: From what I can tell: It's called in `slice_child_count`, which in turn is referenced only by `layer_nth_child`, which in turn is referenced only by `stack_nth_child`, which in turn is referenced only by `slice_nth_child`.

So I wouldn't claim to understand any of that, but it sure looks like a bowl of unused spaghetti to me.

t-wissmann: I think the `slice_nth_child` is used or will be used for the tree interface (that's the interface that is passed to the tree printing algorithm, e.g. in the `herbstclient stack` command)

PR #178, opened 2018-11-30
Read from ArgList using operator>>()

t-wissmann:
Add an istream-style interface to ArgList as a replacement for read(),
e.g. via

    std::string& arg1, arg2;
    input >> arg1 >> arg2;

The operator bool() tells whether all previous reading operations
succeeded. This can be combined:

    std::string& optional_arg;
    if (input >> optional_arg) {
        ...
    }

Since the bool() operator makes automatic casts to ints ambigous, I had
to remove the operator+(), which has been used only once anyway.


t-wissmann:
What do you think, @dnnr, @ypnos? Do you prefer this over the other `input.read({ &arg1, &arg2 })` interface?
At least, this interface now resembles the use of cin / istream.

src/monitormanager.cpp
```diff
@@ -217,58 +217,42 @@ void MonitorManager::removeMonitor(HSMonitor* monitor)
 int MonitorManager::addMonitor(Input input, Output output)
 {
     // usage: add_monitor RECTANGLE [TAG [NAME]]
-    if (input.size() < 2) {
+    std::string cmd, rectString, tagName, monitorName;
+    input >> cmd >> rectString;
```
dnnr: While using the "old" interface, I actually liked that the command was (in a way) treated as a first-class citizen in ArgList, being available via the `command()` method, mainly for use in error messages. I didn't mind having to shift past the command before retrieving the first argument, but with the new stream operator, you even need a local variable to put it in.

So I'd suggest finishing what ArgList already started and turning it into an object that holds *a command and its arguments*, instead holding *a command and a list containing the command followed by its arguments*.

src/monitormanager.cpp
```diff
@@ -217,58 +217,42 @@ void MonitorManager::removeMonitor(HSMonitor* monitor)
...
-    HSTag* tag;
-    if (input.empty()) {
-        tag = find_unused_tag();
-        if (!tag) {
-            output << input.command() << ": There are not enough free tags\n";
-            return HERBST_TAG_IN_USE;
-        }
-    } else {
-        tag = find_tag(input.front().c_str());
+    HSTag* tag = NULL;
```
dnnr: Use `nullptr` for a stupidly tiny bit of type safety?

src/monitormanager.cpp
```diff
@@ -217,58 +217,42 @@ void MonitorManager::removeMonitor(HSMonitor* monitor)
...
+            return HERBST_TAG_IN_USE;
         }
-        if (find_monitor_by_name(name.c_str())) {
-            output << input.command() <<
-                ": A monitor with the same name already exists\n";
+    }
+    // TODO: error message on invalid rectString
+    auto rect = Rectangle::fromStr(rectString);
+    if (input >> monitorName) {
+        auto error = isValidMonitorName(monitorName);
```
dnnr: Good call extracting this into a method!

dnnr:
Overall, I quite like it! This kind of operator overloading makes me a bit uneasy (wasn't even aware of `operator bool()`) for the usual reasons, but I'd certainly prefer it over the `read()` interface.

If we go forward with this change, I'd strongly advocate for adopting the interface everywhere and soon, followed by removal of the then-obsolete methods in `ArgList`. Unless I'm missing some other use-cases of course...

src/monitormanager.cpp
t-wissmann: I agree. So I propose to make Input a class that inherits from ArgList and has command name as a member. Actually, it is not clean that ArgList has a command() method.

src/monitormanager.cpp
t-wissmann: I've changed this to `nullptr`.

t-wissmann:
I fully agree that we should migrate the other code soon and then remove the `read()`-function.

t-wissmann:
I've just removed `ArgList::read()` and made `Input` a class that inherits from `ArgList`. I had to make a little trick in order to fulfill all of the following assumptions:
1. `Input::operator>>` does not give the command name
1. `ArgList::operator>>` gives all elements, since it is not aware of any command name
1. Current code using `operator[]` and `shift()` does not break.
This is why `Input::operator>>` calls `shift()` when being called the first time.

dnnr:
This might be a dumb question, but why *shouldn't* `ArgList` be aware of the command name? Is it used for anything else than parsing commands (where the first argument is always the command itself)?

src/types.cpp
```diff
@@ -0,0 +1,28 @@
...
+    ArgList::operator>>(val);
+    return *this;
+}
+
+std::string Input::command() const {
+    if (c_->cbegin() != c_->cend()) {
+        return *c_->cbegin();
+    }
+    // access the first element without any shifts.
+    if (c_->begin() != c_->end()) {
```
ypnos: duplication

ypnos:
Other than the name might suggest, ArgList is also used in recursive traversal of the object tree.

I really like your work thorsten, however I suggest a different approach:
1. Have an optional command argument to the constructor and treat it separately from the list
2. Do away with all the initial shifts

I volunteer for that.
I think otherwise we run in an overly complicated design and unintuitive separation between ArgList and Input just for the sake of not having to go through the code once.
I also removed operator[] so where are the users :-)

dnnr:
> Other than the name might suggest, ArgList is also used in recursive traversal of the object tree.

Hoo boy, that sounds oddly... unrelated. But I guess there are reasons for that.

Apart from that, I tend to share your doubts about the inheritance.

I haven't really thought it through (and don't know anything about this tree traversal usage), but could it be that the code sharing isn't really worth it here? Instead of cramming the common denominator into a single class, maybe we should just have two separate (albeit similar) classes that are exactly fit for their particular purposes?

ypnos:
No I think both usages do exactly what you expect / what ArgList is designed for: They pass the list of strings from one method to another and each eats away one or several elements.

The name "ArgList", although a very nice name by itself from a bilingual standpoint, is the only part that does not completely fit.

**Another alternative to the command name handling:**
Don't provide the command at all. Use a stringstream buffer as Output. If the return is an error code, output it with prepended command name. If not, just output.

I like this one best so far. Because why should every method prepend the command name? It is just code duplication and it will be a burden if we decide to change the output format in any way.

src/types.cpp
t-wissmann: is fixed.

t-wissmann:
> This might be a dumb question, but why _shouldn't_ `ArgList` be aware of the command name? Is it used for anything else than parsing commands (where the first argument is always the command itself)?

See condition 1 above.

> Other than the name might suggest, ArgList is also used in recursive traversal of the object tree.

Yes. ArgList is simply a list of strings with useful traversal

> I really like your work thorsten, however I suggest a different approach:
> 
>     1. Have an optional command argument to the constructor and treat it separately from the list

I don't think this is a good idea. This is precisely what you do with inheritance actually:

* ArgList is not aware of any command name, it simply is a **list of arguments**.
* An Input is a Comamnd name plus a list of arguments. Hence Input has a member commandName and inherits ArgList.

> I volunteer for that.
> I think otherwise we run in an overly complicated design and unintuitive separation between ArgList and Input just for the sake of not having to go through the code once.
> I also removed operator[] so I hope there is no current code using that one :-)
I want to make small changes. This PR is one change. Going through all commands and migrating them from operator[] to >> is another change.

ypnos:
I don't mind an extra class, but I do mind the hackery. So an extra class that (always) stores a command name is fine, but then it should be `Input` from the start, constructed as such with the command as an extra argument, no weird conversion from `ArgList` to `Input` and initial shift.

t-wissmann:
> I don't mind an extra class, but I do mind the hackery. So an extra class that (always) stores a command name is fine, but then it should be `Input` from the start, constructed as such with the command as an extra argument, no weird conversion from `ArgList` to `Input` and initial shift.

The initial is a **temporary** solution until all commands are migrated from `operator[]` to `operator>>`. Once this is done the list of arguments will no longer hold the command name and there will be no initial shift.

ypnos:
Yes I like it as an initial step, and as I said I would volunteer to go through the code and fix this right away. There is no operator[] so I don't specifically understand what you mean with that though.

Because you ignored my alternative suggestion I assume you're not a big fan?

t-wissmann:
> ### Another alternative to the command name handling:
> 
> Don't provide the command at all. Use a stringstream buffer as Output. If the return is an error code, output it with prepended command name. If not, just output.
> 
> I like this one best so far. Because why should every method prepend the command name? It is just code duplication and it will be a burden if we decide to change the output format in any way.
> 
> Drawback: Can't use the same method for several commands, or need to pass an extra argument accordingly

Here one needs to be carefull with things like `chain`:
```
chain , echo hello , fail message
```
assume that `echo hello` succeeds with output `hello` and that `fail message` fails with output `message`. Then `fail: ` is prepended to `message` and additionally `chain` will be prepended so we get the overall output:
```
chain: hello
fail: message
```
Alternatively, one can provide something like:
```
output.error << "message"
```
and so the `error` does the prepending of the command name.

PR #179, opened 2018-11-30
Improve test suite output if hlwm crashes

t-wissmann:
Explicitly check for undesired termination of herbstluftwm in the fixtures when a client action times out and print an appropriate error message.

In the hlwm_process fixture, don't wait() for herbstluftwm if the process has been wait()'ed already. If a client action times out, the herbstluftwm process has already been wait()'ed and so we avoid a second unnecessary failed assertion.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,29 +10,32 @@ import pytest
 
 GIT_ROOT = os.path.join(os.path.abspath(os.path.dirname(__file__)), '..')
 
-
 class HlwmBridge:
 
     HC_PATH = os.path.join(GIT_ROOT, 'herbstclient')
 
-    def __init__(self,display):
+    def __init__(self, display, hlwm_process):
         self.next_client_id = 0;
         self.env = {
             'DISPLAY': display,
         }
+        self.hlwm_process = hlwm_process
 
     def callstr(self, args, check=True):
         return self.call(*(args.split(' ')), check=check)
     def call(self, *args, check=True):
         assert args
         str_args = [ str(i) for i in args]
-        proc = subprocess.run([self.HC_PATH, '-n'] + str_args,
-                              stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                              env=self.env,
-                              universal_newlines=True,
-                              # Kill hc when it hangs due to crashed server:
-                              timeout=2
-                              )
+        try:
+            proc = subprocess.run([self.HC_PATH, '-n'] + str_args,
+                                  stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                  env=self.env,
+                                  universal_newlines=True,
+                                  # Kill hc when it hangs due to crashed server:
+                                  timeout=2
+                                  )
+        except subprocess.TimeoutExpired:
+            self.hlwm_process.investigate_timeout('calling ' + str(str_args))
         print(list(args))
         print(proc.stdout)
         print(proc.stderr, file=sys.stderr)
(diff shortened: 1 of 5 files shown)


t-wissmann:
And of course it's intentional that the test suite fails :) We only need to revert on of the commits before merging. I also propose to not squash the commits such that we keep the diffs in the history.

tests/conftest.py
```diff
@@ -10,29 +10,35 @@
...
+        try:
+            proc = subprocess.run([self.HC_PATH, '-n'] + str_args,
+                                  stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+                                  env=self.env,
+                                  universal_newlines=True,
+                                  # Kill hc when it hangs due to crashed server:
+                                  timeout=2
+                                  )
+        except subprocess.TimeoutExpired:
+            timeout = True
```
dnnr: Couldn't you do the `client_timeout` call within the `except` block and completely remove the `timeout` variable?

tests/conftest.py
```diff
@@ -10,29 +10,35 @@
 
 GIT_ROOT = os.path.join(os.path.abspath(os.path.dirname(__file__)), '..')
 
-
 class HlwmBridge:
 
     HC_PATH = os.path.join(GIT_ROOT, 'herbstclient')
 
-    def __init__(self,display):
+    def __init__(self,display,hlwm_process):
```
dnnr: Not sure if you leave them out on purpose, but for what it's worth, PEP 8 prescribes spaces after these commas (there are more occurrences within this change)

tests/conftest.py
```diff
@@ -91,14 +122,25 @@ def hlwm_process(tmpdir):
 
     proc = subprocess.Popen([bin_path, '--verbose'], env=env,
                             stdout=subprocess.PIPE)
+    hlwm_proc = HlwmProcess(proc)
```
dnnr: So the `HlwmProcess` wraps around the return value of `Popen`, right? Why not integrate the `Popen` into its constructor? I was a bit confused by this contraction between the class name and its functionality.

tests/conftest.py
dnnr: Ah, I suppose you're doing this to avoid the pytest noise with nested exceptions ("During handling of" yadda yadda)?

Maybe this should be explained in a short comment, because it's not obvious at all, and only necessary for a specific type of pytest traceback reporting (which happens to be the default...).

tests/conftest.py
```diff
@@ -52,26 +58,51 @@ def create_client(self):
         command = ['xterm', '-hold', '-class', wmclass, '-e', 'true']
         # enforce a hook when the window appears
         self.call('rule', 'once', 'class='+wmclass, 'hook=here_is_'+wmclass)
-        # run process that wait for the hook
+        # the following is still racy because xterm might connect to X before
```
dnnr: Could we start a single, persistent `hc --idle` along with hlwm to avoid this race condition?

tests/conftest.py
```diff
@@ -52,26 +58,51 @@ def create_client(self):
...
-
+    return HlwmBridge(display,hlwm_process)
+
+
+class HlwmProcess:
+    def __init__(self,proc):
+        self.proc = proc
+
+    def client_timeout(self,reason):
+        """if a client complains about a timeout, investigate the
```
dnnr: By "client" you're specifically referring to herbstclient, right? Either way, a clarification would probably be good, given the potential for confusion within this file.

tests/conftest.py
The-Compiler: That's a Python 3 feature, not something specific to pytest - I'd expect you to see it with `--tb=plain` as well. However, you could do `raise Exception(...) from None` inside `client_timeout` to suppress it.

t-wissmann: Right @dnnr, I want to avoid the noise here. Not sure if I understand @The-Compiler correctly, but I will try it!

tests/conftest.py
t-wissmann: I'll change it!

tests/conftest.py
t-wissmann: Right, that sounds more reasonable. I just tried to make the diff as small as possble.

tests/conftest.py
t-wissmann: I'm using the `from None` now.

tests/conftest.py
```diff
@@ -10,29 +10,40 @@
...
     HC_PATH = os.path.join(GIT_ROOT, 'herbstclient')
 
-    def __init__(self,display):
+    def __init__(self, display, hlwm_process):
         self.next_client_id = 0;
         self.env = {
             'DISPLAY': display,
         }
+        self.hlwm_process = hlwm_process
+        self.hc_idle = subprocess.Popen(\
```
dnnr: This trailing backslash is not necessary

tests/conftest.py
```diff
@@ -10,29 +10,40 @@
...
+                    [self.HC_PATH, '--idle', 'rule', 'here_is_.*'],
+                    bufsize=1, # line buffered
```
dnnr: Did you confirm that this really enables line-buffering? The docs make it sound like it requires `universal_newlines=True` to work.

dnnr: On the other hand, line buffering is probably the default anyway?

tests/conftest.py
```diff
@@ -52,26 +63,81 @@ def create_client(self):
...
         # once the window appears, the hook is fired, and the --wait exits:
-        hc_wait.wait(2)
-        winid = hc_wait.stdout.read().decode().rstrip('\n').split('\t')[-1]
+        winid = self.wait_for_window_of(wmclass)
 
         return SimpleNamespace(proc=proc, winid=winid)
 
+    def wait_for_window_of(self, wmclass):
+        """Wait for a rule hook of the form "here_is_" + wmclass """
+        # we don't need to investigate the rule here
```
dnnr: I'm not sure what is meant here by "investigate the rule"

tests/conftest.py
```diff
@@ -52,26 +63,81 @@ def create_client(self):
...
         # enforce a hook when the window appears
         self.call('rule', 'once', 'class='+wmclass, 'hook=here_is_'+wmclass)
-        # run process that wait for the hook
-        hc_wait = subprocess.Popen([self.HC_PATH, '--wait', 'rule', 'here_is_'+wmclass],
-                    env=self.env,
-                    stdout=subprocess.PIPE)
-        # start the process...
-        proc = subprocess.Popen(command,env=self.env)
+        proc = subprocess.Popen(command, env=self.env)
         # once the window appears, the hook is fired, and the --wait exits:
```
dnnr: `and the --wait exits` isn't accurate anymore here, is it?

tests/conftest.py
t-wissmann: I agree with both of your comments and the documentation https://docs.python.org/3/library/subprocess.html really sounds a bit contradictory. So maybe lets set `universal_newlines=True` to be on the save side?

tests/conftest.py
t-wissmann: We don't need to read the second argument of the hook and don't need to check that is indeed equals `"here_is_" + wmclass`. But we need to check this once we create clients simultaneously.

t-wissmann:
I've incorporated your comments and this branch is now ready for a proper (non-squash) merge.

tests/conftest.py
dnnr: Seems reasonable :+1:

PR #180, opened 2018-11-30
Add remaining tag attributes

t-wissmann:
Re-implement tag attributes from master (frame_count, client_count, curframe_windex, curframe_wcount). This allows detailed testing of the tiling commands.

### DIFF
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -1,43 +1,48 @@
+import pytest
 
-def explode_and_remove(hlwm,create_client,number_clients):
-    assert number_clients >= 2, "explode behaves as auto for one client"
+@pytest.mark.parametrize("number_clients", [0,1,2])
+def test_single_frame_layout(hlwm,create_client,number_clients):
     for i in range(0, number_clients):
         create_client()
     assert hlwm.get_attr('tags.0.curframe_windex') == '0'
+    assert int(hlwm.get_attr('tags.0.client_count')) == number_clients
     assert int(hlwm.get_attr('tags.0.curframe_wcount')) == number_clients
-    assert hlwm.get_attr('tags.0.frame_count') == '1'
+
+@pytest.mark.parametrize("number_clients", [2,3])
+def test_explode(hlwm,create_client,number_clients):
+    assert number_clients >= 2, "explode behaves as auto for one client"
+    for i in range(0, number_clients):
+        create_client()
     hlwm.callstr('split explode')
-    # half of the clients in the upper frame, the other in the lower
     assert hlwm.get_attr('tags.0.curframe_windex') == '0'
     number_upper = (number_clients + 1) // 2
     assert int(hlwm.get_attr('tags.0.curframe_wcount')) == number_upper
     assert int(hlwm.get_attr('tags.0.client_count')) == number_clients
     assert hlwm.get_attr('tags.0.frame_count') == '2'
-    # after calling 'focus down' for number_upper-times, the focus switches to
-    # the lower frame
-    for i in range(0, number_upper):
-        assert int(hlwm.get_attr('tags.0.curframe_windex')) == i
-        assert int(hlwm.get_attr('tags.0.curframe_wcount')) == number_upper
-        hlwm.callstr('focus down')
-    # after the last 'focus down' we fall into the other frame
-    assert hlwm.get_attr('tags.0.curframe_windex') == '0'
-    assert int(hlwm.get_attr('tags.0.curframe_wcount')) \
-            == number_clients - number_upper
-    assert hlwm.get_attr('tags.0.frame_count') == '2'
 
-    hlwm.call('remove')
+@pytest.mark.parametrize("number_clients", [0,1,4])
+def test_remove(hlwm,create_client,number_clients):
+    for i in range(0, number_clients):
+        create_client()
+    hlwm.callstr('split explode')
+    hlwm.callstr('remove')
     assert int(hlwm.get_attr('tags.0.curframe_wcount')) == number_clients
     assert int(hlwm.get_attr('tags.0.client_count')) == number_clients
     assert hlwm.get_attr('tags.0.frame_count') == '1'
     # TODO: reasonably handle focus, e.g. to have
     #assert hlwm.get_attr('tags.0.curframe_windex') == '2'
 
-def test_explode_remove_2(hlwm,create_client):
-    explode_and_remove(hlwm,create_client, 2)
-
-def test_explode_remove_3(hlwm,create_client):
-    explode_and_remove(hlwm,create_client, 3)
-
-def test_explode_remove_4(hlwm,create_client):
-    explode_and_remove(hlwm,create_client, 4)
+@pytest.mark.parametrize("number_clients", [3,4])
+def test_focus_wrap(hlwm,create_client,number_clients):
+    assert number_clients >= 2, "explode behaves as auto for one client"
+    for i in range(0, number_clients):
+        create_client()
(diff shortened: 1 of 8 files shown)


tests/test_layout.py
```diff
@@ -0,0 +1,43 @@
+
+def explode_and_remove(hlwm,create_client,number_clients):
```
dnnr: I'm not really sure what's going on here, so I can't make concrete suggestion yet, but it looks like this test case could be much improved.

For unit tests, there's a proven pattern called *Arrange-Act-Assert*. It encourages to focus on a single thing being tested per case, by first *arranging* the preconditions for the doing the thing, then *acting* out the thing, then *asserting* the expected outcome. Those three phases are commonly separated by blank lines. Having a lot of lines per phase, or growing a desire for additional blank lines is usually a good indicator for a vaguely scoped test case or lack of a useful fixture.

It often leads to quite repetitive test code (see some of my recently added tests), so I know it's hard to live by. But ultimately, it yields much more maintainable and understandable tests, in my experience.

The-Compiler: Let me also note pytest's [parametrization feature](https://docs.pytest.org/en/latest/parametrize.html) - often, if you have very similar tests (which only differ in data/arguments/...), that can make things much nicer.

t-wissmann: I agree, this is way to long and it contains two tests. I can try to separate it and I'll definitely use the pytest parameter feature!

dnnr:
I've added more powerful fixtures to simplify these kinds of tests (#185). See a0cdf7fbabbd6787c5339af8b5eed4e9176f15ba for a commit that adopts the new fixtures into your `test_layout.py`

tests/test_layout.py
dnnr: This is *much* better already! (see my new comment, though)

ypnos:
Great and important work! What I don't like about this PR:

1. You introduce a new macro at a weird place, to
2. initialize members in the constructor instead of the class declaration,
3. so they can access very short getter methods,
4. which are private so not really getters as one would understand them

So I wonder if maybe one or several of these steps should be taken:

1. Initialize the attributes in the declaration and set the getters explicitely in the constructor
2. Use lambdas at that point instead of class methods,
3. Or rename the methods `getFrameCount()` → `frameCount()` etc and make them public
4. If you want to expose the fact that these methods are not 'for free', use 'compute' prefix instead of 'get' (however I believe it's not warranted)

Apart from it looks great to me and I am also happy if you want it merged without taking any of the proposed steps.

t-wissmann:
I see your point, but should we use `this` in the initialization in the declaration? (This is a question both on the level of C++-features and of style). If it's valid C++, we can agree on this.

Why is it bad to have the method private? In my opinion it's always desirable to keep the public surface as small as possible. I'm fine with prefixing the names with `compute…` instead of `get…` to make it clearer what they do.

t-wissmann:
By the way, it is not possible at the moment to initialize a dynamic attribute in the constructor, because there is no constructor without arguments. I regard this as a feature, because one can not forget to initialize a dynamic attribute.

ypnos:
No, I don't think you can use ´this` in the declaration, that's why I suggest to set the methods explicitely in the constructor like we do with validators or signal connections.

For the private methods I just think it is unnecessary boilerplate if there is no other use of them and I would do lambdas instead. Or if you say they might be handy at other places I suggested public getters.

t-wissmann:
I've just made two of the getters lambdas in the constructor and renamed the other two names from `get…` to `compute…`.

I have a strong opinion about the following convention: If something can be asserted via the type system of a language (e.g. "A DynAttribute_ always can be read because it always holds a getter") then we should always assert this via the type system and not via tests.

Another strong opinion: functions/methods are there to structure the code and cut it into reasonable pieces. Re-using a certain piece of code in different places is one reason to introduce methods, but not the only reason. This is why it's reasonable to have `computeFrameCount` and `computeClientCount` as a separate method in HSTag and not as a lambda in the HSTag-constructor, even though it is used only once.

ypnos:
I think your first opinion is basically true but not practical. For example, the assertion "An Attribute always works as expected" is already violated if you forget to put it into the wireAttributes() call. So technically any Attribute is already broken at construction time without further work. So in this specific case I see no real benefit in polluting the way attributes are initialized (Attribute_ vs DynAttribute_). (Note: we could always have a default getter that returns an empty string if that makes you more happy?)

I heartfully agree with your second strong opinion and I also like your resolution. The question of which amount of code warrants a new method is obviously not written in stone.

t-wissmann:
> I think your first opinion is basically true but not practical. For example, the assertion "An Attribute always works as expected" is already violated if you forget to put it into the wireAttributes() call. So technically any Attribute is already broken at construction time without further work.

The object of the (Dyn)Attribute class itself works fine (i.e. there is no way of letting it crash by calling `operator()`), it simply isn't available via herbstclient when the wireAttributes() is forgotten.

> So in this specific case I see no real benefit in polluting the way attributes are initialized (Attribute_ vs DynAttribute_). (Note: we could always have a default getter that returns an empty string if that makes you more happy?)

Where is the pollution? I think it's perfectly fine to initialize normal attributes in the member definition and DynAttributes in the constructor initalizer list; I don't regard this as unclean or inconsistent.

I also don't like the default case because then we have to define some default value in every instance of the Converter<>-template. 

> I heartfully agree with your second strong opinion and I also like your resolution. The question of which amount of code warrants a new method is obviously not written in stone.
Sure!

ypnos:
A default getter is easy:
`DynAttribute_() : getter_([] () -> T { return {};}) {}`
I'm sorry I don't follow what you mean with Converter?

In any case we have to answer to question whether we want to have it (the ability to wire a proper getter after construction) or not. Maybe @dnnr may cast a final vote on this or are you vetoing it?

t-wissmann:
I mean this by converter: https://github.com/herbstluftwm/herbstluftwm/blob/winterbreeze/src/types.h#L20

> A default getter is easy:
> `DynAttribute_() : getter_([] () -> T { return {};}) {}`

Right, I did not think about `{}`. But does this work if the underlying class does not have a default constructor? (e.g. `Direction` or `Color~)

> I'm sorry I don't follow what you mean with Converter?
> 
> In any case we have to answer to question whether we want to have it (the ability to wire a proper getter after construction) or not. Maybe @dnnr may cast a final vote on this or are you vetoing it?
Yes, lets wait for @dnnr's comment.

ypnos:
You would be surprised, both of these classes are default-constructable :-)

t-wissmann:
I am sorry in which direction the discussion in this PR went. This PR originally introduced four dynamic attributes in the class HSTag and added tests for it. My original tests were horrible, but I hope they're better now, thanks to your help.

However, the latter discussion concerns the orginization of the framework of Attributes and DynAttributes. I assume this was triggered because I introduced the macro
```
#define BIND_THIS(X) ([this]() { return this->X(); })
```
in `attribute_.h` (not sure if one regards this as "changes the framework of (Dyn)Attributes")

This is why I've changed the initialization of the dynamic attributes such that it looks exactly like in e.g. [monitor.cpp](src/monitor.cpp).

If there aren't any further concerns with the diff of the present PR, I propose to merge this PR and continue the framework discussion in subsequent PRs or Issues. What do you think, @ypnos?

ypnos:
I also believe the discussion could have been more productive because there have been quite a few misunderstandings and I think it should have gone much smoother in a different setting. Certainly it is not always the best to drag along PRs with lengthy fundamental discussions and I already took note of that.

I still think the PR improved since its introduction and is fine to merge as well.

PR #181, opened 2018-11-30
HSStack::top: Turn array of GLists into vector

dnnr:


dnnr:
I broke something here, will look later... PR closed for now.

PR #182, opened 2018-12-02
Turn HSSlice::layer into an std::set

dnnr:


src/stack.cpp
```diff
@@ -55,7 +54,7 @@ HSSlice* slice_create_window(Window window) {
 
 HSSlice* slice_create_frame(Window window) {
     HSSlice* s = slice_create_window(window);
-    s->layer[0] = LAYER_FRAMES;
+    s->layers.insert(LAYER_FRAMES);
```
dnnr: I'm don't why the old code didn't do a `layer_count = 1` here. Was this a bug?

ypnos:
Looks good to me and I guess it was a bug; count would be not initialized at all.

src/stack.cpp
t-wissmann: `slice_create_window` calls `slice_create` which sets `layer_count = 1`. So for `slice_create_frame` the `layers` consists of only `LAYER_FRAMES`. However now, `layers` contains both `LAYER_FRAMES` and `LAYER_NORMAL`. Since frames should not be in the `LAYER_NORMAL` you should do `s->layers.clear()` here first.

t-wissmann: And maybe the `layers` should be empty in both cases and only be filled in `slice_add_layer`.

dnnr: Ah! I missed that part. I'll take another look...

dnnr: Fixed now

PR #183, opened 2018-12-02
HSStack::top: Turn array of GLists into vector

dnnr:


dnnr:
This includes #182, which should be merged first.

t-wissmann:
I've added some tests for the stack and for the raise command in c32733029c499beba35c58e90d89df7f131c3c9d and boldly pushed them to the winterbreeze branch to have them for the old stack implementation and in all PRs

PR #184, opened 2018-12-02
Make ChildByIndex assume ownership of children

dnnr:
This plugs the remaining leaks as detected by ASAN: When destroying
a MonitorManager or TagManager, no dtor takes care of deleting their children.

IMO, this patch only qualifies as a temporary solution, and it probably only works
because MonitorManager and TagManager happen to be the only subclasses
of ChildByIndex.

For what it's worth, it does facilitate enabling ASAN for test execution in the CI build.


t-wissmann:
Great, thanks! I'm fine with temporary solutions like this because it is progress :-)

PR #186, opened 2018-12-03
Clean up and extend test cases for stack

dnnr:
In addition to some (mostly cosmetic) cleanup, this also verifies the
stack more strictly, which confirms the absence of a subtle,
almost-merged mistake in slice_create_frame().


t-wissmann:
Great, thanks!

PR #188, opened 2018-12-05
Pass Attribute owner directly to constructor

t-wissmann:
This new approach makes the signature for Attribute_ and DynAttribute_ very uniform. HSTag and HSMonitor are already adapted to the new framework. Other classes still use the old constructors that are kept for the time being for compatibility reasons.

### DIFF
--- a/src/attribute_.h
+++ b/src/attribute_.h
@@ -17,14 +17,33 @@ public:
     // escalated to the user.
     using Validator = std::function<std::string(T)>;
 
-    Attribute_(const std::string &name, const T &payload)
-        : Attribute(name, false)
-        , payload_ (payload)
-    {
-    }
+    /** The constructors for Attribute_ and also for DynAttribute_ follow
+     * the a common pattern:
+     *
+     *   - The first argument is the owner object
+     *
+     *   - The second argument is the name displayed to the user
+     *
+     *   - The third argument the reading behaviour of the attribute:
+     *     * for a plain Attribute_ this is simply the initial value
+     *     * for a DynAttribute_ this is a member of the owner that returns the
+     *       value of the attribute.
+     *
+     *   - The fourth argument describes the writing behaviour of the
+     *     attribute: If the fourth argument is absent, the attribute is
+     *     read-only.
+     *     * The fourth argument for an Attribute_ is a member of owner that
+     *       validates a new value of the attribute and returns an error
+     *       message if the new value is not acceptable for this attribute
+     *       (E.g. because a name is already taken or does not resepct a
+     *       certain format).
+     *     * The fourth argument of a DynAttribute_ is a member of owner that
+     *       internally processes the new value (e.g. parsing) and returns
+     *       an error message if the new value is not acceptable.
+     */
 
     //! A read-only attribute of owner of type T
-    Attribute_(const std::string &name, Object* owner, const T &payload)
+    Attribute_(Object* owner, const std::string &name, const T &payload)
         : Attribute(name, false)
         , payload_ (payload)
     {
@@ -36,17 +55,24 @@ public:
 
     //! A writable attribute of owner of type T
     template <typename Owner>
-    Attribute_(const std::string &name, Owner* owner, const T &payload,
-               std::string(Owner::*validator)(T))
+    Attribute_(Owner* owner, const std::string &name, const T &payload,
+              std::string(Owner::*validator)(T))
         : Attribute(name, true)
-        , payload_ (payload)
         , validator_(validator)
+        , payload_ (payload)
     {
         // the following will call Attribute::setOwner()
         // maybe this should be changed at some point,
         // e.g. when we got rid of Object::wireAttributes()
         owner->addAttribute(this);
-        writeable_ = true;
+    }
+
+    //! Deprecated constructor, that will be removed and only remains for
+    //compatibility
+    Attribute_(const std::string &name, const T &payload)
+        : Attribute(name, false)
+        , payload_ (payload)
+    {
     }
 
     // set the method called for validation of external changes
@@ -151,7 +177,7 @@ public:
     {
         hookable_ = false;
     }
-    DynAttribute_(const std::string &name, Object* owner, std::function<T()> getter)
+    DynAttribute_(Object* owner, const std::string &name, std::function<T()> getter)
         : Attribute(name, false)
         , getter_(getter)
     {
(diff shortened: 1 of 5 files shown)


t-wissmann:
Please also note the [documentation on attributes](src/attribute_.h) describing the idea behind the constructor parameters for `Attribute_` and `DynAttribute_`.

src/attribute_.h
```diff
@@ -17,6 +17,70 @@ class Attribute_ : public Attribute {
...
+     *     read-only.
+     *     * The fourth argument for an Attribute_ is a member of owner (or
+     *       lambda) that validates a new value of the attribute and returns an
+     *       error message if the new value is not acceptable for this
+     *       attribute (E.g. because a name is already taken or does not
+     *       resepct a certain format).
+     *     * The fourth argument of a DynAttribute_ is a member of owner (or
+     *       lambda) that internally processes the new value (e.g. parsing) and
+     *       returns an error message if the new value is not acceptable.
+     */
```
t-wissmann: Please also note the documentation on attributes describing the idea behind the constructor parameters for Attribute_ and DynAttribute_.

src/attribute_.h
```diff
@@ -17,6 +17,70 @@ class Attribute_ : public Attribute {
...
+    Attribute_(Object* owner, const std::string &name, const T &payload,
+              Validator validator)
+        : Attribute(name, true)
+        , validator_(validator)
+        , payload_ (payload)
+    {
+        // the following will call Attribute::setOwner()
+        // maybe this should be changed at some point,
+        // e.g. when we got rid of Object::wireAttributes()
+        owner->addAttribute(this);
```
dnnr: Just a thought: Maybe `registerAttribute` would be an (even) more descriptive name for that method?

Apart from that, I agree with the comment above. I think/hope `setOwner` as a public method can be fully replaced with initializers in the attribute ctors.

t-wissmann: Yes, and then it's clear that `registerAttribute` does not call `Attribute::setOwner`.

src/attribute_.h
dnnr: The explanation was very useful to me, but in order to be a proper docstring for Doxygen, it should (afaik) be turned into a docstring for the class and/or the individual constructors. Probably a combination of both, to avoid repeating everything for each ctor.

src/tag.h
```diff
@@ -16,9 +16,9 @@ class HSTag : public Object {
     HSTag(std::string name, Settings* settings);
     ~HSTag() override;
     std::shared_ptr<HSFrame>        frame;  // the master frame
-    Attribute_<unsigned long> index = {"index", 0};
-    Attribute_<bool>         floating = {"floating", false};
-    Attribute_<std::string>  name = {"name", {}};   // name of this tag
+    Attribute_<unsigned long> index;
+    Attribute_<bool>         floating;
+    Attribute_<std::string>  name;   // name of this tag
```
dnnr: While at it, could you fix or remove the member name indentation here? (I'd prefer removing it)

src/attribute_.h
t-wissmann: I think this should be attached to the header file and be included in the class description for both Attribute_ and DynAttribute_, I just don't know how to express this syntactically.

src/tag.h
t-wissmann: What about the indentation in the constructor calls in the monitors.cpp file?

dnnr: Same. I don't see the benefit of that style, to be honest. (There are some very rare cases where I think otherwise, but this is not one of them.)

src/tag.cpp
```diff
@@ -34,34 +34,18 @@ void tag_destroy() {
...
 HSTag::HSTag(std::string name_, Settings* settings)
-    : frame_count("frame_count", std::bind(&HSTag::computeFrameCount, this))
-    , client_count("client_count", std::bind(&HSTag::computeClientCount, this))
-    , curframe_windex("curframe_windex",
+    : index(this, "index", 0)
+    , floating(this, "floating", false, [](bool){return "";})
+    , name(this, "name", name_, &HSTag::validateNewName)
+    , frame_count(this, "frame_count", &HSTag::computeFrameCount)
+    , client_count(this, "client_count", &HSTag::computeClientCount)
+    , curframe_windex(this, "curframe_windex",
```
dnnr: I don't see anything about this lambda variant in itself, but I'm wondering:
Maybe it would be better to reduce the number of ctor variants in `DynAttribute_` by dropping this one? In this particular case, `HSTag` would need to provide additional getters that are basically just these one-liners.

It's clearly just a trade-off, but I feel that additional getters would "cost less" in terms of complexity than this ctor variant does at *both* its point of usage (i.e., right here) *and* within `DynAttribute_`.

src/tag.h
t-wissmann: The benefit is: it emphasizes their similarities and their differences. This means, that thanks to the indentation, I see directly, that `this` is passed to every attribute as the first parameter.

dnnr: Hold on, I was looking at the member declarations in monitor.h, not the ctor calls. Sorry. Okay, the ctor calls aren't that bad, so I actually *do* see the benefit there. I wouldn't do like that personally, but I also won't complain about it. ;-)

src/tag.cpp
dnnr: I just realized that you've actually already been doing it exactly like that before, just with `BIND_THIS`. So you've probably got your reasons... feel free to ignore me here.

t-wissmann: I'm fine with introducing additional getters.

However in settings.cpp, e.g. for `border_width`, we will still need the `DynAttribute_` constructor that has a lambda as a getter and as a setter because in the settings, just a member pointer is not enough.

dnnr: I guess it's true, but I don't fully understand why. Is it because there are a lot of attributes that are very similar and you don't want to turn this single `getIntAttr()` helper into `n` almost identical getters, right?

t-wissmann: Right. I want to re-use `getIntAttr()` for every int typed setting. But this isn't possible anymore if we drop the constructor accepting the lambdas. This issue is also connected with the issue that I need to use C++-style member function pointers and can't use lambdas in the templated constructor of `DynAttribute_`.

dnnr: Hmmm. I'm wondering if there is a good workaround. But that's not really relevant here. So I see why we should keep the lambda constructor variant. At the same time appreciate that you're open to adding simple getters instead of using lambdas wherever possible.

I feel that in the concrete case of HSTag, readability of the initializer list would be improved by sticking to the non-lambda variant.

t-wissmann: My conclusion is that we need both but should use the non-lambda whereever possible.

PR #189, opened 2018-12-05
Use running_clients fixture in layout tests

dnnr:


dnnr:
Unless you plan on doing a merge commit here (not squash it), no rebase should be necessary because I've merged the (then current) winterbreeze branch into this PR branch. What is shown in the *Files changed* tab is what ultimately counts.

I'll rebased it anyway, just in case.

t-wissmann:
Right! Sorry for the extra work!

PR #190, opened 2018-12-05
Rename fails() to call_xfail(), use it much more

dnnr:


dnnr:
Oh yeah, that was actually almost the entire point of that change, I just forgot it. Thanks!

t-wissmann:
I don't like the name `call_xfail` because it does not really speak by itself.

Can we have a more advanced syntax of the form "*some command* fails with *some message*"?
I know this is a bit unclean and untested:
```
    def fails(self, *args):
        call = self.call(*args, check=False)
        assert call.returncode != 0
        call.with = lambda message: assert call.stderr == message
        return call
```
But one can write:
```
   hlwm.fails('sdfsdf') \
       .with('error: Command "sdfsdf" not found')
```
or alternatively we can design something that:
```
   hlwm.expectfail('error: Command "sdfsdf" not found')
       .call('sdfsdf')
```

ypnos:
Or we have sth along the lines of:

```
hlwm.call('foo').fail()
hlwm.call('foo again').failWith("Never call foo on me again!")
hlwm.call('bar').succeedWith("I liked bar. I still like bar.")
ret = hlwm.call('bar').succeed
# do more stuff with ret
```

t-wissmann:
> Or we have sth along the lines of:
> 
> ```
> hlwm.call('foo').fail()
> hlwm.call('foo again').failWith("Never call foo on me again!")
> hlwm.call('bar').succeedWith("I liked bar. I still like bar.")
> ret = hlwm.call('bar').succeed()
> # do more stuff with ret
> ```

I've also considered this originally but discarded the idea because we have to write `.succeed()` in all normal calls that are expected to succeed.

ypnos:
Yes if you only want to see if it succeeds you write more. If you want a simple test on the resulting output, you write less. If you want to do additional stuff with the output, again you write more.

I don't think adding `succeed()` there is not too bad because it also helps the reader. I wonder about the ratio of `succeed()` and `succeedWith()` usages we would end up in the tests. Because the latter would be a really nice feature.

The-Compiler:
> because it does not really speak by itself

FWIW "xfail" is what pytest (and other frameworks?) uses to say "expected to fail"

dnnr:
> I don't like the name `call_xfail`

I was already kind of expecting that because I wasn't too happy with the name either.

Just for sake of completeness, let me explain my motivations here. My goal was to:
 1. reduce the call interface to two kinds of expectations: *must succeed* and *must fail* (I don't see any use case for "don't care")
 2. get rid of the `check` argument because:
   2.1. the expectation is too important to be hidden away like that at the end of the argument list
   2.2. the name `check` doesn't fit the new meaning
 3. not have a method  called `fails()` because that is just really opaque

I chose `call_xfail` precisely because of what @The-Compiler already said: It's established terminology in the context of tests.

That said, I'm intrigued by the fluent style interface that you suggested.

My slight doubt with that is that it further weakens the separation between `Act` and `Assert`. Admittedly, my change is already guilty of the same thing by integrating the return code expectation into the call. So maybe (me making) this argument is pointless.

In that regard, the only thing I'd have to say in defense of my change is that it reduces the duplication that you have with `c = hlwm.call(..., check=False); assert c.returncode != 0`. That is basically first saying that the call doesn't *need* to succeed, followed by asserting that it *must fail*. That of course only is an advantage over the current state, not over your suggestions.

ypnos:
For me this PR cleans up quite some test code and adds only a few lines. I would merge it and if somebody wants to come up with a more clever solution have it as a separate PR.

We could not include the name change but I agree with @dnnr that 'fails' is not necessarily better than 'call_xfail'. While 'call_xfail' sounds weird, it is supposed to be intuitive to people that write tests.

t-wissmann:
I agree, let's merge this!

PR #191, opened 2018-12-05
 Integrate client launching into HlwmBridge 

dnnr:
While the fixture was kind of cool, tying the lifetimes of all clients
directly to that of the WM is just much simpler for all parties
involved.


dnnr:
This PR includes #189, which should be merged first.

t-wissmann:
I hope I resolved the conflicts correctly...

PR #192, opened 2018-12-05
Fully match stderr messages in tests

dnnr:


dnnr:
I think this make sense given that we're probably soon going to change the way we put together error messages. In hindsight, I was probably misguided anyway in just matching an arbitrary message suffix.

PR #193, opened 2018-12-06
 Separate concerns in tag tests more clearly

dnnr:


dnnr:
Only the topmost commit on this branch is relevant to this PR, the rest of it matches #191, which should be merged first.

tests/test_tags.py
```diff
@@ -7,27 +10,36 @@ def test_add_tag(hlwm):
     hlwm.callstr('add foobar')
 
     assert hlwm.get_attr('tags.count') == '2'
+    assert hlwm.get_attr('tags.1.client_count') == '0'
+    assert hlwm.get_attr('tags.1.client_count') == '0'
+    assert hlwm.get_attr('tags.1.curframe_wcount') == '0'
+    assert hlwm.get_attr('tags.1.curframe_windex') == '0'
```
dnnr: I'm not 100% sure how to separate the tests for *tags*, *layouts* and *frames* (should the latter be another group of tests?). These `curframe_` attributes seem to fit into multiple topics.

tests/test_tags.py
```diff
@@ -7,27 +10,36 @@ def test_add_tag(hlwm):
...
+    assert hlwm.get_attr('tags.1.frame_count') == '1'
+    assert hlwm.get_attr('tags.1.index') == '1'
     assert hlwm.get_attr('tags.1.name') == 'foobar'
 
 
-def test_move_focused_client_to_new_tag(hlwm, create_client):
```
dnnr: FYI, this old test case is a good example of how tests can slowly degenerate into testing more than one thing.

At first sight, it makes sense to have these "before vs. after" assertions. But the first two assertions are actually testing the (non-existent) effects of `add` on `client_count`, which was decidedly not the purpose of the test. Once this is extracted into its own test case, this one can solely focus on how `move` affects the `client_count`.

There are of course counter-examples, where it is preferable to walk through a length "story" with a continuous stream of actions and assertions, but I don't think is one of them.

PR #195, opened 2018-12-08
Turn GQueue for rules into an std::vector

dnnr:


t-wissmann:
We don't have tests for rules yet. But if the rules are seriously broken, then the whole client creation will fail, so this looks good to me :)

PR #196, opened 2018-12-08
Turn HSRule struct into a class

dnnr:


dnnr:
This PR includes #195 and needs to be updated after that once was merged.

t-wissmann:
> This PR includes #195 and needs to be updated after that once was merged.

Does it really? If this contains #195, why do we have conflicts then?

t-wissmann:
Can you take a look at the conflicts?

Thanks for cleaning this file up :-)

dnnr:
I probably caused those conflicts by rebasing one of the branches. I've fixed them, but had to include #199 into this PR, so that one should go first (but it should anyway). I'll merge again afterwards.

PR #200, opened 2018-12-08
Turn rule label into std::string 

dnnr:


src/command.h
```diff
@@ -108,6 +108,7 @@ int call_command_no_output(int argc, char** argv)
 int list_commands(int argc, char** argv, Output output);
 int complete_command(int argc, char** argv, Output output);
 
+void try_complete(const char* needle, std::string to_check, Output output);
```
dnnr: This function already existed, it was just not declared in the header. I assume it was okay to add it.

t-wissmann: Sure

PR #203, opened 2018-12-08
Change call interface to optimize for common cases

dnnr:


tests/test_root_commands.py
```diff
@@ -1,28 +1,30 @@
...
+
 def test_sprintf(hlwm):
     cnt = hlwm.get_attr('tags.count')
     wmname = hlwm.get_attr('settings.wmname')
-    printed = hlwm.callstr('sprintf X %s/%s tags.count settings.wmname echo X').stdout
-    hlwm.call_xfail('sprintf X %s/%s tags.count settings.wmname')
-    hlwm.call_xfail('sprintf X %s/%s tags.count')
+    printed = hlwm.call('sprintf X %s/%s tags.count settings.wmname echo X').stdout
+    hlwm.call_xfail('"sprintf X %s/%s tags.count settings.wmname"')
+    hlwm.call_xfail('"sprintf X %s/%s tags.count"')
```
t-wissmann: These `call_xfail` calls should be without the quotes, because it checks the error handling for lacking arguments or commands

t-wissmann: Of course, this was wrong before. And I think it supports your point to check also error messages

dnnr: Oops. I didn't look at those tests too closely and just assumed that it is actually supposed to be single string. I'll fix it.

PR #206, opened 2018-12-08
Switch to CMake build

ypnos:
Use simple file VERSION that contains the version string
Let determine full version string (in case of git with commit hash and
current date) by the build system

Workflow as described in INSTALL:
mkdir build; cd build; cmake ..; make

Use cmake -G'Ninja' for faster building (then call ninja)

The following is currently not supported:
· make *-nodoc; switch off documentation in (c)cmake instead
· make tar; TBD?
· make uninstall; could do 'xargs rm < install_manifest.txt' instead
· make www/cleanwww; TBD?
· make clang-tidy (or cppcheck, etc); can be easily added
· release.sh; needs to be adapted to VERSION file and see make tar


ypnos:
With clang we get the narrowing warnings → errors again that were conveniently turned off in the old makefile. The question is do we want to fix the narrowing problems? We had discussions about them before.

CMakeLists.txt
```diff
@@ -0,0 +1,88 @@
...
+## macros & functions (may be moved to cmake/ and included)
+# get project dependencies
+macro (get_external_deps) # – macro for scope
+	find_package(X11 REQUIRED)
+	include(FindPkgConfig)
+	pkg_check_modules(GLIB2 REQUIRED glib-2.0)
+endmacro()
+
+# extract git commit hash and append to project version
+function (set_gitversion)
```
dnnr: Why not put nicely encapsulated functions like this into some kind of utilities `.cmake` file?

src/CMakeLists.txt
```diff
@@ -0,0 +1,83 @@
+## The 'herbstluftwm' executable ##
+
+add_executable(herbstluftwm main.cpp)
+install(TARGETS herbstluftwm DESTINATION ${BINDIR})
+
+# additional sources – core/architectural stuff
+target_sources(herbstluftwm PRIVATE
+	arglist.cpp	arglist.h
```
dnnr: I don't know which viewer setting (for how to visualize tabs) you need for this to *not* look broken, but in every other case, it's not great (including GitHub). I recommend spaces here, at least between the words. In fact, I don't see why a single space wouldn't be just fine here.

CMakeLists.txt
ypnos: The comment above states that we could do it. However it would add a directory and a file and an include statement and in the end I didn't do it as right now we have only a couple of those.

dnnr: Sorry for not noticing the comment. Yeah, I'm fine keeping it that way. Maybe it'll work better with some more separation (like: not have this function figure out *and* set the version variable).

ypnos: Yes though, returning a value from a function? Too sophisticated for CMake! All you can do is pass another argument that tells the function which variable to set, and then have awkward code in the set statement. So yeah, that's that.

.travis.yml
```diff
@@ -59,6 +62,7 @@ jobs:
           dist: xenial
           before_install:
               - CC=clang-7 CXX=clang++-7
+              - CXXFLAGS="$CXXFLAGS -Wno-sign-compare -Wno-narrowing"
```
dnnr: I like applying suppressions like these only where they are really needed (i.e., here in this file, for this compiler). :+1: 

But it looks to me that GCC also complains about narrowing, but the build doesn't fail because `-Werror` wasn't carried over from the Makefile?

CMakeLists.txt
```diff
@@ -0,0 +1,103 @@
...
+endfunction()
+
+# add custom build flags to the DEBUG build type
+function (add_debugflags TARGET)
+	set(FLAGS
+		-Og
+		-pedantic -Wall -Werror
+		-Wno-sign-compare -Wno-narrowing -Wno-error=cast-function-type
+		-fsanitize=address,leak,undefined
+		-fno-omit-frame-pointer -fuse-ld=gold)
```
ypnos: I took gold linker option from travis. But I am not so sure if we should put that here and if this option is supported by clang. What's the rationale?

INSTALL
```diff
@@ -1,13 +1,14 @@
...
-Copyright 2011-2013 Thorsten Wißmann. All rights reserved.
+Copyright Thorsten Wißmann and contributors. All rights reserved.
 
 This software is licensed under the "Simplified BSD License".
 See LICENSE for details.
 
 ==== Requirements ====
 Build dependencies:
-    - build-environment (gcc/other compiler, make)
+    - build environment (e.g. gcc, clang; make, ninja)
```
t-wissmann: Does the `;` have a meaning or is it a typo? Or do you want to express something like:
```
   (e.g. gcc or clang, make or ninja)
```

CMakeLists.txt
t-wissmann: I also don't know which option is for which compiler. @dnnr?

.travis.yml
t-wissmann: But we also need these flags for gcc on xenial, right? At least this is why the build fails currently.

t-wissmann:
Thanks a lot! I think we can merge this once the CI is running again, and we can deal with the open build features (release, tar, www) later.

CMakeLists.txt
dnnr: IIRC, I had to use gold when using gcc-8.2 with sanitizers, due to some bug in GCC (missing space between arguments passed to the linker). Generally, gold is better™, but there is no inherent need for it. So if you get it work without it, it's fine.

INSTALL
dnnr: I think you can just say "make" here because it's the default generator in CMake. I'd expect anybody that knows of ninja, to just try if it works instead of assuming otherwise just because it's not mentioned her.

PR #208, opened 2018-12-09
Add more tests for unrule and completion

dnnr:


dnnr:
This doesn't include #203, but makes use of it's changes. So the CI build should just pass when rerunning it after #203 was merged.

t-wissmann:
The error message of unrule needs to be adjusted in the test case:
```
_________________________ test_remove_nonexistent_rule _________________________
hlwm = <conftest.HlwmBridge object at 0x7f7ea1a3d240>
    def test_remove_nonexistent_rule(hlwm):
        call = hlwm.call_xfail('unrule nope')
    
>       assert call.stderr == 'Couldn\'t find any rules with label "nope"'
E       assert 'Couldn\'t find rule: "nope"' == 'Couldn\'t fin...h label "nope"'
E         - Couldn't find rule: "nope"
E         + Couldn't find any rules with label "nope"
```

dnnr:
Yup. I changed the message in #209, but that doesn't appear here yet.

PR #209, opened 2018-12-09
Introduce RuleManager

dnnr:


dnnr:
This includes #208, #203, and #207.

I've migrated some commands already. But for the add command, I first want to write some more tests.

dnnr:
I've extracted one more PR out of this: #212

PR #210, opened 2018-12-09
Test parsing and printing of almost all valid rules

dnnr:


dnnr:
Beware: This contains a fix for the output of "list_rules" that will cause conflicts with #209, depending on the merge order. I recommend merging this PR here first.

PR #214, opened 2018-12-11
Turn HSClientChanges into class

dnnr:


src/rules.cpp
```diff
@@ -481,28 +479,12 @@ int rule_remove_command(int argc, char** argv, Output output) {
...
-    }
-    if (changes->monitor_name) {
-        g_string_free(changes->monitor_name, true);
-    }
-}
+HSClientChanges::HSClientChanges(HSClient *client)
+    : focus(false)
+    , switchtag(false)
+    , manage(true)
+    , fullscreen(ewmh_is_fullscreen_set(client->window_))
```
t-wissmann: I think `focus`, `switchtag`, and `manage` should be initialized in the header as well. I don't know why I explicitly initalized them even though the memory is set to `0` before.

dnnr: I was a bit torn here, and it seems to be a reoccurring theme. When having to decide between this and spreading the initialization over decl and impl, I seem to always slightly prefer the *other* variant and end up oscillating a few times.

So I'll happily move it to the header. But at some point, I really like to have a solid, generic heuristic for making that choice and being satisfied with it. Any suggestions?

dnnr: Oh and maybe I misunderstood you here: Do you want them to be initialized in *both* places? If so, that was already the case, and I just worsened it with my update. Let me know.

t-wissmann: You understood me correctly, and I like the change. And I don't have good heuristics, but here it's better in the header  because the initializations are quite "simple" (whatever this means on a general level)

PR #221, opened 2018-12-14
 Cleanup interface of Input

ypnos:
Input is an ArgList that also stores the name of the command that
was called. It can only be constructed from a command name and (optionally) a
list of arguments to prevent wrong use.

All accessors from ArgList can be used. operator>> is overridden to
return the correct reference.

### DIFF
--- a/src/arglist.cpp
+++ b/src/arglist.cpp
@@ -53,15 +53,3 @@ ArgList &ArgList::operator>>(string &val) {
     }
     return *this;
 }
-
-ArgList ArgList::replaced(const std::string& from, const std::string& to) const {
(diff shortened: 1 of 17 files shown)


t-wissmann:
Great! Thanks a lot. Looking over the changes, I think we can merge them once the CI finishes :-)

ypnos:
The failure is some nasty crap, I didn't spot it because I had changed my debug flags, will fix it.

t-wissmann:
Take your time!

ypnos:
That's it from me for today :-)

t-wissmann:
That was nasty indeed...

PR #218, opened 2018-12-12
Improve herbstclient output passthrough in tests

dnnr:
This removes the python version selection from .travis.yml because
it doesn't work as expected, so no f-strings for us.

PR #224, opened 2018-12-15
Move rule_add_command() into RuleManager

dnnr:
The tables for mapping conditions and consequences to their respective
handlers are turned into std::maps, and HSCondition/HSConsequence now 
identify themselves by name instead of table index ("type").

As an improvement, rule parsing now validates the correct usage of "not".


dnnr:
This includes #218, which should be merged first

PR #227, opened 2018-12-17
Remove redundant local in Decoration::redrawPixmap

dnnr:
Since client->dec is always be identical to "this", the local copy is
pointless.


dnnr:
This change seems a bit *too* obvious, so I hope I'm not wrong with what I asserted in the PR description.

t-wissmann:
It's not too obvious, because we've missed this so far. This was simply part of the C-code where the variables of the struct were copied into the local context ;-)

PR #230, opened 2018-12-17
Move monitor globals into ManagerManager module

dnnr:
This eliminates the need for monitor_{init,destroy}.

The dtor of MonitorManager still has to call clearChildren() explicitly
(although the base class dtor does the same thing) to ensure that it
done before stack is deleted.


t-wissmann:
You can make the global variables members of MonitorManager (except for `g_monitors` itself).

dnnr:
Hmmm, let me check again. I either got a good reason not do to it (yet), or I've got another branch where I already did it. Or I just forgot about that step.

dnnr:
Oh yeah right: I wasn't sure if this is really what we want, and since it would necessitate changes in all points of usages, I decided not to do it. For the latter reason, I'd still like to postpone it for separate PR.

PR #231, opened 2018-12-17
Enable more warnings in debug flags

dnnr:


ypnos:
Unrelated to the warnings, but related to the debug flags:

I found that -Og, other than the name suggests, optimized out a variable that I needed to watch in debugging. Maybe I should revert it to -O1?

dnnr:
I was thinking about that back when you made the switch to CMake: For using ASAN, it's actualyl recommended to use at least -O1 in order to the minimize the runtime impact of the code instrumentation. I considered suggesting it, but given that the build type also needs to be usable for debugging, I thought that runtime performance is less important.

That being said, I'm really surprised that -Og worsens the debugging experience over -O1, but if it does, feel free to revert to -O1.

PR #232, opened 2018-12-17
Add some completion tests

t-wissmann:
Add tests for various types of completion functions.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,12 +31,19 @@ class HlwmBridge:
         # by self.hc_idle
         self.wmclass2winid = {}
 
-    def _checked_call(self, cmd, expect_success=True):
+    def _possibly_split_command(self, cmd):
+        """
+        Split a string using shlex but keep a list of strings intact.
+        """
         if isinstance(cmd, list):
             args = [str(x) for x in cmd]
             assert args
         else:
             args = shlex.split(cmd)
+        return args
+
+    def _checked_call(self, cmd, expect_success=True):
+        args = self._possibly_split_command(cmd)
 
         try:
             proc = subprocess.run([self.HC_PATH, '-n'] + args,
@@ -92,6 +99,16 @@ class HlwmBridge:
 
         self.client_procs.append(proc)
         return winid
+
+    def complete(self, cmd):
+        """
+        List all completions for the next argument for the given command.
+        This is more a helper to test completions.
+        """
+        args = self._possibly_split_command(cmd)
+        position = len(args)
+        res = self.call(['complete', position] + args)
+        return sorted(res.stdout.splitlines(False))
 
     def create_clients(self, num):
         return [self.create_client() for i in range(num)]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 import os
 import os.path
+import re
 import shlex
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -31,12 +32,19 @@ def __init__(self, display, hlwm_process):
         # by self.hc_idle
         self.wmclass2winid = {}
 
-    def _checked_call(self, cmd, expect_success=True):
+    def _possibly_split_command(self, cmd):
```
dnnr: I found this name a bit confusing. I think somethink like `_prepare_cmd_args` (or "parse") would be better. (Don't say what it does, but say what it's good for.)

tests/conftest.py
```diff
@@ -93,6 +101,67 @@ def create_client(self, title=None):
...
+                    raise Exception(("completion for \"{}\" returned the partial "
+                                    + "completion item \"{}\"").format(cmd, i)
+                                    ) from None
+                else:
+                    items.append(i[0:-1])
+        return sorted(items)
+
+    def list_children_via_attr(self, object_path):
+        """
+        list the names of children of the
```
dnnr: Start sentence with capital letter ;-)

t-wissmann: These and the `_possibly…` thing fixed in 8bd78b8

tests/test_clients.py
```diff
@@ -18,3 +18,6 @@ def test_alter_fullscreen(hlwm):
     assert hlwm.get_attr('clients.focus.fullscreen') == 'true'
     hlwm.call('attr clients.focus.fullscreen toggle')
     assert hlwm.get_attr('clients.focus.fullscreen') == 'false'
+
+def test_fullscreen_completion(hlwm):
+    assert hlwm.complete("fullscreen") == 'false off on toggle true'.split(' ')
```
dnnr: It feels a bit hacky to me, to use `split` like that. If most completion tests are going to look like that, I'd propose a different approach. If we were using pyhamcrest, we could introduce a custom matcher here. Within out toolset, we could make `hlwm.complete()` return an object offers a matcher method that takes care of the string splitting.

tests/test_rules.py
```diff
@@ -158,6 +158,15 @@ def test_complete_unrule_offers_all_rules(hlwm, rules_count):
 
     assert call.stdout == '\n'.join(rules + ['-F', '--all']) + '\n'
 
+def test_complete_rule(hlwm):
```
dnnr: Add another blank space above here (PEP-8 rule)

tests/test_rules.py
```diff
@@ -158,6 +158,15 @@ def test_complete_unrule_offers_all_rules(hlwm, rules_count):
...
+    global string_props, numeric_props, consequences
```
dnnr: There is no need for this `global` statement

tests/test_rules.py
```diff
@@ -158,6 +158,15 @@ def test_complete_unrule_offers_all_rules(hlwm, rules_count):
...
+    assert hlwm.complete('rule', partial=True) == sorted(
+        [ i + ' ' for i in '! not prepend once printlabel'.split(' ')]
```
dnnr: There's not supposed to be whitespace after opening brackets (another PEP-8 rule)

dnnr:
Besides my comments, this looks good to me overall!

tests/test_clients.py
t-wissmann: I'm really unsure about the future of tests for completions. I don't want to copy the entire completions to the test suite https://github.com/herbstluftwm/herbstluftwm/blob/master/src/command.cpp#L31

dnnr: It's a very large surface to cover. Maybe we could try a heuristic approach?

 * for every command dispatched in the test suite, verify that it would have been "reachable" via 
completions
 * enumerate and run everything that is offered via completions, expect it to succeed (not sure if this is feasible)

PR #234, opened 2018-12-17
Integrate matchers/appliers into rule classes

dnnr:


t-wissmann:
I think all the members can be private. They have been private in the C-style before, namely static functions in the c-file.

PR #235, opened 2018-12-17
CMake: Make use of ccache if available

dnnr:


ypnos:
I never used ccache so I wonder: Might there be a case where you don't want to use ccache? Because this will always use it when installed. Even for people who just download/pull and build one time (e.g. build scripts).

dnnr:
There might be such a case, but in my experience, it's pretty safe to use. The only unwanted side-effect I could think of is the increased disk usage (it does limit cache size to 5G by default).

Given that it's quite noticeable when ccache is enabled, I'd wager that anybody who has it installed *and* doesn't want its side-effects, will notice and turn it off.

The knowing-how-to-turn-it-off part is the only step I'm not so sure about. Maybe there is a better way of making this visible to the user? At least in Qt Creator, it shows up in the project configuration.

ypnos:
All I see in QtCreator is that a CCACHE _path_ variable pops up, and I am unsure what I would have to do there to disable it. Am I missing something else?

Maybe we should have a logic like for the Xinerama option ([see here](https://github.com/herbstluftwm/herbstluftwm/blob/bf75ecd0d7b3d1d83e7f91d649a64bc7c7ded82c/src/CMakeLists.txt#L64-L70)), like having a ENABLE_CCACHE option with the default being OFF. I feel like only developers would want to use ccache.

dnnr:
I'm putting developers and users that compile hlwm themselves into the same basket here.

But I agree, having an ENABLE_CCACHE flag seems preferable. My primary goal here was being able to use ccache in Qt Creator without having to set the launcher variables manually. Apart from defaulting to OFF, such a flag should achieve the same thing. I'll change the PR.

PR #236, opened 2018-12-17
Turn monitor-related globals into members

dnnr:


dnnr:
This includes PR #230

PR #240, opened 2018-12-19
Enable -Wshadow (except in ctors) for clang builds

dnnr:


dnnr:
This includes PR #238

PR #241, opened 2018-12-20
Completion context as an object

t-wissmann:
This introduces a new completion framework that has the current completion context as an object that is passed to the completion functions. It already migrates the completion for the commands rule, unrule, object_tree, get_attr, fullscreen and pseudotile. This should cover all kinds of completions. The remaining commands will be migrated subsequently.


t-wissmann:
I think the most interesting parts are the changes in the command table in main.cpp and the completion helpers in rulemanager.cpp and clientmanager.cpp.

src/completion.h
```diff
@@ -0,0 +1,66 @@
...
+#include "arglist.h"
+#include "types.h"
+
+/** The completion object holds the state of a running
+ * command completion, that is the list of args and the index
+ * of the element in the arglist that needs to be completed.
+ *
+ * The intended use is to pass the completion state as the reference and to
+ * return possible completions via this Completion object. This is why the
+ * operator= and the copy constructor are private. It ensures that a completion
```
dnnr: I think it makes more sense to move the explanation for why they are private to their declaration. It doesn't seem relevant to the class in general (except maybe in terms of just saying that it is non-copyable).

t-wissmann: Alright, I've moved them.

PR #243, opened 2018-12-20
CMake: Extract versioning into separate file

dnnr:


ypnos:
This does not work. The project() call _must_ be in the top-level CMakeLists.txt and cannot included.
I don't even think it is that unfortunate; I think hiding away the project() in some obscure file is not a good design. Maybe you can include version.cmake before the project call?

dnnr:
You're totally right. Somehow, I've missed that this is the project() call.

PR #244, opened 2018-12-20
Explain our compiler flags, add more warnings

dnnr:
This is mostly restructuring, but also adds the following flags:
-Wshadow, -Wshadow-all, -Wduplicated-cond, -Wduplicated-branches,
-Wlogical-op, -Wuseless-cast

...and removes a suppression for shadow warnings in ctors.


dnnr:
This depends on the fixes that come with #242 (build will fail until then)

PR #246, opened 2018-12-21
Fix typo in valgrind-xephyr.sh

blarz:


blarz:
I don't think this was intended, right? ;)

t-wissmann:
Thanks :) I was typing `../val`<kbd>Tab</kbd> all the time.

PR #248, opened 2018-12-23
More completions

t-wissmann:
Add completions and completion tests for the following commands:

- Settings: set, get, cycle_value
- RootCommands: attr, set_attr


t-wissmann:
PR #249 needs to be merged first.

src/rootcommands.cpp
```diff
@@ -375,3 +379,29 @@ void RootCommands::get_attr_complete(Completion& complete) {
     else complete.none();
 }
 
+void RootCommands::set_attr_complete(Completion& complete) {
+    if (complete == 0) {
+        completeObjectPath(complete, true, \
```
ypnos: Why the backslash?

src/settings.cpp
```diff
@@ -140,6 +142,9 @@ Settings::Settings(Root* root)
         root->monitors()->lock_number_changed();
     });
     g_settings = this;
+    for (auto i : attributes()) {
+        i.second->setWriteable();
```
ypnos: Why do we need this now and not before? Isn't this redundant with the individual setWriteable() statements above?

src/settings.cpp
```diff
@@ -243,6 +263,24 @@ int Settings::toggle_cmd(Input argv, Output output) {
     return 0;
 }
 
+void Settings::toggle_complete(Completion& complete) {
+    // types that can be toggled
+    std::vector<Type> l = {
```
ypnos: I suggest using std::set here and then later on std::set::find() instead of std::find(). Better readability.

src/settings.cpp
```diff
@@ -269,6 +307,17 @@ int Settings::cycle_value_cmd(Input argv, Output output) {
     return 0;
 }
 
+void Settings::cycle_value_complete(Completion& complete) {
+    if (complete == 0) {
+        for (auto& a : attributes()) {
```
ypnos: The reference designation doesn't make much sense here.

src/settings.cpp
t-wissmann: This didn't cover all settings. And instead of explicitly making the remaining ≈5 settings I've decided to simply mark all settings writable, because all settings must be writable. We can omit the explicit `setWritable()` calls.

src/settings.cpp
t-wissmann: I think it's overkill to construct a set (i.e. binary tree/heap) with three elements. Because of #249 / #250, I don't need the vector/set anyway because after #250, only boolean settings can be toggled.

src/settings.cpp
t-wissmann: I've removed the earlier setWritable() calls.

src/rootcommands.cpp
t-wissmann: It's gone. I can't explain where it came from :)

src/settings.cpp
dnnr: Just a sidenote (even if it doesn't matter anymore here): A decision like this sounds like a textbook example for when *not* to take performance/overhead into account. Creating a set is NlogN (I think), not N^9000, so for three elements it just doesn't matter. That's all we should care about when writing code like that, and I appreciate the fact that's even saying so in the Goals & Principles.

PR #249, opened 2018-12-23
Toggle settings of type unsigned long

t-wissmann:


winterbreeze.asciidoc
```diff
@@ -14,7 +14,7 @@ the following, all user-visible changes are listed:
     * +monitors..by-name.+ is *invalid*
 
  - the 'toggle' command does not toggle between zero and the last non-zero value
-   anymore but simply between 0 and 1. For the former behaivour, one needs to use
+   anymore but simply between 0 and 1. For the former behaviour, one needs to use
```
dnnr: This behavior is purposely different from master, right? Since it's an incompatible change anyway, wouldn't it make sense (in terms of orthogonality) to just *not* offer toggling for integer settings at all, but only for bools?

t-wissmann: Right, and in general it seems preferable to have an error message instead of just different behaviour.

t-wissmann: See #250

PR #250, opened 2018-12-23
Toggle now only works for boolean settings

t-wissmann:


tests/test_settings.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.mark.parametrize('cannot_toggle',
+    ['window_border_width',
+    'frame_border_active_color',
+    'default_frame_layout',
+    'wmname']
+)
+def test_cannot_toggle_non_numeric(hlwm, cannot_toggle):
+    hlwm.call_xfail("toggle " + cannot_toggle)
```
dnnr: Check the error message, too?

tests/test_settings.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+
+@pytest.mark.parametrize('can_toggle',
+    ['update_dragged_clients']
+)
+def test_toggle_numeric_settings(hlwm, can_toggle):
```
dnnr: The test case name should probably be something like `test_toggle_bool_setting` now

dnnr: (similarly for its counterpart)

PR #251, opened 2018-12-27
Completion for inputless commands

t-wissmann:
Automatically set up the completion function in CommandBindings for commands that do not have any input arguments. Inputless commands are those that have only an Output parameter but no Input parameter. These are currently:

    quit true false reload version list_commands list_monitors list_rules
    list_keybinds lock unlock close_or_remove close_and_remove remove rotate
    mouseunbind use_previous lock unlock


src/command.h
```diff
@@ -49,6 +49,16 @@ class CommandBinding {
         // Ignore the output parameter
         : command([cmd](Input args, Output) { return cmd(args); })
     {}
+    // A command that doesn't have input
+    CommandBinding(function<int(Output)> cmd);
+    // A command that doesn't have input or output
+    CommandBinding(function<int()> cmd);
+    // A normal command and it's completion
```
dnnr: Typo: `its`

tests/test_completion.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+import shlex
+
+commands_without_input = shlex.split(
```
dnnr: I suppose it's mostly a matter of preference, but I would simply put it something like this:
```
commands_without_input = [
    quit,
    true,
    false,
    ...etc...
]
```

Either that, or put it all in a single line (not my preference, though). To me, formatting a list like a text paragraph seems to combine the worst of both worlds:
 * hard to read diffs on removal/addition
 * duplicate entries are not easily spotted (case in point: you already have duplicates)
 * to future contributors, it's unclear how they're supposed to order/group things, because it seems arbitrary

It's a minor thing *here*, but I've seen this pattern in a couple of places, so I felt the need to voice my opinion here.

tests/test_completion.py
```diff
@@ -0,0 +1,21 @@
...
+        mouseunbind
+        use_previous
+        lock unlock
+    """)
+
+
+@pytest.mark.parametrize('name', commands_without_input)
+def test_inputless_commands(hlwm, name):
+    assert hlwm.call_xfail_no_output('complete 1 ' + name) \
+        .returncode == 7
```
dnnr: I'd advise against comparing it to a numeral that is not really defined/hardcoded/documented. That's going to be a pain point if we ever touch that exit code enum.

We should to at least one of these:
 * add a FIXME here explaining where the `7` comes from and that it's somewhat fragile
 * explicitly assign a numeral to each `HERBST_` exit code to clarify that they are assigned for all eternity (if that's our intention)
 * make the enum values available in the tests, so that we can use them right here instead of numerals

src/arglist.cpp
```diff
@@ -14,6 +14,12 @@ ArgList::ArgList(const std::string &s, char delim) {
     reset();
 }
 
+ArgList::ArgList(Container::const_iterator from, Container::const_iterator to)
+{
+    c_ = std::make_shared<Container>(from, to);
```
dnnr: Not really related to this PR, but I think that this `c_` member deserves a meaningful name. Even calling it `data_` (still very generic) would be an improvement.

tests/test_completion.py
t-wissmann: I've formatted it to one line per entry, but I'd like to keep the shlex.split().

dnnr: Fair enough :+1:

src/arglist.cpp
t-wissmann: `container_`?

tests/test_completion.py
dnnr: (Btw: The man page for herbstclient refers to that of herbstluftwm for a list of exit codes, but I couldn't find it there. In case I just missed it, strike my mention of "undocumented".)

src/arglist.cpp
dnnr: Yup, sounds good to me!

tests/test_completion.py
t-wissmann: I've opened #253 for the lack of documentation and added a FIXME here.

dnnr: Oh and: If we're going with the eternal assignments (which is probably the consequence of documenting them), I'd say we could simply duplicate them in conftest.py. I've mentioned technical solutions for this before (you could parse ipc-protocol.h using ctypeslib's clang2py), but I don't think they're worth the effort in that case.

PR #252, opened 2018-12-27
tests: Automatically validate completion suggestions

t-wissmann:
In the tests, generate all full-word completions (i.e. something like -F or --all but not something like label= which still requests some user input) up to a certain length (currently 4) and validate that they run successfully.


t-wissmann:
This includes #251 which needs to be merged first.

t-wissmann:
I've deactivated the parallelization for tox for the moment. In the tests for completion, one test case generates the commands and another test case checks them. So they can't be run in parallel. Maybe this can be solved by using `pytest-dependency`.

t-wissmann:
> I've deactivated the parallelization for tox for the moment. In the tests for completion, one test case generates the commands and another test case checks them. So they can't be run in parallel. Maybe this can be solved by using `pytest-dependency`.

Negative.... this is in the python-dependency-docs:

> 1.2.2 Interaction with other packages
> 
> pytest-xdist features test run parallelization, e.g. distributing tests over separate processes that run in parallel. This is based on the assumption that the tests can be run independent of each other. Obviously, if you are using pytest-dependency, this assumption is not valid. Thus, pytest-dependency will only work if you do not enable parallelization in pytest-xdist.

Maybe python-xdist offers some way to define dependencies...

t-wissmann:
Now, the test waits actively for the cache to be filled.

src/client.cpp
```diff
@@ -624,14 +625,20 @@ HSClient* get_client(const char* str) {
...
+Window get_window(const std::string& str) {
     // managed window?
-    auto client = get_client(str);
+    auto client = get_client(str.c_str());
     if (client)
         return client->window_;
 
     // unmanaged window? try to convert from base 16 or base 10 at the same time
-    return std::stoul(str, nullptr, 0);
+    try {
```
ypnos: I suggest using `Converter<unsigned long>::parse()` here for consistency on how we parse string input. Alternatively, remove the two superfluous arguments to stoul(). You can also use `catch (...)` instead of the two catches.

src/client.cpp
```diff
@@ -413,8 +413,9 @@ Rectangle HSClient::outer_floating_rect() {
     return dec.inner_to_outer(float_size_);
 }
 
-int close_command(int argc, char** argv, Output) {
-    auto window = get_window((argc > 1) ? argv[1] : "");
+int close_command(Input input, Output) {
+    std::string winid;
+    auto window = get_window((input >> winid) ? winid : "");
```
ypnos: Is this better than `get_window((input.empty() ? "" : input.front())`? I personally don't like the shift being encapsulated like that (side-effect hidden in forming a method call).

src/command.h
```diff
@@ -49,6 +49,16 @@ class CommandBinding {
         // Ignore the output parameter
         : command([cmd](Input args, Output) { return cmd(args); })
     {}
+    // A command that doesn't have input
+    CommandBinding(function<int(Output)> cmd);
+    // A command that doesn't have input or output
+    CommandBinding(function<int()> cmd);
+    // A normal command and it's completion
```
ypnos: typo+lang "A regular command and its completion"

src/command.h
```diff
@@ -49,6 +49,16 @@ class CommandBinding {
         // Ignore the output parameter
         : command([cmd](Input args, Output) { return cmd(args); })
     {}
+    // A command that doesn't have input
+    CommandBinding(function<int(Output)> cmd);
+    // A command that doesn't have input or output
```
ypnos: typo "input nor output" (nitpick level >9000)

tests/test_completion.py
```diff
@@ -0,0 +1,197 @@
...
+        use_previous
+        lock unlock
+    """)
+
+
+def command2str(command):
+    s = ' '.join(command)
+    maxlen = 20
+    if len(s) > maxlen:
+        s = s[0:maxlen - 3] + '...'
```
ypnos: I suggest `s = s[0:maxlen - 1] + '…'`

tests/conftest.py
```diff
@@ -44,7 +44,7 @@ def _parse_command(self, cmd):
             args = shlex.split(cmd)
         return args
 
-    def _checked_call(self, cmd, expect_success=True, expect_error_output=False):
+    def unchecked_call(self, cmd, log_output=True):
```
dnnr: I'm not too happy with the new name. While it makes sense in the context of the diff, it doesn't on its own, because it's not even clear anymore what "checking" is supposed to mean when just looking at the function.

That said, I thought about it and couldn't come up with a much better suggestion. If it was still internal, I'd call it `_call`, but it's not. So maybe just add a docstring along the lines of "Performs a herbstclient call without checking for its success"?

tests/conftest.py
```diff
@@ -61,28 +61,30 @@ def _checked_call(self, cmd, expect_success=True, expect_error_output=False):
         outcome = 'succeeded' if proc.returncode == 0 else 'failed'
         allout = proc.stdout + proc.stderr
         if allout:
-            print(f'Client command {args} {outcome} with output:\n{allout}')
+            if log_output:
+                print(f'Client command {args} {outcome} with output:\n{allout}')
+            else:
+                print(f'Client command {args} {outcome} with output.')
```
dnnr: I found this confusing at first. How about printing `Client command {args} {outcome} (output suppressed)`?

tests/test_clients.py
```diff
@@ -19,5 +19,11 @@ def test_alter_fullscreen(hlwm):
     hlwm.call('attr clients.focus.fullscreen toggle')
     assert hlwm.get_attr('clients.focus.fullscreen') == 'false'
 
+
 def test_fullscreen_completion(hlwm):
     assert hlwm.complete("fullscreen") == 'false off on toggle true'.split(' ')
+
+
+def test_close_without_clients(hlwm):
+    # close may return HERBST_INVALID_ARGUMENT if there is no window to close
```
dnnr: But within the controlled environment of this test we should be able to *know* that there are no windows to close, right?

Shouldn't this be split into two tests like
 * `test_close_focused_window`
 * `test_close_nonexistent_focused_window`

src/client.cpp
ypnos: Yes I don't see a point in this kind of consistency, I would use the accessor that fits best in the circumstances. For example, there is other places where the number of arguments is unknown, and there it's best to just have a for-loop.

tests/test_completion.py
t-wissmann: I had this originally, but unfortunately pytest prints unicode characters quoted:
```
test_my_test_case[some_long_para\u2026]
```
Currently, this helper function is unused anyway so I've commented this out for the moment.

ypnos: Doh! Looks like they are totally weird about this as well (https://github.com/pytest-dev/pytest/issues/2482)

tests/conftest.py
t-wissmann: Fine for me.

tests/test_clients.py
t-wissmann: I've splitted the test.

tests/conftest.py
t-wissmann: I've added a doc-string here and to the `call()`, which checks there exit code.

src/client.cpp
t-wissmann: I've droped the two default arugments of `stoul()` but kept the (only) two catch statements.

dnnr: I guess from a Python programmer's point of view it's always better to catch with strict typing. But in C++, and for this kind of standard function, is there any reason to assume that it could throw anything other than what is documented? (Not a rhetorical question, tbh...)

I figure that in this case, the distinction between exceptions type is meant for when you need to know what exactly went wrong (which we don't), so I'd second the suggestion to use `catch (...)`. 

For what it's worth, catching `std::logic_error` would also work.

tests/test_completion.py
dnnr: Bummer. I didn't even know we had an inside man there.

src/client.cpp
dnnr: I agree with @ypnos' concerns, but I also think that we kind of paved the way for non-ideal situations like this when we introduced `>>` such that it has a meaningful return value.

Elsewhere, this mostly just leads to `if (input >> foo) {…}`, which is only slightly less bad than wrapping a function call around it like here, in my opinion.

So while I'm with @thorsten here in that I'd prefer sticking to `>>` whenever possible, I see the point of not sacrificing readability for consistency.

So maybe there is a way to use `>>` but retain readability. Wouldn't it be possible to simply put it like this?

```
std::string winid;
input >> winid;
auto window = get_window(winid);
```

ypnos: Yes this is possible. The usage within `if` for me is already bad style (side-effect of an if-condition! noes!) but it is tolerable because it is somewhat contained and makes the code much more readable where we use it. (Also look at http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0305r0.html urgs).

I do not see the rationale behind using >> whenever possible. It should be used where it works best. Writing three lines instead of one that is simply understood for me is not best use.

dnnr: Well, from my pov, it's not about using `>>` whenever possible, but about providing an interface that is universally preferred. I had hoped that the stream operator is able to do that, and I view every situation where we "fall back" to `input.front()` as reason to reconsider the merits of `>>`. Maybe it can still be improved?

That being said, it is of course always possible that the best solution is two solutions. That seems to be the case with ArgList's interfaces.

The points you make against my suggestion are perfectly valid, nothing to argue there.

src/client.cpp
t-wissmann: If at some point, we add `noexcept` to the function signature, we know that we did not forget any exception. But if you both prefer `catch(...)` I can change this.

src/client.cpp
t-wissmann: I've updated the function to
```
int close_command(Input input, Output) {
    std::string winid = "";
    input >> winid; // try to read, use "" otherwise
    auto window = get_window(winid);
    if (window != 0)
        window_close(window);
    else return HERBST_INVALID_ARGUMENT;
    return 0;
}
```
If you want to discuss the usage of `>>` please open a issue, but it's not in the scope of this PR.

PR #255, opened 2019-01-07
CMake: Use 4 spaces for indentation, add modelines

dnnr:


t-wissmann:
Cool. Can we also check this in a C++-linter that this specific modeline is respected?

PR #256, opened 2019-01-07
Remove global variable g_current_rule_birth_time

dnnr:


t-wissmann:
It's good to get rid of the global but at the same time a pitty that the timestamp is a member of the condition and not of the rule. Maybe we can improve this at some point.

PR #257, opened 2019-01-07
Turn apply_rules() into RuleManager method

dnnr:


dnnr:
Includes the changes from #256

t-wissmann:
I've just fixed my wrong merge commit and additionally I've removed the declaratoin of rules_apply() in rules.h

PR #258, opened 2019-01-07
Turn g_rules into private member of RuleManager

dnnr:


t-wissmann:
Looks good to me, thanks!

PR #259, opened 2019-01-07
Make matcher functions take const HSClient pointer

dnnr:


t-wissmann:
I think you should also add a const in the definition&declaration of HSCondition::matchers.

t-wissmann:
I've proposed some changes to `HSCondition::matchers` by pushing to your branch.

dnnr:
Good idea, thanks for the improvement! I made a tiny whitespace change on top, and from my side, this change is good to go.

dnnr:
Agreed, making the methods const was close too the upper bound of pedantry. I wouldn't hesitate to revert this (at least for the `Matchers` signature) if it really becomes necessary.

"Morally non-const" is an interesting concept, btw. :laughing:

PR #260, opened 2019-01-08
Handle pseudotile consequence via HSClientChanges

dnnr:
This adds a shim for std::optional taken from
https://github.com/akrzemi1/Optional (including the open PR #47).


src/optional.h
```diff
@@ -0,0 +1,1100 @@
+// Copyright (C) 2011 - 2012 Andrzej Krzemienski.
+//
+// Use, modification, and distribution is subject to the Boost Software
+// License, Version 1.0. (See below or copy at http://www.boost.org/LICENSE_1_0.txt)
```
dnnr: I've changed the file header to include the license directly (see https://github.com/akrzemi1/Optional/blob/master/optional.hpp for comparison). I'm not well-versed in licensing, so please double-check if this reasonably waterproof.

src/clientmanager.cpp
```diff
@@ -118,6 +118,10 @@ HSClient* ClientManager::manage_client(Window win, bool visible_already) {
         }
     }
 
+    if (changes.pseudotile.has_value()) {
+        client->pseudotile_ = changes.pseudotile.value();
```
dnnr: There is also a `set_pseudotile` method, which I *suppose* isn't what we want to use here, because it also calls `applyLayout`.

It's also entirely unused. Should it, by any chance, be registered as a setter at the pseudotile attribute?

src/rules.h
```diff
@@ -70,6 +71,8 @@ class HSClientChanges {
     bool            manage = true; // whether we should manage it
     bool            fullscreen;
     std::string     keymask; // Which keymask rule should be applied for this client
+
+    std::experimental::optional<int> pseudotile;
```
t-wissmann: Why is this an `int` and not `bool`?

src/clientmanager.cpp
t-wissmann: Right, we don't want `set_pseudotile` and I assume we don't need `set_pseudotile` at all. I guess that setting `client->pseudotile_` also calls `applyLayout` because there is a signal connection from the `pseudotile_.changed` signal. Maybe we need an extra `connectSignals()` method in `Client` that is called after the rules are applied.

src/rules.h
dnnr: Because I'm not a clever man. Fix is coming up.

PR #261, opened 2019-01-08
Add "const" to rule applier methods

dnnr:


dnnr:
This includes changes from #260 

We might not want to merge this (but we could), so this is primarily meant as a basis for discussion. See the TODO in HSConsequence declaration.

Should I resolve the TODO by giving the `applyEwmh…` methods the same treatment I gave `applyPseudotile` in #260? I wasn't entirely sure here...

t-wissmann:
If we do so, then (except for `hook=`) all applier methods will look the same, namely:
```
HSConsequence::applyFoo(....) {
   clientchanges->foo = string_to_bool/int/...(value);
}
```
I don't think this is a bad thing but simply room for further optimizations (that are not urgent).

dnnr:
Alright! Then let's just keep this PR open until it's done.

dnnr:
Includes #265 now, too

dnnr:
Should be good to go now (finally...)

PR #277, opened 2019-01-12
Drop discouraged "using namespace std;"

dnnr:
In some cases, for rarely used symbols, "std::" is added to the
point-of-use.

In every other case, a "using std::…" statement is added.

For consistency, any existing "std::" prefixes are removed if they're
not necessary thanks to a corresponding "using" statement.

### DIFF
--- a/src/arglist.cpp
+++ b/src/arglist.cpp
@@ -1,8 +1,9 @@
 #include "arglist.h"
 
-using namespace std;
+using std::string;
+using std::stringstream;
 
-ArgList::ArgList(const std::initializer_list<std::string> &l)
+ArgList::ArgList(const std::initializer_list<string> &l)
     : container_(std::make_shared<Container>(l))
 { reset(); }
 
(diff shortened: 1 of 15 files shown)


dnnr:
I've tried to be minimally invasive here. But as an idea, we could also devise a set of standard library types that we always pull into the local namespace, by convention. Those could be, for instance, `string`, `vector`, `unique_ptr`, etc.

From that list, I'd decidedly exclude things like `max` because that's not an "obvious" standard library symbol (that's subjective of course) and might lead to confusion if used without the `std::` prefix.

What do you think?

dnnr:
We could put it in the coding style section (hoping it doesn't grow too large), but I'd also be fine with an unwritten rule.

Implementation-wise, I *wouldn't* use any kind of common header file, but still just pull in what is actually needed in the individual cpp files.

t-wissmann:
Yes, the too obvious solution is a header file consisting only of `using …` statements, but this feels unclean, even though I can't really explain why.

Let's keep it as an unwritten law or lets put it in the `HACKING` file.

PR #279, opened 2019-01-16
Extract monitor description into method

dnnr:


dnnr:
This includes #278

t-wissmann:
Looks good!

PR #280, opened 2019-01-16
Rewrite stack tree printing

dnnr:
This implements print_stack_command() using a new subclass of
TreeInterface, making the old HSTreeInterface (kept for compatibility)
obsolete.


dnnr:
This includes #278 and #279

t-wissmann:
Great and much cleaner than before! And I agree that it's fine to ignore the monitor-slice case in `getSliceLabel()`.

PR #281, opened 2019-01-16
Replace g_strdup with strdup to drop GLib in utils

dnnr:


dnnr:
This is not possible yet, but the build should pass as soon as #280 is merged.

PR #286, opened 2019-01-17
Extract keybind tests, add test with key injection

dnnr:


tests/test_keybind.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def test_trigger_single_key_binding(hlwm):
+    hlwm.call('keybind x use tag2')
+    hlwm.call('add tag2')
+    assert hlwm.get_attr('monitors.0.tag') != 'tag2'
+
+    subprocess.call('xdotool key x'.split())
+
+    assert hlwm.get_attr('monitors.0.tag') == 'tag2'
```
t-wissmann: Is a race-condition? The `xdotool` can finish before hlwm starts processing the key. But can it happen that the `herbstclient` makes it into hlwm's event queue before the key arrives?

dnnr: I was wondering the same thing, but... well... it *does* work. ;-)

dnnr: More seriously: Even though it might be a race-condition, its worst-case impact would be a flaky tests. I'd say it's reasonable to ignore this until we actually observe such flakiness in this test case.

t-wissmann: I agree.

PR #288, opened 2019-01-17
Turn g_key_binds from GList into std::vector

dnnr:
This purposely leaves the C-style nature of KeyBinding (not being a
class, etc.) untouched in order to tackle that part separately.


t-wissmann:
Seems fine.

PR #289, opened 2019-01-17
Use std::strings for string2… helpers in key.cpp

dnnr:


dnnr:
This includes #288

src/key.cpp
```diff
@@ -103,51 +107,60 @@ int keybind(int argc, char** argv, Output output) {
...
+    for (auto &sep : KEY_COMBI_SEPARATORS) {
+        std::replace(keySpec.begin(), keySpec.end(), sep, baseSep);
+    }
+
+    // Split spec into tokens:
+    vector<std::string> tokens;
+    std::string token;
+    std::istringstream tokenStream(keySpec);
+    while (std::getline(tokenStream, token, baseSep)) {
+        tokens.push_back(token);
```
t-wissmann: Don't you need to strip off `baseSep` from the end of token?

dnnr: The delimiter gets dropped by `std::getline` ("the delimiter character is extracted from input, but is not appended to str." https://en.cppreference.com/w/cpp/string/basic_string/getline)

t-wissmann: Thanks!

PR #292, opened 2019-01-18
Make KeyBinding a class, use Input for keybind cmd

dnnr:
For this, the global g_key_binds is turned into a vector of unique_ptrs
to KeyBinding.

In future work, most of the standalone functions may be turned into
methods of KeyBinding.

This also extends the keybinding test suite a bit.

### DIFF
--- a/src/key.cpp
+++ b/src/key.cpp
@@ -15,6 +15,7 @@
 #include "ipc-protocol.h"
 #include "utils.h"
 
+using std::unique_ptr;
 using std::vector;
 
 static unsigned int numlockmask = 0;
@@ -24,7 +25,7 @@ unsigned int* get_numlockmask_ptr() {
     return &numlockmask;
 }
 
-static vector<KeyBinding> g_key_binds = {};
+static vector<unique_ptr<KeyBinding>> g_key_binds = {};
 
 void key_init() {
     update_numlockmask();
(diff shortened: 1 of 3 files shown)


src/key.cpp
```diff
@@ -85,31 +79,39 @@ const char* modifiermask2name(unsigned int mask) {
...
+    newBinding->cmd = {input.begin(), input.end()};
+
+    // Remove existing binding with same keysym/modifiers
+    key_remove_bind_with_keysym(newBinding->modifiers, newBinding->keysym);
+
+    // Grab for events on this keycode
+    grab_keybind(newBinding.get());
+
+    // Add keybinding to list
+    g_key_binds.push_back(std::move(newBinding));
```
dnnr: Please note that I've swapped the order of `grab_keybind` and `push_back` (because it made things easier). I'm assuming that it doesn't make a difference.

t-wissmann: Is fine I think.

t-wissmann:
Looks good to me.

I think it's clear that we need something like a `KeyManager` here that also has a handle to the `XConnection`. But it's not so clear to me, whether something like `grab_keybind` should become a member of `KeyManager` or of `KeyBinding`. I tend to put it in the `KeyManager` to make `KeyBinding` a self-contained dump datastructure (similar to the `TilingResult`).

PR #293, opened 2019-01-20
Use std::string over GString for key binding names

dnnr:


src/key.cpp
```diff
@@ -346,9 +343,8 @@ void key_find_binds(const char* needle, Output output) {
 static void key_list_binds_helper(KeyBinding* b, std::ostream* ptr_output) {
     Output output = *ptr_output;
     // add keybinding
-    GString* name = keybinding_to_g_string(b);
-    output << name->str;
-    g_string_free(name, true);
+    auto name = keybinding_to_string(b);
+    output << name;
```
t-wissmann: Why not `output << keybinding_to_string(b);` directly?

t-wissmann:
Thanks! Looks good

src/key.cpp
dnnr: No particular reason, I've changed it. :+1:

PR #296, opened 2019-01-20
Clean up initialization order for manager objects

dnnr:
This introduces an alphabetical ordering where possible, and extracts
special cases (having dependencies) into a separate block.


t-wissmann:
What is it good for?

dnnr:
I expected that this might be a bit controversial. When trying to add the KeyManager here, I was a bit frustrated by the seemingly arbitrary order of things. This PR is an attempt and making it more clear where a new thing should go.

PR #297, opened 2019-01-20
Extract keybind commands into new KeyManager

dnnr:


t-wissmann:
It's fine. But am I assuming it correctly that nearly all other functions from key.h will also become member functions of KeyManager?

dnnr:
No, I don't think so. But I'm taking this step-by-step, so I haven't given it too much thought yet. Btw, even though I've just introduced it, I'll also explore possibilities to get the X-specific code back out of KeyManager. Right now, I'm still more focused on moving functions into classes and getting rid of GLib.

PR #298, opened 2019-01-26
Do not deny key bindings for unknown symbols

dnnr:
We should not deny the setup of a new keybinding only because the KeySym
is unknown at the time of setup. Instead, unknown KeySyms are skipped
whenever (re)grabbing them.


t-wissmann:
Sure! I've also described the new behaviour in the winterbreeze.md file.

dnnr:
Good call, thanks!

PR #299, opened 2019-01-27
Move/reword keymask documentation for clarity

dnnr:


dnnr:
By the way, I was a bit confused that this mechanism is only explained as a rule consequence. It's a standalone attribute that can be used without rules, isn't it?

t-wissmann:
Yes. Actually it's the attribute that needs this documentation and the rule consequence just sets the attribute.

dnnr:
I've shoehorned it into the attributes table now. Not sure how bad that is. See for yourself.

t-wissmann:
It's fine, thanks!

PR #301, opened 2019-01-27
Use std::regex for keymask, add tests

dnnr:
This implicitly fixes a memory leak in the old regex code (missing
regfree).

The new test is prepared to cover multiple scenarios, only one of which
works in current keymask implementation.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -111,7 +111,7 @@ class HlwmBridge:
         winid = self.wait_for_window_of(wmclass)
 
         self.client_procs.append(proc)
-        return winid
+        return winid, proc
 
     def complete(self, cmd, partial=False, position=None):
         """
@@ -174,7 +174,7 @@ class HlwmBridge:
         return sorted(children)
 
     def create_clients(self, num):
-        return [self.create_client() for i in range(num)]
+        return [self.create_client()[0] for i in range(num)]
 
     def wait_for_window_of(self, wmclass):
         """Wait for a rule hook of the form "here_is_" + wmclass """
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -271,3 +272,14 @@ def running_clients(hlwm, running_clients_num):
...
     return hlwm.create_clients(running_clients_num)
+
+
+@pytest.fixture()
+def keyboard():
+    class KeyBoard:
+        def press(self, s):
+            subprocess.call('xdotool key x'.split())
+            # Bad workaround for keypress injection not being synchronous:
+            time.sleep(.5)
```
dnnr: This is a really bad hack and I'm not sure if we want to merge this. Any ideas on how to solve it properly?

t-wissmann: We can read herbstluftwm's stderr and wait for the keypress to be processed. Shouldn't this be `s` instead of `x` in the xdotool-call?

t-wissmann: What we can do in this (test-)case: run a `read -n 1` in the terminal, type the `x` and expect the terminal to close. (We can make a waitpid with a timeout of 5 seconds so we only wait the full 5 seconds if it fails).

dnnr: Yeah, I forgot to actually use the parameter... will fix it.

So, there is no generic way of checking whether a keypress has been processed by *any* client, right? Is there some X queue that we could inspect?

dnnr: I when with the `read -n 1` solution. PRs #305 and #304 are included and need to be merged first.

PR #303, opened 2019-01-28
Create client custom cmd

dnnr:


dnnr:
Oops, I think I created this by mistake... no idea how it happened.

PR #306, opened 2019-01-28
Use (obvious) dummy IDs in stack tree expectation

dnnr:


ypnos:
Or we could just call it `<windowid` to express that it is a placeholder.

dnnr:
Sure, why not. Changed it. :+1:

PR #307, opened 2019-02-02
Clean up and extend tests for keymask

dnnr:


dnnr:
This includes #302

tests/test_keybind.py
```diff
@@ -95,9 +97,14 @@ def test_keymask(hlwm, keyboard, maskmethod, order):
...
     if maskmethod == 'set_attr':
         hlwm.call('set_attr clients.focus.keymask ^x$')
-    if order == 'keybinding_added_later':
+    if whenbind == 'added_later':
         hlwm.call('keybind x add tag2')
 
+    if refocus:
+        hlwm.create_client()
+        hlwm.call('cycle +1')
+        hlwm.call('cycle -1')
```
dnnr: Hold on, I think I made a mistake here. The new client is already focused, so after the two `cycle`s, we should end up on the new client again, which doesn't have the keymask, right?

So, does this test pass because of a bug in hlwm?

t-wissmann: New clients don't get the focus per default (only if `hc rule focus=on`, which is in the default autostart). So after the two `cycle`s, the client is focused for which the keymask has been added. So I think this is fine.

dnnr: Ah thanks, I didn't know that. :+1:

PR #309, opened 2019-02-03
Add tests for invalid regex in keymask

dnnr:


t-wissmann:
Looks good and I also agree with the comment

PR #312, opened 2019-02-03
Integrate completion for keyunbind into KeyManager

dnnr:


dnnr:
The build should pass as soon as #311 is merged.

PR #313, opened 2019-02-04
Use Link_ for pointers, Child_ for  ownership

t-wissmann:
Makes the two different applications of Child_ explicit. The new class
Symlink_<T> expresses a pointer to another object (of type T) and
a Child_<T> member expresses a child object (of type T) that is owned
(i.e. initialized and freed by the class with this Child_<T>-member).


t-wissmann:
The next step (in the next branch) is to add a `changed` signal to the `SymLink_` class. We can also discuss whether it makes sense to remember the initialization order of children in order to `reset()` them in the reverse order automatically.

src/symlink.h
```diff
@@ -0,0 +1,42 @@
...
+    T* operator()() {
+        return pointer;
+    }
+    T* operator->() {
+        return pointer;
+    }
+private:
+    Object& parent_;
+    std::string name_;
+    T* pointer = nullptr;
```
dnnr: Should be called `pointer_`, I think

src/child.h
```diff
@@ -2,42 +2,41 @@
...
+#include <memory>
 
 template<typename T>
 class Child_ {
-    // implement a pointer to a child object
-    // in the object tree.
-    // if this is assigned a new value, then the child object
-    // in the owner is updated automatically
+    // implement a static child object in the object tree. Static means that
+    // init() is called on start up and reset() is called on shutdown.
```
dnnr: If you're changing the comment anyway, you could turn it into a proper Doxygen docstring by putting it in front of the class declaration in this format:
```
/*!
 * ...
 */

File: src/root.cpp [hunk 3]
```diff
@@ -59,17 +59,17 @@ Root::~Root()
     tags()->setMonitorManager({});
 
     // Note: delete in reverse order of initialization!
-    delete monitors();
-    delete clients();
+    monitors.reset();
```
dnnr: Aren't those redundant because it's going to happen in the destructor of `Child_` anyway (because `pointer` is now a `unique_ptr`?

I don't really mind offering `Child_::reset()`, but I'm wondering if there's ever an actual need for it.

That being said, I'm hugely in favor of dropping the `operator=` in `Child_` (as you did), and being more explicit in that it is a wrapper around `unique_ptr`.

t-wissmann: In root.cpp, the destruction order matters and this is why I'd like to make it explicit. I'm optimistic that we can get rid of this at some point.

t-wissmann:
I've incorporated your suggestions an renamed `SymLink_` to `Link_`. And again, the commit messages aren't used in the squash-merge commit message on github...

src/root.cpp
dnnr: :+1:

ypnos:
Do I understand correctly that `init()`, as supposed to constructor arguments, is also only needed because of the initialization order in Root?

The benefit of not having init, nor reset, would be that we can always count on a Client object being valid. Otherwise we would theoretically always need to check. Does not apply to a Link, so one would know where dealing with links or not (better yet, have an explicitely different interface).

Would we want to have this property eventually?

ypnos:
Regarding the link, when no client is focused you could delete the Link instead of resetting it, to fulfill the same property.

PR #314, opened 2019-02-06
FrameTree class for global actions on the entire tree

t-wissmann:
The FrameTree class is intended to be a common entry point for other modules that want to modify the tree. So FrameTree keeps track of the entire frame tree (insert clients, ask for the focused frame, ask for the focused client). Formerly, these methods (e.g. lookup, dump, getFocusedFrame) were part of three classes (HSFrame, HSFrameLeaf, HSFrameSplit). Instead, they are now only a single method in FrameTree.

Additionally, the frame commands (shift, split, remove frame,...) will become methods in FrameTree. (They are still global functions in layout.h/layout.cpp)

### DIFF
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -47,3 +47,19 @@ def test_focus_wrap(hlwm, running_clients, running_clients_num):
         assert int(hlwm.get_attr('tags.0.curframe_windex')) == expected_idx
         if i < running_clients_num - 1:
             hlwm.call('focus down')
+
+
+@pytest.mark.parametrize("path", '@ 0 1 00 11 . / /. ./'.split(' '))
+@pytest.mark.parametrize("running_clients_num", [3])
+@pytest.mark.parametrize("num_splits", [0, 1, 2])
+def test_dump(hlwm, running_clients, path, running_clients_num, num_splits):
+    for i in range(0, num_splits):
+        hlwm.call('split explode')
+    layout = hlwm.call('dump').stdout
+    layout_part = hlwm.call('dump "" ' + path).stdout
+    if num_splits > 0:
+        assert layout_part in layout
+        assert len(layout_part) < len(layout)
+    else:
+        assert layout_part == layout
+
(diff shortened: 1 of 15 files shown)


t-wissmann:
In the work-in-progress PR #315 I try to show how the FrameTree class can be used to implement the commands that act on the tiling tree. (Currently there's only `cycle` but more will follow).

tests/test_layout.py
```diff
@@ -47,3 +47,19 @@ def test_focus_wrap(hlwm, running_clients, running_clients_num):
...
             hlwm.call('focus down')
+
+
+@pytest.mark.parametrize("path", '@ 0 1 00 11 . / /. ./'.split(' '))
+@pytest.mark.parametrize("running_clients_num", [3])
+@pytest.mark.parametrize("num_splits", [0, 1, 2])
+def test_dump(hlwm, running_clients, path, running_clients_num, num_splits):
+    for i in range(0, num_splits):
+        hlwm.call('split explode')
+    layout = hlwm.call('dump').stdout
```
dnnr: Could you add blank lines around the "act" phase? (i.e., before the first dump call and after the second dump call)

tests/test_layout.py
```diff
@@ -47,3 +47,19 @@ def test_focus_wrap(hlwm, running_clients, running_clients_num):
         assert int(hlwm.get_attr('tags.0.curframe_windex')) == expected_idx
         if i < running_clients_num - 1:
             hlwm.call('focus down')
+
+
+@pytest.mark.parametrize("path", '@ 0 1 00 11 . / /. ./'.split(' '))
+@pytest.mark.parametrize("running_clients_num", [3])
+@pytest.mark.parametrize("num_splits", [0, 1, 2])
```
dnnr: Uh nice, I wasn't aware that it would work like this. I've stolen this for the keymask tests (see #320)

tests/test_layout.py
t-wissmann: I've done so. Here I'm not quite sure what the 'act' phase is. It's clear that `dump` is tested here, but actually `dump` only observes but does not act.

PR #315, opened 2019-02-06
Move frame commands to FrameTree

t-wissmann:
Move tiling commands to the FrameTree class.


t-wissmann:
This includes #314 and this serves as a motivation why we need #314.

PR #316, opened 2019-02-06
Client: Call ensureKeymask() only if focused

dnnr:


t-wissmann:
Where does this come from? Is this a code optimization or does this fix a bug?

dnnr:
It's not a bug, it's just more correct. You could take it as an optimization (avoiding pointless calls), but I guess it serves a small but more important purpose: Making it more obvous *why* the call is being done in the first place. (the answer being: because if the keymask of the focused client changes, that change must be applied immediately)

t-wissmann:
I see, thanks!

PR #317, opened 2019-02-06
Clean up member order/grouping in keycombo.h

dnnr:


dnnr:
Sorry about this tiny PR, I just wanted keep the noise out of #319

PR #326, opened 2019-02-10
Fix command completion: Use Complete::none()

dnnr:


dnnr:
This PR is intended as a "shouldn't it be like this?" question, because I'm not entirely sure myself. Opening a PR seemed like the easiest medium to ask. ;-)

t-wissmann:
Right! Maybe we should rename the methods, e.g. `ifParameterExpected()`. Or we should solve this differently....

dnnr:
I'd say let's merge this first, regardless of the planned renames.

PR #327, opened 2019-02-10
Implement completion for "keybind" in KeyManager

dnnr:


dnnr:
This is [WIP] because of an unsolved issue with command completion (see FIXMEs)

PR #329, opened 2019-02-10
Rename Stack members to camelCase

dnnr:


dnnr:
Note that I've fixed a (supposed?) typo along way (`raise_slide` -> `raiseSlice`)

PR #330, opened 2019-02-10
Add MouseManager scaffolding, replace init/destroy

dnnr:


src/mouse.cpp
```diff
@@ -27,7 +27,7 @@ static Client*        g_win_drag_client = nullptr;
 static Monitor*       g_drag_monitor = nullptr;
 static MouseDragFunction g_drag_function = nullptr;
 
-static Cursor g_cursor;
+Cursor g_cursor;
```
t-wissmann: Please make this a member of MouseManager. `g_cursor` is only used in `mouse_init` and `mouse_destroy`.

dnnr: I wanted to do as little as possible along with the introduction of the class, that's why I left it. But you're right, moving it inside is better than making it temporarily global.

tests/conftest.py
```diff
@@ -191,6 +191,23 @@ def wait_for_window_of(self, wmclass):
                 'waiting for hook triggered by client \"{}\"'.format(wmclass))
         return line[-1]
 
+    def wait_queue_empty(self, timeout=5):
```
dnnr: Neat trick! :+1: 
Just some nitpicks:
 * The `ps_cmd` is only used once and its definition is unnecessarily far away from its point-of-use. I guess it came from a habit of optimization, but I don't think that makes sense here and I'd rather move it right before the `Popen` call.
 * Is there `print` there on purpose? It looks like a leftover debug statement.
 * Maybe `wait_until_idle` makes a bit more sense as a method name? I don't mind `wait_queue_empty` though... just an idea.
 * Would it be wrong to call this at the end of every method in the `keyboard` and `mouse` fixtures instead of making the test case take care of it?

t-wissmann: Yes, `wait_until_idle` is much better. The print is a leftover debug statement. We can also use this in the `keyboard` and `mouse` fixtures. This only means that these two fixtures need the `hlwm` fixture as a dependency.
But I think the main issue is: the CI still does not succeed and I don't understand why.

dnnr: True. I don't get it either...

Maybe, just out of desperation, make it sleep for at least 2 seconds? Just to make sure that it actually _is_ a race condition?

(Having `hlwm` as dependency in `keyboard` and `mouse` shouldn't be an issue.)

PR #331, opened 2019-02-10
Add basic set of tests for mousebind/mouseunbind

dnnr:


tests/test_mousebind.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.mark.parametrize('button', [1, 2, 3, 4, 5])
+def test_trigger_mouse_binding_without_modifier(hlwm, mouse, button):
+    hlwm.call(f'mousebind Button{button} call close')
+    client_id, _ = hlwm.create_client()
+
+    mouse.click(str(button), client_id)
+
+    assert hlwm.get_attr('tags.0.client_count') == '0'
```
t-wissmann: This is very racy! `close` does not close the window but sends a request to the client to close its window. So it is very likely, that the next hlwm-command in the queue (the `get_attr`) is fully executed before the client fully handles the close-request and long before the notification about the closed window reaches the event queue in hlwm.

If you want to test the binding, you can create a custom attribute (`new_attr string my_test`) and set this attribute in the mouse binding (`set_attr my_test something`), in order to assert its value

tests/test_mousebind.py
```diff
@@ -0,0 +1,35 @@
...
+@pytest.mark.parametrize('button', [1, 2, 3, 4, 5])
+def test_trigger_mouse_binding_with_modifier(hlwm, keyboard, mouse, button):
+    hlwm.call(f'mousebind Mod1-Button{button} call close')
+    client_id, _ = hlwm.create_client()
+
+    keyboard.down('Alt')
+    mouse.click(str(button), client_id)
+    keyboard.up('Alt')
+
+    assert hlwm.get_attr('tags.0.client_count') == '0'
```
t-wissmann: Here's the same race condition. You can even create two bindings:
```
hlwm.call('new_attr string my_test')
hlwm.call(f'mousebind Button{button} call set_attr my_test nomods')
hlwm.call(f'mousebind Mod1-Button{button} call set_attr my_test withmods')
```
and then you can assert that the right binding has been used:
```
assert hlwm.get_attr('my_test') == 'withmods'
```

tests/test_mousebind.py
dnnr: Thanks for noticing! The `set_attr` approach is a good idea, I'll change it.

dnnr:
It still failing, even though it works on my machine. I suppose we still have a race-condition between sending the button press event and the `get_attr` check?

tests/conftest.py
```diff
@@ -284,4 +284,24 @@ class KeyBoard:
...
+            subprocess.call(['xdotool', 'keyup', key_spec])
+
     return KeyBoard()
+
+
+@pytest.fixture()
+def mouse():
+    class Mouse:
+        def move_into(self, win_id):
+            subprocess.call(f'xdotool mousemove --window {win_id} 1 1', shell=True)
```
t-wissmann: Here, we should call `mousemove --sync …`. I'm not sure whether this solves the race-condition.

dnnr: Could be, and makes sense anyway. Let's see how it goes.

dnnr: Does not seem to help...

tests/test_mousebind.py
```diff
@@ -0,0 +1,40 @@
...
+    hlwm.call(f'mousebind Mod1-Button{button} call set_attr my_press yup')
+    hlwm.call(f'mousebind Button{button} call remove_attr my_press')  # canary bind (should not trigger)
+    client_id, _ = hlwm.create_client()
+
+    keyboard.down('Alt')
+    mouse.click(str(button), client_id)
+    keyboard.up('Alt')
+
+    assert hlwm.get_attr('my_press') == 'yup'
+    hlwm.call('remove_attr my_press')  # avoids memory leak (TODO: plug the leak)
```
t-wissmann: We can remove this, once #339 is merged.

dnnr:
Alright, I've pushed the stderr matching mechanism that we discussed. This is the first time running it on Travis (which is the only place I've seen the race-condition actually fail), so let's see how this goes.

tests/conftest.py
```diff
@@ -191,6 +199,22 @@ def wait_for_window_of(self, wmclass):
                 'waiting for hook triggered by client \"{}\"'.format(wmclass))
         return line[-1]
 
+    def wait_queue_empty(self, timeout=5):
+        """Wait for the hlwm process to be in sleep state"""
+        pid = self.hlwm_process.proc.pid
+        ps_cmd = 'ps --no-headers -o state -p {}'.format(pid)
```
The-Compiler: How platform dependent is this? Maybe consider using [pslib](https://psutil.readthedocs.io/en/latest/) instead?

tests/conftest.py
```diff
@@ -223,9 +248,77 @@ def __init__(self, tmpdir, env):
...
+                key.data.write(ch)
+                key.data.flush()
+
+                # Store in temporary buffer for string matching:
+                if key.fileobj == self.proc.stderr:
+                    stderr += ch
+                if key.fileobj == self.proc.stdout:
+                    stdout += ch
+
+            if selected != []:
```
The-Compiler: You can just do `if selected:`

dnnr: Somehow, I *knew* you we're going to say this. ;-)
I did this on purpose to make it more obvious that `selected` is either a non-empty list (implying that the `for` loop had something to iterate over) or it's empty. It's not `False`, it's not `None`, it's exactly `[]` *or* a non-empty list.

I'm aware that technically there is no difference, but I'm communicating to future contributors here, not to the interpreter. I don't want to make them look up the selectors API just because they wonder what else could be in `selected`.

t-wissmann:
Right, it sounds like we are interpreting the travis output in a completely wrong way...

tests/conftest.py
The-Compiler: My $0.02: `if not selected:` clearly says "if nothing is selected". I've had to read `if selected != []:` a few times (because it feels like yoda reading a double negotiation or something) until I was sure I understood it the right way.

dnnr:
Alright, I'm fairly confident that testing with mouse buttons 4 and 5 was the underlying cause of the failures that we saw (see comment in the code).

How should we proceed with regards to the stderr matching and `wait_queue_empty`? For this situation, I prefer the stderr matching for being more precise. That said, I see potential value for the process state checking in other scenarios.

t-wissmann:
Yes, I also prefer the stderr matching. I also suggest to revert/drop my commits that introduced `wait_queue_empty`.

PR #337, opened 2019-02-10
Add CI check to enforce "using std::..." style

dnnr:
This only enforces it for string (for now), to allow a gradual transition.


dnnr:
This build should pass as soon as #336 and #337 are merged.

ypnos:
Great stuff

PR #339, opened 2019-02-11
Automatic cleanup of RootCommands and its user defined attributes

t-wissmann:


tests/test_root_commands.py
```diff
@@ -131,3 +131,19 @@ def complete(partialPath):
     assert len(complete('monitors.focus.')) >= 8
     assert complete('t') == ['tags.', 'theme.', 'tmp.']
     assert complete('') == [l + '.' for l in hlwm.list_children_via_attr('')]
+
+@pytest.mark.parametrize('attrtype', ['int', 'bool', 'string', 'color', 'uint'])
+@pytest.mark.parametrize('name', ['my_test', 'my_foo'])
```
dnnr: It's not really related to this PR, but another simple but good test case you could add for `new_attr` is testing its behavior when the attribute name doesn't start with `my_`.

tests/test_root_commands.py
```diff
@@ -131,3 +131,19 @@ def complete(partialPath):
...
+@pytest.mark.parametrize('object_path', ['', 'clients', 'theme.tiling.active'])
+def test_new_attr(hlwm, attrtype, name, object_path):
+    path = (object_path + '.' + name).lstrip('.')
+
+    hlwm.call(['new_attr', attrtype, path])
+
+    hlwm.get_attr(path)
+
+    hlwm.call(['remove_attr', path])
```
dnnr: So this test is actually verifying parts of the behavior of three different things, `new_attr`, `get_attr` and `remove_attr`, right? It's obviously not an overly large test case, but the way it is now, I think it's just too easy to miss some interesting corner cases for the individual commands:
 * `new_attr`: What if the name isn't valid? (I already said that above)
 * `get_attr`: What if the attribute doesn't exist? (a test for that is already "hidden" in here)
 * `remove_attr`: What if the attribute doesn't exist? What if you try to remove a non-user-defined attribute?
 * And another one, that hasn't appeared yet `set_attr`: Can the custom attribute be written at all?

So, from test design point of view, I think it would make a lot of sense to start with a `test_new_attr_` prefix(!), think of which specific behavior could be tested there, and then continue this with the other two commands. I'd even consider putting this into a separate `test_attr.py`.

Sidenote: While writing this, I already discovered that the error message for attempting to remove a non-custom attribute is rather not helpful and should probably be improved (fwiw, it's worse in master). Things like that are one of the reasons I see value in these stupidly small test cases. It's not _just_ me being overly pedantic about simple behavior that is unlikely to actually break by accident.

t-wissmann: I think this only tests `new_attr` and `remove_attr`. And I don't claim to fully test them. Actually I only wanted to create a test case that failed before this PR and works after this PR. But I have just noticed that I messed this up (in order to check the fix of this PR, I must not remove the user-defined attribute again).

dnnr: Yeah, these kinds of tests (for the absence of a previously known bug) are sometimes at odds with how you'd put the tests if the bug had never existed. And don't get me wrong, I didn't assume any claim about the completeness of the test, it just looked to me like the foundation of a messy set of tests. Maybe mostly because of the generic name.

How about turning this into a specific "test for the absence of a memory leak with custom attributes" test and naming it accordingly? After all, having a leak due to entities created at runtime is not far-fetched at all (even if we didn't actually have this bug).

t-wissmann: Yes, I've clarified the name of the test case and fixed the testcase to test the cleanup during hlwm shutdown.

src/root.h
```diff
@@ -45,8 +47,8 @@ class Root : public Object {
     Child_<Theme> theme;
     Child_<Tmp> tmp;
 
-    RootCommands* root_commands;
     Globals globals;
+    std::unique_ptr<RootCommands> root_commands;
```
ypnos: Is there a reason for this to be a unique_ptr instead of a plain member object? It gets initialized at the beginning of the constructor and is not explicitely deleted as well.

t-wissmann: I don't want to include `rootcommands.h` in `root.h` (which is included in many cpp-files at the moment). So I want to keep the dependencies between include files at a minimum.

ypnos: I see. Maybe that's worth a comment at that point? E.g. "// wrapped in unique_ptr to avoid include"
As rootcommands.h by itself doesn't bring more baggage in form of includes, I wouldn't expect this particular include to matter though.

t-wissmann: And as I just found it, this is called _pimpl_-idiom https://www.fluentcpp.com/2017/09/22/make-pimpl-using-unique_ptr/

ypnos: Yes I heard about this before, from Qt I guess. Even a comment "// pimpl" would help though. When I see a smart ptr, I assume there must be some other reason to it.

dnnr: Wouldn't `optional` work here as well (and be slightly more correct)?

Beside that, I'm alson in favor of putting a comment here. I'd write `// Using "pimpl" to avoid huge #include`

ypnos: I like your suggestion, however I feel unique_ptr is indeed the better choice here. As the lifetime of the object here is synchronized with the lifetime of its owner. Optional invites you to always check if its there.

dnnr: You mean this as a semantic distinction, right? Because technically, they're both almost the same (esp. with regards to lifetime and needing to check validity).

In any case, I don't mind unique_ptr either. Go with what you prefer. :-)

ypnos: Yes.

t-wissmann: I've added the comment. I think `optional` is less correct here, because the root commands object here is mandatory (during the lifetime of Root) and not optional.

PR #340, opened 2019-02-16
Test new_attr and remove_attr commands

t-wissmann:


tests/conftest.py
```diff
@@ -84,6 +85,9 @@ def call_xfail(self, cmd):
         proc = self.unchecked_call(cmd)
         assert proc.returncode != 0
         assert proc.stderr != ""
+        def f(self2, reg):
```
dnnr: Maybe explain a bit what is going on here. At least I didn't find it obvious.
If I'm getting it right, how about: "Extend the Popen object with a match() method that runs a regex against the process sterr"

t-wissmann: Right! I've added a docstring explaining this.

tests/conftest.py
```diff
@@ -84,6 +85,9 @@ def call_xfail(self, cmd):
...
+            assert re.search(reg, self2.stderr)
+        proc.match = types.MethodType(f, proc)
```
dnnr: I think you can simplify it a bit like this:
```
def f(reg):
    assert re.search(reg, proc.stderr)
proc.match = f
```

tests/test_root_commands.py
```diff
@@ -1,4 +1,13 @@
 import pytest
+import re
+
+ATTRIBUTE_TYPES = ['int', 'bool', 'string', 'color', 'uint']
+# example values for the respective types
+ATTRIBUTE_VALUES = [[23, 42, -8],
+                    ['true', 'false'],
+                    ['foo', 'baz', 'bar'],
+                    ['#ff00ff', '#9fbc00'],  # FIXME: include named colors
+                    [23, 42]]
```
dnnr: Why not structure these as a dictionary? Having two independent lists with some implicit relation between their orders is just unnecessarily hard to read and maintain, in my opinion.

I suspect you did it like this so that you can use the globals for parametrization (as-is), but it don't immediately see why it wouldn't work with a dictionary's `.items()` list as well.

tests/test_root_commands.py
```diff
@@ -132,12 +141,81 @@ def complete(partialPath):
...
+def test_new_attr_without_removal(hlwm, attrtype, name, object_path):
     path = (object_path + '.' + name).lstrip('.')
 
     hlwm.call(['new_attr', attrtype, path])
 
     hlwm.get_attr(path)
+
+
+@pytest.mark.parametrize('attrtype', ATTRIBUTE_TYPES)
+def test_new_attr_existing_attribute(hlwm, attrtype):
```
dnnr: Is there any better name for "not a user attribute"? Maybe "static", or "builtin"?

In order to avoid confusion with "existing user attribute", I'd rather avoid using "existing" to refer to "builtins" as well.

tests/test_root_commands.py
```diff
@@ -132,12 +141,81 @@ def complete(partialPath):
...
+@pytest.mark.parametrize('path', ['my_foo', 'monitors.my_bar'])
+def test_new_attr_writable(hlwm, attrtype, values, path):
+    hlwm.call(['new_attr', attrtype, path])
+    for v in values:
+        hlwm.call(['set_attr', path, v])
+        assert hlwm.get_attr(path) == str(v)
+
+
+@pytest.mark.parametrize('attrtype', ATTRIBUTE_TYPES)
+def test_new_attr_right_type(hlwm, attrtype):
```
dnnr: Just a hint: Don't be afraid to use verbs in unit test names. ;-) Those test names are totally fine, but "test_new_attr_has_right_type" and "test_new_attr_is_writable" would be even better, in my opinion.

tests/conftest.py
t-wissmann: I strongly prefer my code as it is, because 1) closures in pythons are difficult and 2) it will become a member function anyway, so it's fine to have a self/this argument.

tests/conftest.py
```diff
@@ -82,11 +82,17 @@ def call(self, cmd):
         return proc
 
     def call_xfail(self, cmd):
+        """call the command and expect it to have non-zero exit code
+        and some output on stderr. The returned finished process handle is
+        extende by a match() method that runs a regex against the process sterr
```
dnnr: Typo "extended"

tests/test_root_commands.py
t-wissmann: Right, a dictionary is better

tests/conftest.py
dnnr: One of those is mine :D

dnnr:
Looks good to me!

PR #342, opened 2019-02-17
Travis: Call pytest with -v

dnnr:
This makes it much easier to spot correlations between failing and
suceeeding tests.


t-wissmann:
Is fine for me.

PR #344, opened 2019-02-19
Use proper PCRE look behind for std:: prefix check

dnnr:


dnnr:
This needs #343 to be merged in order to pass

t-wissmann:
As an alternative to the PCRE, we can replace the grep line by two grep calls:
```
grep --color=always -nv 'using [ ]*std::' src/*.cpp  | grep --color=auto "std::$symbol\>"
```
Do you prefer the two-grep-calls or the PCRE?

t-wissmann:
Oh and if you prefer the PCRE: you're passing files to grep, so I don't see why you need the `-r` in your grep call.

dnnr:
I prefer the PCRE variant for its brevity. But you're totally right about `-r`, I've removed it. :+1:

PR #345, opened 2019-02-20
Remove std:: prefixes for smart pointer symbols

dnnr:
Along the way, turn the CI check blacklist into a whitelist of remaining
symbols.

### DIFF
--- a/ci-check-using-std.sh
+++ b/ci-check-using-std.sh
@@ -14,20 +14,18 @@ set -o nounset
 reporoot=$(realpath "$(dirname "$0")")
 
 # Assemble list of std symbols to pull in:
+# (exclude some that are not enforced yet)
 usethis=$(grep --no-filename 'using std::' "$reporoot/src/"*.cpp \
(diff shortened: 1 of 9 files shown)


ci-check-using-std.sh
```diff
@@ -19,7 +19,15 @@ usethis=$(grep --no-filename 'using std::' "$reporoot/src/"*.cpp \
...
-usethis="string vector"
+usethis=" \
+    make_shared \
+    make_unique \
+    shared_ptr \
+    string \
+    unique_ptr \
+    vector \
+    weak_ptr \
+    "
```
t-wissmann: You don't need the `\` here because bash is waiting for the final `"` anyway.

dnnr: True, thanks. But actually... let me just turn this into a whitelist instead. There's not too many symbols left.

t-wissmann: Is also fine for me!

PR #349, opened 2019-02-22
Turn g_mouse_binds into vector in MouseManager

dnnr:


dnnr:
This includes #347

src/mouse.cpp
```diff
@@ -187,14 +183,14 @@ int mouse_bind_command(int argc, char** argv, Output output) {
...
+    MouseBinding mb;
+    mb.button = button;
+    mb.modifiers = modifiers;
+    mb.action = function;
     for (int i = 3; i < argc; i++) {
-        mb->cmd.push_back(argv[i]);
+        mb.cmd.push_back(argv[i]);
     }
-    g_mouse_binds = g_list_prepend(g_mouse_binds, mb);
+    Root::get()->mouse->binds.push_back(mb);
```
dnnr: I was assuming here that there is no significance to the order (prepend vs. append).

t-wissmann: Oh right. It should be prepend actually! By preprending, I implicitely shadow an old binding to the same Modifier-Button-Combination

dnnr: Now using an std::list and push_front() instead

dnnr: I've also added a test confirming the behavior of overlapping bindings.

t-wissmann: Great!

PR #351, opened 2019-02-23
Integrate "mousebind" completion into MouseManager

dnnr:
This eliminates the need for the last remaining code in key.h/key.cpp,
as well as some completion helpers.


dnnr:
This includes #350

src/mousemanager.cpp
```diff
@@ -12,6 +27,117 @@ MouseManager::MouseManager() {
...
+            "Button2",
+            "Button3",
+            "Button4",
+            "Button5",
+            "B1",
+            "B2",
+            "B3",
+            "B4",
+            "B5",
+        };
```
dnnr: Note that this list is not intended to remain forever, it's just the easiest way for now.

src/mousemanager.cpp
```diff
@@ -12,6 +27,117 @@ MouseManager::MouseManager() {
...
+        for (auto button : buttons) {
+            complete.full(normNeedle + button);
+        }
+    } else if (complete == 1) {
+        complete.full({"move", "resize", "zoom", "call"});
+    } else if (complete[1] == "call") {
+        complete.completeCommands(2);
+    }
```
t-wissmann: Here, you can add:
```
else {
   complete.none();
}
```

PR #354, opened 2019-02-23
Add CI job for running IWYU checks

dnnr:
For now, this only fails if the order of includes is not correct.

The actual IWYU check results are printed but do not lead
to build failures.


dnnr:
Until #353 is merged, the build will fail.

iwyu.sh
```diff
@@ -0,0 +1,45 @@
...
+iwyu_ver_maj=$(echo "$iwyu_ver" | cut -d. -f 1)
+iwyu_ver_min=$(echo "$iwyu_ver" | cut -d. -f 2)
+
+if ! (( iwyu_ver_maj > 0 || iwyu_ver_min >= 11 )); then
+    echo >&2 "Error: Expecting include-what-you-use version >= 0.11, got $iwyu_ver"
+    exit 1
+fi
+
+tmpdir=$(mktemp -d)
+trap "{ rm -rf $tmpdir; }" EXIT
```
t-wissmann: Do we really need this? The combination of `rm -rf` together with an unquoted variable looks **very fragile**...

t-wissmann:
Can we extract a patch/commit from the build log? E.g. can we configure `iwyu` to actually modify the source code, and then we can `git commit -m "Fix include order"` and `git format-patch --stdout HEAD~1`

iwyu.sh
dnnr: Strictly speaking it's not necessary, but useful for running the script manually, outside of a throwaway Docker container. In bash, it's not as easy to do as in other languages, but I'd rather not leave temporary files behind.

You're right about the quoting though, I've fixed it.

dnnr:
It is possible to tell IWYU to simply make the suggested changes (but it often breaks).
However, as soon as the existing complaints are fixed and the script actually makes the build fail, I don't expect us to need it ever again. At that point, you're probably looking at change suggestions that are short enough to be applied manually.

And besides that, the IWYU suggestions are not perfect and not always correct. We might need to tune it with a custom mapping configuration. So my plan was to apply its suggestions in smaller batches with manual review and see how far we can go.

t-wissmann:
Sounds convincing.

PR #357, opened 2019-02-25
Implement the cycle_all command

t-wissmann:
Make the cycle_all command functional again and add test cases for it.
Also implement cycle_frame, which is a subtask of cycle_all.


tests/test_layout.py
```diff
@@ -134,3 +134,66 @@ def test_layout_command(hlwm, running_clients, running_clients_num, num_splits):
     lines = [re.sub(r' [0-9]+% selection=[01]', '', l) for l in lines]
     assert dumped == '\n'.join(lines).replace(':', '')
 
+
+@pytest.mark.parametrize("running_clients_num,client2focus",
+            [(cnt, idx) for cnt in [3, 5] for idx in range(0, cnt)])
```
dnnr: Maybe I'm missing something here, but wouldn't it be simpler to do this with two `@pytest.mark.parametrize` decorators? It's a cartesian product of [3,5] and range(0, cnt), right?

t-wissmann: It's not a cartesian product, because `cnt` is not fixed by the first component, i.e. an element of `[3,5]`. Can I let one `@pytest.mark.parametrize` depend on another `@pytest.mark.parametrize`?

t-wissmann: The answer is: no. You can have dependencies between fixtures, but not between parameters. The following code fails:
```
@pytest.mark.parametrize("running_clients_num", [3, 5])
@pytest.mark.parametrize("client2focus", range(0, running_clients_num))
@pytest.mark.parametrize("num_splits", [0, 1, 2, 3])
def test_jumpto_within_tag(hlwm, running_clients, client2focus, num_splits):
    assert 0 == 0
```
It fails with:
```
_______________________ ERROR collecting tests/test_layout.py _______________________
tests/test_layout.py:139: in <module>
    @pytest.mark.parametrize("client2focus", range(0, running_clients_num))
E   NameError: name 'running_clients_num' is not defined
!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!
```

dnnr: Oh well, seems like i *was* missing something. ;-) Then I guess it's fine.

src/frametree.cpp
```diff
@@ -399,3 +399,21 @@ void FrameTree::cycle_frame(int delta) {
     focusFrame(frames[index]);
 }
 
+int FrameTree::cycle_frame_cmd(Input input, Output output) {
```
dnnr: Does this purposely deviate from our (new) naming scheme for members (`camelCase`)? If not, I'd suggest `cycleFrameCommand`. :-)

t-wissmann: I'm having trouble when I have to decide between naming the function like the command  and naming the function in camel case. Shall we agree on camel case as the naming convention? If yes, I suggest to fix the suffix `Cmd` for commands.

dnnr: I'm fine with `Cmd` as suffix, but keep in mind that all of the command functions that I have touched since we agreed on camelCase having been suffixed with `Command` (see RuleManager, TagManager, KeyManager, MouseManager). If you prefer `Cmd`, we should change it everywhere else, too. That should not stop us, though.

dnnr: Oh and: I've struggled with naming those methods, too. I ended up rather ignoring the command name when choosing a method name. So we have, for example, `addKeybindCommand` for `hc keybind` and `removeKeybindCommand` for `hc keyunbind`. Made more sense to me.

t-wissmann: I've renamed the commands in `FrameTree` to camelCase and added a `Command` suffix.

PR #358, opened 2019-02-26
Speed up CI infra with Docker and lots of caching

dnnr:


.travis.yml
```diff
@@ -1,91 +1,59 @@
...
+                  - $HOME/.ccache-clang-7
+                  - $HOME/.ccache-gcc-8
+                  - $HOME/.tox-cache
           script:
-              - cmake -G'Ninja' -DCMAKE_BUILD_TYPE=Debug . # run tests with sanitizer
-              - ninja -v -k 10
-              - tox -- -n auto -v
+              # Prepare Docker image with bleeding edge Ubuntu 19.04 (disco)
+              - docker pull dnnr/hlwm-ci:disco
+              - docker build -t disco --cache-from dnnr/hlwm-ci:disco - <Dockerfile.ci-disco
```
t-wissmann: The docker file used here is independent from the docker file that you introduce with this PR, right?

.travis.yml
```diff
@@ -1,91 +1,59 @@
...
-            - herbstclient quit
-            - wait
+              # Run builds and tests in Docker container
+              - docker run -u $UID --privileged -t --rm
+                --volume=/etc/passwd:/etc/passwd:ro
+                --volume=$HOME/.ccache-gcc-8:$HOME/.ccache-gcc-8:rw
+                --volume=$HOME/.ccache-clang-7:$HOME/.ccache-clang-7:rw
+                --volume=$PWD:/hlwm:rw
+                --volume=$HOME/.tox-cache:/hlwm/.tox:rw
+                disco
```
t-wissmann: What is disco?

.travis.yml
dnnr: No, it's exactly that file.

But actually, you reminded me of something: After all this fiddling around (see commits in this branch...) I forgot to reintroduce the IWYU check. The `Dockerfile.iwyu` will end up being removed then.

.travis.yml
dnnr: It's the name of the upcoming Ubuntu release: Disco Dingo

dnnr: Oh and in case that wasn't the question: It's also the name of the temporary Docker image that is being used for this `docker run` invocation.

dnnr:
It seems that the persistent ccaches do not yet behave the way they're supposed to. Apparently, Travis treats PR builds a bit different than the branch builds I tested this on.

.travis.yml
```diff
@@ -1,91 +1,59 @@
...
+              directories:
+                  - $HOME/.ccache-clang-7
+                  - $HOME/.ccache-gcc-8
+                  - $HOME/.tox-cache
           script:
-              - cmake -G'Ninja' -DCMAKE_BUILD_TYPE=Debug . # run tests with sanitizer
-              - ninja -v -k 10
-              - tox -- -n auto -v
+              # Prepare Docker image with bleeding edge Ubuntu 19.04 (disco)
+              - docker pull dnnr/hlwm-ci:disco
```
dnnr: Note that this is pulling a reference image from my personal Docker Hub repository.
We'll probably want to switch to a generic org if we're going to merge this (I'm already squatting `hlwm` for this).

dnnr: Btw, because it's not really obvious: The images at `dnnr/hlwm-ci` get built automatically by Docker, through a connection with my GitHub repo. So while this *is* an external artifact, it does not need to be maintained manually.

dnnr:
If anybody's interested in the numbers: this brings down the build duration to just under 5 minutes if ccache has 100% hit rate.

It's not directly comparable to the existing numbers, because:
1. The new setup is not doing all the same things:
  * no documentation gets built; nor are the dependencies for that installed (we'll need to discuss this)
  * there is one less configuration being built because it was mostly redundant (Ubuntu 16.04, gcc-7)
2. There is less parallelism than before, so it's unclear what to compare to. In the existing setup, the total machine time consumed by a build ranges around to 20 to 25 minutes. The real duration is more like 8 to 13 minutes. The reason for reducing the parallelism is to avoid repeating the same costly steps on multiple machines (e.g., pulling the Docker image). I'm guessing that Travis takes the total machine time into account for scheduling decisions, so I'm hoping that being a good citizen™ pays off in the end.

Factoring out these limitations, I'm guesstimating we're still looking at a solid improvement (~2x).

As a another positive side-effect, this rework sets us up for being more flexible in our choice of CI providers. I've already experimented with CircleCI, which is vastly superior (imho) but has a strict quota for FOSS accounts (1000 minutes per month). I'm thinking that we could dynamically downgrade from CircleCI to Travis within each month, without too much added effort for maintaining configs for two providers.

PR #368, opened 2019-03-04
Turn process match() into assertable says()

dnnr:


t-wissmann:
I'm not really a fan of being more verbose. What is unclear about the following?
```
hlwm.call_xfail('remove_attr monitors.count') \	
        .match('is not a user defined attribute')
```
I read this as: 1. call it, 2. expect it to fail, 3. this matches the specified string.

On the other hand, I don't like the new semantics because if I forget the `assert` in
```
call = hlwm.call_xfail('remove_attr monitors.count')
call.says('is not a user defined attribute')
```
then the test case passes even if the call did not say "is not a user defined attribute".

The goal when designing APIs for me is to abstract away programming patterns that appear very often. And this explicit assert when checking the error message of a failed command is such a pattern.

dnnr:
I'm looking at the part about "forgetting the assert" from the other way around: This way, the `assert` is no longer hidden away within a call that is not *obviously* an assertion.
After all, judging only from the caller' point of view, `match()` could mean plenty other things than `assert_that_stderr_matches()`. For instance, Python's `re.match()` doesn't assert anything either, just returns the result.

While writing this, I even thought about going a bit further like this:
```
call = hlwm.call('something wrong')
assert call.fails()
assert call.says('meh')
```
It might be more lines, but it's *so much* more expressive.

Another possible route would be to go with PyHamcrest's interface. Then it would look something like this:
```
call = hlwm.call('something wrong')
assert_that(call, fails_with('meh'))
```
That might combine the best of both. See here for more info: https://github.com/hamcrest/PyHamcrest

t-wissmann:
Sorry but I strongly disagree. I want to design the test cases in such a way that 1) *If a mistake is made, we notice it* and 2) I don't write reundand code over and over. You told me that in `call_xfail` the `xfail` is standard python test termonology for *expect it to fail*, so it's absolutely fine that there is an assert in it.

If you don't like the name `match` and have something better, I'm fine with it. What about `expect_stderr` or `mentioning`? I'm also fine with renaming `call_xfail` to something else like `call_and_expect_fail`.

t-wissmann:
Regarding that the separation of *action* and *verification*: for me `call_xfail` is not an action but already *verification*. For example, the above
```
hlwm.call_xfail('remove_attr monitors.count') \	
        .match('is not a user defined attribute')
```
Is not to be understood as "remove_attr runs and perfoms things and then we verify later that the result of these computations is some failure" but this test case is "no action is performed at all for the specified reason".

dnnr:
Alright, you raise valid points and even though I feel differently, I don't really want to convince you at all costs either. So I'm fine with retracting my change.

Regarding an alternative name for `match`, I like your suggestion `expect_stderr`. In fact, this whole PR started from my desire for a better name than `match`, which led to me to `says`, which made me think that, language-wise, returning a bool is more logical than making an assertion, which made me extract the `assert`. Long story short, `expect_stderr` addresses my original issue as well, albeit by going in the opposite direction. :+1:

PR #372, opened 2019-03-04
Run fix_include checker on all source files

dnnr:


dnnr:
The build will fail until #371 is merged

dnnr:
Note that I've changed it later to cover *all* source files. At this point, it was easier than adding the rest later.

PR #378, opened 2019-03-07
Use endl instead of \n

dnnr:


dnnr:
This is meant as a testing PR to see if mergify maybe only operates on new PRs.

PR #379, opened 2019-03-07
Mergify: Do not auto-merge PRs with "wip" label

dnnr:


dnnr:
Coincidentally, the WIP app is showing us right here what is not ideal about it: You basically cannot make a PR with "wip" appearing anywhere in the title. I guess I'll have to remove the condition for that app from .mergify.yml first.

t-wissmann:
I've just removed the WIP app. I think you need to push another commit to trigger a re-run of the checks.

PR #383, opened 2019-03-08
ci-build.py: Fail if IWYU finds redundancies

dnnr:


dnnr:
This will fail until it won't. :-)

dnnr:
Alright, so in case this wasn't already obvious, I'm proposing this middle ground: Have the CI build only fail if IWYU has any suggestions to *remove* an include or forward declaration.

After examining things a bit further, I found myself disagree more and more with IWYU about its suggestions tot *add* things. Often in relation with templates, it gets rather overzealous, wanting to add some of the transitive dependencies without clear reasons. (Such as adding vector when using a subclass of ChildByIndex_.)

*But:* While making all these non-controversial include cleanups this weekend, I found IWYU to be very agreeable with regards to removals. There is only one small catch (which could well be considered "working as intended"): Whenever removing an include as suggested by IWYU, a missing include might reveal itself elsewhere. Of course, fixing these is usually straightforward.

So in other words: My proposal is to accept the missing includes that already exist in our code base, but to always listen to IWYU when it detects redundant includes, and *then* also fix missing includes as it becomes necessary for a successful build. (If this happens to be basically exactly what @t-wissmann meant to propose in the other PR, then I probably just didn't get it back then.)

t-wissmann:
I absolutely agree with all you wrote! (Yes, this is what I meant in the other PR)

PR #384, opened 2019-03-08
Implement all suggestions by IWYU

dnnr:


dnnr:
You could accept this as-is, but it is also intended as a basis for discussion. I suppressed some stupid suggestions with new mapping file entries, but there are still some things that might be controversial.
I'm most suspicious about:
 * `root.h` now including all manager headers instead of using fwd decls (I expected this to cause issues with include cycles, but apparently, it doesn't)
 * all headers with template types being pulled in wherever there is an indirect interaction (this goes for our own headers as well as things like `<utility>` for pair)

If you want to see what exactly caused a specific suggestion, do the following:
 1. revert the change (e.g., remove an include)
 2. change `ci-build.py` to append `--verbose=3` to the IWYU invocation
 3. run IWYU and see the exact line where it considers the symbol in question to be used
 4. be confused because it's often really not obvious, especially with template types

(recommended one-long-liner for running IWYU via ci-build.py: `docker run -u $UID --rm -ti --volume=$PWD/.ccache-docker:/.ccache:rw --volume=$PWD:/hlwm:rw hlwm/ci:disco /hlwm/ci-build.py --cxx=clang++-7 --cc=clang-7 --build-type=Debug --iwyu --build-dir=/hlwm/build-docker --ccache`)

src/root.h
```diff
@@ -2,23 +2,22 @@
...
-class HookManager;
-class KeyManager;
-class MonitorManager;
-class MouseManager;
 class RootCommands;
-class RuleManager;
-class Settings;
-class TagManager;
-class Theme;
-class Tmp;
```
t-wissmann: Here, I disagree. Why do we need to include every manager header in root.h? It looks like IWYU does not understand, that we can define a `Child_<X>` member even if the class `X` is only forward declared.

src/rootcommands.h
```diff
@@ -6,12 +6,13 @@
...
 #include <functional>
 #include <memory>
+#include <string>
 #include <vector>
 
+#include "attribute.h"
 #include "types.h"
 
 class Root;
-class Attribute;
```
t-wissmann: Here, I also don't see why `class Attribute` should be replaced by `#include "attribute.h"`.

src/rules.cpp
```diff
@@ -3,8 +3,12 @@
 #include <X11/X.h>
 #include <X11/Xlib.h>
 #include <algorithm>
+#include <array>
 #include <cstdio>
+#include <exception>
+#include <ostream>
 
+#include "attribute_.h"
```
t-wissmann: Why do we need attribute_.h in rules.cpp?

src/tagmanager.cpp
```diff
@@ -1,15 +1,22 @@
 #include "tagmanager.h"
 
+#include <cstddef>
 #include <memory>
+#include <ostream>
+#include <stdexcept>
 
+#include "attribute_.h"
```
t-wissmann: In principle, whenever `attribute_.h` is included in a cpp-file, then something is wrong (except for `rules.cpp`, where it's fine).

src/root.h
dnnr: I suppose we can use a pragma to make it understand: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

I haven't looked into it, though.

t-wissmann:
Ah... I should have first read your comment before going through the code.

I didn't look at the reason why IWYU wants these particular changes, but since we disagree with IWYU, I think we should configure IWYU or disregard its recommendations?

src/root.h
t-wissmann: I can't find a pragma that fixes this here. And I'd avoid flooding our code just because IWYU does not get the template right (namely at every occurence of `Child_<X>` where `X` is only forward declared).

dnnr:
> I didn't look at the reason why IWYU wants these particular changes, but since we disagree with IWYU, I think we should configure IWYU or disregard its recommendations?

It's a trade-off decision, I think.

If we want IWYU to be really useful to us, we should follow *all* of its recommendations not only by convention but by enforcement (as a CI check). To achieve such a setup, we might want to compromise and accept some includes that don't appear ideal to us  (and also cannot be "configured away").

What we'd ultimately gain from that is that we'd rarely need to *think* about includes anymore while at the same time having an assurance that they are in fact maintained by a strict and consistent policy (even if not exactly our preferred policy).

As long as those compromises don't entail significant drawbacks, such as increased build durations or an annoyingly dense dependency graph between source files, I'm quite willing to live with an include policy that does not 100% match my own understanding.

That said, the first step is certainly always to see if we can convince IWYU to make a different recommendation.

t-wissmann:
IWYU has some document on this https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYUIsDifficult.md but they don't say how to solve this....

t-wissmann:
> > I didn't look at the reason why IWYU wants these particular changes, but since we disagree with IWYU, I think we should configure IWYU or disregard its recommendations?
> 
> It's a trade-off decision, I think.
> 
> If we want IWYU to be really useful to us, we should follow _all_ of its
> recommendations not only by convention but by enforcement (as a CI check).

To be clear: I don't care whether IWYU is useful to us. I want good code. If
the IWYU suggestions improve the code, it's fine to use IWYU as a CI check. If
IWYU suggests to include headers even though a forward declaration is enough
because we _don't use these headers_, we should not run IWYU.

> To achieve such a setup, we might want to compromise and accept some includes
> that don't appear ideal to us (and also cannot be "configured away").

> What we'd ultimately gain from that is that we'd rarely need to _think_ about includes anymore while at the same time having an assurance that they are in fact maintained by a strict and consistent policy (even if not exactly our preferred policy).

To be honest, I never experienced any inner struggle where to put which includes: If I
only needa forward declaration of a type, I just forward declare it. If I call
a function, then I include the header that declares this function.

The only purpose of IWYU that I see here is that if I remove a function call,
then I don't want to check whether the corresponding include can be removed. So
the primary purpose of IWYU is "include _only_ what you use".

The check that "enough files are included" is performed by the compiler. So I
don't see the reason at all why IWYU ever suggests to include a file that has
not been included yet in a compiling code base.

> As long as those compromises don't entail significant drawbacks, such as increased build durations or an annoyingly dense dependency graph between source files, I'm quite willing to live with an include policy that does not 100% match my own understanding.

Right, the "annoyingly dense depency graph" is my critisism. Currently, 11 files include root.h, and in this PR, 10 new hlwm headers are included in root.h. This means, that by this change, 110 unnecessary dependencies from .h to .cpp files are introduced.


> That said, the first step is certainly always to see if we can convince IWYU to make a different recommendation.

I've spend some time with the IWYU documentation and I still don't see a way
how to configure IWYU properly here. Can we simply configure IWYU to _never_ suggest adding includes?

dnnr:
> The check that "enough files are included" is performed by the compiler. So I
> don't see the reason at all why IWYU ever suggests to include a file that has
> not been included yet in a compiling code base.

This is a crucial point. I thought so too at first, but I'm not sure if this is actually correct. I might totally be on the wrong track here, but here's what I'm thinking:

A compiling code base does *not* imply that no include is missing. It just implies that all necessary declarations *happen to be available* where needed. It is entirely possible that this is only the case because there are redundant includes (i.e., forgotten and wrong) that accidentally ensure a successful build through their transitive includes.

In other words, a redundant include might *mask* the fact an include is actually missing. And that might  even be in some entirely different place, in case the redundant include is within a header file.

That's why I suspect you cannot separate the "include *only* what you use" part from the "include *everything* you use". You need to do the latter so that you can consistently do the former.

Of course, all this probably has no bearing whatsoever on the issue of fwd decls vs. includes. I just want to avoid jumping to the wrong conclusions about the overall merits of IWYU.

(Also: I'm not intending to ignore your other points, but I wanted to clarify this first.)

t-wissmann:
> > The check that "enough files are included" is performed by the compiler. So I
> > don't see the reason at all why IWYU ever suggests to include a file that has
> > not been included yet in a compiling code base.
> 
> This is a crucial point. I thought so too at first, but I'm not sure if this is actually correct. I might totally be on the wrong track here, but here's what I'm thinking:
> 
> A compiling code base does _not_ imply that no include is missing. It just implies that all necessary declarations _happen to be available_ where needed. It is entirely possible that this is only the case because there are redundant includes (i.e., forgotten and wrong) that accidentally ensure a successful build through their transitive includes.
>
> In other words, a redundant include might _mask_ the fact an include is actually missing. And that might even be in some entirely different place, in case the redundant include is within a header file.

Right, but I don't know what is negative about the transitive includes. I mean if x.cpp needs vector but only includes y.h which includes vector, then I think this is fine. And if at some point, y.h gets rid of vector, then IWYU tells us to remove the vector-include from y.h and then x.cpp does not compile anymore. But this is all fine, because gcc tells us that we have to include vector in x.cpp now. Or am I missing something here?

I've read https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md and I think they say that my approach is only problematic for large code bases (such as google's), because it gets annoying if you have to add vector to many x.cpp-files.

> That's why I suspect you cannot separate the "include _only_ what you use" part from the "include _everything_ you use". You need to do the latter so that you can consistently do the former.
> 
> Of course, all this probably has no bearing whatsoever on the issue of fwd decls vs. includes. I just want to avoid jumping to the wrong conclusions about the overall merits of IWYU.
> 
> (Also: I'm not intending to ignore your other points, but I wanted to clarify this first.)

Okey, thanks for your quick comment! I also didn't want to ignore your PR even though I did not react for a week!

dnnr:
> But this is all fine, because gcc tells us that we have to include vector in x.cpp now. Or am I missing something here?

No, I don't think you're missing anything, but I'd definitely consider that scenario a bug (for lack of a better word).

If `x.cpp` needs vector, it should include vector, simple as that. From a practical perspective: If I'm working on `y.h` I don't want to be bothered with fixing what is a shortcoming of an unrelated module (`x.cpp`) that "my change" has nothing to do with.

To be honest, I did expect this to be controversial, but maybe it is. :man_shrugging:
Until now, I was simply assuming that there is a shared desire to literally *include exactly what we use*, and it was just matter getting the tool to agree with us on the details.

Compare it to how people treat dependencies in package management. I guess nobody would dispute that any, let's say, Debian package shall declare absolutely all of its dependencies (except maybe some kind of invariably available base set, dunno...). Of course, the impact of a mistake is much more grave than with includes, you probably won't notice immediately.

> Okey, thanks for your quick comment! I also didn't want to ignore your PR even though I did not react for a week!

No worries! I'm considering the ball to be in my court anyway, due to the fwd decls issue. I was planning on taking another look soon.

t-wissmann:
> > But this is all fine, because gcc tells us that we have to include vector in x.cpp now. Or am I missing something here?
> 
> No, I don't think you're missing anything, but I'd definitely consider that scenario a bug (for lack of a better word).
> 
> If `x.cpp` needs vector, it should include vector, simple as that. From a practical perspective: If I'm working on `y.h` I don't want to be bothered with fixing what is a shortcoming of an unrelated module (`x.cpp`) that "my change" has nothing to do with.
> 
> To be honest, I did expect this to be controversial, but maybe it is. 
> Until now, I was simply assuming that there is a shared desire to literally _include exactly what we use_, and it was just matter getting the tool to agree with us on the details.
> 
> Compare it to how people treat dependencies in package management. I guess nobody would dispute that any, let's say, Debian package shall declare absolutely all of its dependencies (except maybe some kind of invariably available base set, dunno...). Of course, the impact of a mistake is much more grave than with includes, you probably won't notice immediately.

I understand this argument and I also accept it for large code bases (such as google's). But the crucial difference between hlwm and debian is, that if a debian maintainer changes a package, then there is no CI check running within 10 Minutes checking that still every single debian package compiles. But with herbstluftwm, we have a single CI checking that herbstluftwm (i.e. all cpp files) still compiles. And furthermore, all maintainers are equally responsible for all cpp files of hlwm.

dnnr:
I'm aware of these differences, but to me, they only affect *how* severe a bug I consider such inconsistencies. And of course, they are not very severe at all. But that doesn't make them less wrong. We might simply disagree here, and that's okay.

Maybe there's one thing haven't properly conveyed: With this entire topic, I don't care so much about the quality of the code. After all, it's not like bad includes have any potential of ever turning our code into an unmaintainable tire fire. What I care about here is, let's call it "developer comfort":
1. not being annoyed with having to fix unrelated files because of incomplete includes
2. having strict assurances that I'm not going to merge mistakes that are going to turn into someone else's annoyance (see 1.) down the road

t-wissmann:
> I'm aware of these differences, but to me, they only affect _how_ severe a bug I consider such inconsistencies. And of course, they are not very severe at all. But that doesn't make them less wrong. We might simply disagree here, and that's okay.
> 
> Maybe there's one thing haven't properly conveyed: With this entire topic, I don't care so much about the quality of the code. After all, it's not like bad includes have any potential of ever turning our code into an unmaintainable tire fire. What I care about here is, let's call it "developer comfort":
> 
> 1. not being annoyed with having to fix unrelated files because of incomplete includes
> 2. having strict assurances that I'm not going to merge mistakes that are going to turn into someone else's annoyance (see 1.) down the road

I think I understand your motivation but I don't share your conclusion. The issue in 1. can be avoided easily by avoiding includes wherever possible. So whenever we can put a forward declaration instead of an include, we avoid that at some point in the future a developer is annoyed by issue 1.

dnnr:
> can be avoided easily by avoiding includes wherever possible

For real: How do you do that? This sounds like a legitimate superpower to me.

Using nothing by my naked eyes, I am entirely incapable of reliably spotting an unnecessary include. Not even in the context of a change that might make an existing include obsolete. I might have a chance *if* I consciously check for it *and* the source file is very small, but not in any real-world scenario.

Is this just me?

t-wissmann:
I've just deleted my drafted long answer because:

This discussion is ridiculous and off-topic. The change set of this PR regarding src/root.h is unacceptable for me, for the reasons listed in https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md

The fact that this change set has been written by a computer program instead of a human does not make it better.

PR #385, opened 2019-03-08
Remove status-success=WIP from .mergify.yml

dnnr:
Apparently, Mergify is fixated on the configuration file as it is in the
default branch, regardless of a PR's base branch. So without this,
Mergify will waits for status-sucess=WIP in PRs made against
winterbreeze.


dnnr:
Note that this PR is against master!

PR #388, opened 2019-03-20
Implement the load command

t-wissmann:
Implement the load command. Other than in the old implementation in
master, this implementation makes a clear distinction between parsing
and application. The new class FrameParser is only responsible for
parsing, and the resulting tree is then merged with the existing frame
tree in order to have the same behaviour as before.


t-wissmann:
This is still work in progress. The parsing is nearly finished, but I did not start with the application to the frame tree.

t-wissmann:
I think this branch is ready for being merged: `load` has the same features as in master, and it has sufficient test cases. There are still some places were the code is not beautiful (but there are only a few!)

t-wissmann:
Note that the CI only fails because `https://codecov.io/bash` is 503 at the moment.

tests/test_load.py
```diff
@@ -0,0 +1,133 @@
...
+    ('(split horizontal:0.5:0 (clients max:1', 38),
+])
+def test_syntax_errors_position(hlwm, invalid_layout, error_pos):
+    c = hlwm.call_xfail(['load', invalid_layout])
+    m = re.match(r'^load: Syntax error at ([0-9]*): ', c.stderr)
+    assert m is not None
+    assert int(m.group(1)) == error_pos
+
+
+example_partial_layouts = [
```
dnnr: In other test files, we've put global definitions like this at the top (and used upper-case for their names, as prescribed by PEP-8, I think). I'd suggest either doing that, or inlining it into its only point of use in the decorator below. Right now it seems kind of out of place.

The same applies to `example_full_layouts` further be low.

t-wissmann: Originally, I planned to use these in multiple test cases. But now I don't see the necessity anymore, so I've inlined them.

src/framedata.cpp
```diff
@@ -0,0 +1,52 @@
+#include "framedata.h"
+
+#include "completion.h"
+#include "layout.h"
+
+using std::string;
+
+size_t LayoutAlgorithmCount() {
```
dnnr: I'm seeing some room for improvement around this whole `g_layout_names` thing, but I'm guessing that you purposely left this untouched to minimize the diff for now (and tackle it later on).
Given that this is newly added code, maybe a TODO/FIXME indicating the temporary nature of it would be good. (That is, of course, only if I'm guessing correctly...)

Also: Why TitleCase for a function name?

src/framedata.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * The frame data header holds the classes that describe the members used for
+ * the actual logic of the frame tree tiling. So e.g. the 'selection' is a
+ * member of both of the following classes, whereas HSFrameLeaf::last_rect is
+ * not.
+ *
+ * Also note that the following classes only hold the plain members and does
+ * not set up class inheritance!
+ */
```
dnnr: I'm having a hard time understanding what this comment is trying to tell me. I'm aware that this is not a constructive review comment, but there's a circular dependency I'm struggling with...

Maybe there is more of an existing understanding expected of me (as in: the intended audience), so I'm not sure if my thoughts while reading this are helpful at all (and I'm not purposely trying to play dumb here):
 * What are "members used for the logic of the tiling"? Members as in functions?
 * Which "the 'selection'" is this about? Does that refer to the `selection_` member below?
 * What inheritance between which classe is *not* being set up here?

In other words, the crux might be that I have no idea what the question was that this is supposed to be the answer to.

src/framedata.cpp
t-wissmann: Yes I tried to minimize the diff, and the `g_layout_names` is still used in some command in `layout.cpp` which I did not want to touch.

The TitleCase is because this was originally part of the enum. But I think I can make it camelCase if you prefer.

src/framedata.h
t-wissmann: * I mean members as in 'member variables'. And the 'logic of the tiling' means: the following classes hold the member variables which are exposed to the user (via `dump`) and even explicitly in the man page in the section `TILING ALGORITHM`. In this section of the manpage, it is described what 'vertical' and 'horizontal' splittings mean, and thus this is part of the definition of the following classes. On the other hand, internal caching details like `HSFrameLeaf::last_rect` are just implementation details for caching certain things and this is thus in `HSFrameLeaf` and not in `FrameDataLeaf`
* The `selection` refers to both `FrameDataLeaf::selection` and to `FrameDataSplit<T>::selection_`.
* Note that HSFrameLeaf and HSFrameSplit inherit from a common base class HSFrame. But `FrameDataLeaf` and `FrameDataSplit<T>` do not have a common base class.

t-wissmann: The comment answers the question: why are some member variables part of `FrameDataSplit`/`FrameDataLeaf` and why are some other member variables part of `HSFrameSplit`/`HSFrameLeaf`.

t-wissmann: I've just clarified this comment.

dnnr: Thanks! I still feel like I'll have to dig a bit deeper some day to get the full picture, but at least this comment makes a lot more sense to me now. :+1:

src/framedata.cpp
t-wissmann: I've renamed it.

src/framedata.cpp
```diff
@@ -0,0 +1,52 @@
...
+    for (size_t i = 0; g_layout_names[i] != nullptr; i++) {
+        if (source == g_layout_names[i]) {
+            return (LayoutAlgorithm) i;
+        }
+    }
+    throw std::invalid_argument("Invalid split align name: \"" + source + "\"");
+}
+
+template<> string Converter<LayoutAlgorithm>::str(LayoutAlgorithm payload) {
+    return string(g_layout_names[(int)payload]);
```
dnnr: The `string(...)` should not be necessary here (also, a space after the cast would be nice).

The same applies to the `str` method

src/framedata.h
```diff
@@ -0,0 +1,61 @@
...
+protected:
+    std::vector<Client*> clients;
+    int selection = 0;
+    LayoutAlgorithm layout = LayoutAlgorithm::vertical;
+};
+
+template<typename BaseClass>
+class FrameDataSplit {
+protected:
+    SplitAlign align_ = SplitAlign::vertical;         // SplitAlign::vertical or SplitAlign::horizontal
```
dnnr: Why those spaces between the `;` and the `//`? Actually, the entire comment seems a bit superfluous, as it just describes the `SplitAlign` enum, doesn't it?

src/framedata.cpp
t-wissmann: I've fixed this (and updated this to the new Converter interface, cf PR #436)

src/framedata.h
```diff
@@ -0,0 +1,61 @@
...
+    std::shared_ptr<BaseClass> a_; // first child
+    std::shared_ptr<BaseClass> b_; // second child
+
+    int selection_ = 0;
+    int fraction_ = 0; // size of first child relative to whole size
+                  // FRACTION_UNIT means full size
+                  // FRACTION_UNIT/2 means 50%
```
dnnr: At first I suspected these comments were somehow misplaced, but this seems to be a multiline explanation for `fraction_`. Maybe it's best to turn this into a proper Doxygen docstring, some way like this:
```
int selection_ = 0;

/*!
 * Size of first child relative to whole size.
 * For example, a value of FRACTION_UNIT means full size
 * and FRACTION_UNIT/2 means 50%.
 */
int fraction_ = 0;
```

src/framedata.h
t-wissmann: Right, now that we have the enum, this comment is a bit redundant; so I've removed it!

src/framedata.h
t-wissmann: I've applied your suggestion.

t-wissmann:
IWYU says:

    /hlwm/src/frameparser.cpp should remove these lines:
    - #include <sstream>  // lines 5-5

    /hlwm/src/settings.cpp should remove these lines:
    - #include "layout.h"  // lines 9-9

However, these files don't compile without these includes. Am I missing something?

dnnr:
From what I can tell, settings.cpp is missing an include of framedata.h (because it's using layoutAlgorithmCount). With this include, it should be possible to remove the layout.h include.

For frameparser.cpp, I have to admit that I'm not really sure why, but removing iostream and keeping sstream seems to make both the compiler and IWYU happy. It doesn't have any "workaround smell", so I think we can just do it like this.

dnnr:
There is a "conflict" that didn't show up during the merge: `layoutAlgorithmCount` needs to be removed from `layout.cpp`. That should be the only reason for the build to fail.

src/layout.cpp
```diff
@@ -64,9 +57,12 @@ HSFrame::~HSFrame() = default;
 
 HSFrameLeaf::HSFrameLeaf(HSTag* tag, Settings* settings, weak_ptr<HSFrameSplit> parent)
     : HSFrame(tag, settings, parent)
-    , selection(0)
 {
-    layout = (LayoutAlgorithm) settings->default_frame_layout();
+    auto l = settings->default_frame_layout();
+    if (l >= layoutAlgorithmCount()) {
```
dnnr: Isn't this check already performed by the validator of `default_frame_layout`? Do we really need it here?

t-wissmann: In the sense of modularity: the cast from `int` to `LayoutAlgorithm` happens here but with only 'local knowledge', we don't know that the range of the int has already been checked. This check does not cost much, and it does not make the code ugly, so I'd vote for keeping it.

dnnr: Okay, agreed.

I suppose the "real" fix would be not using an `int` internally when we already know it to be a `LayoutAlgorithm`. Would it be possible/feasible to turn the attribute into a `LayoutAlgorithm` (not right now, but in general)?

t-wissmann: In principle yes. The current interface for the user is quite inconvenient, because the user has to enter a number instead of a layout name.

t-wissmann:
Any updates/opinions? Shall we just merge this? I think it's fine if you're not confident with the implementation but with the tests :-)

dnnr:
I was still in the (admittedly very slow) process of looking at all the *new* code, but I can understand if you're getting impatient. I haven't looked into the tests yet, but I will focus on those now, and that shouldn't take too long anymore, I think.

That said, if you really want to merge it immediately, go ahead. My gut feeling is that there aren't any major issues in the parts I haven't looked at yet.

tests/test_load.py
```diff
@@ -0,0 +1,131 @@
...
+    ('(clients grid:0 x)', 16),
+    ('(split horizontal:0.5:0 x)', 24),
+    ('(split horizontal:0.5:0 (split horizontal:0.5:1', 47),
+    ('(split horizontal:0.5:0 (split horizontal:0.5:1 ', 48),
+    ('(split horizontal:0.5:0 (split horizontal:0.5:1 )', 49),
+    ('(split horizontal:0.5:0 (split horizontal:0.5:1 )))', 50),
+    ('(split horizontal:0.5:0 (clients max:1', 38),
+])
+def test_syntax_errors_position(hlwm, invalid_layout, error_pos):
+    c = hlwm.call_xfail(['load', invalid_layout])
```
dnnr: Wouldn't it be possible (and simpler) to put this in place of the re.match and the two asserts?
 ```
c.expect_stderr(f'^load: Syntax error at {error_pos}: ')`

tests/test_load.py
```diff
@@ -0,0 +1,131 @@
...
+    m = re.match(r'^load: Syntax error at ([0-9]*): ', c.stderr)
+    assert m is not None
+    assert int(m.group(1)) == error_pos
+
+
+def is_subseq(x, y):
+    # from https://stackoverflow.com/a/24017747/4400896
```
dnnr: The SO link might not work forever, so maybe add proper docstring like
```
""" Checks if x is a subsequence (not substring) of y. """
```

tests/test_load.py
t-wissmann: Right! I've fixed it, thanks!

tests/test_load.py
t-wissmann: Right! I've added the docstring.

tests/test_load.py
```diff
@@ -0,0 +1,130 @@
...
+    "(split vertical:0.3:0 (split vertical:0.4:1))",
+])
+@pytest.mark.parametrize('num_splits_before', [0, 1, 2])
+def test_valid_layout_syntax_partial_layouts(hlwm, layout, num_splits_before):
+    for i in range(0, num_splits_before):
+        hlwm.call('split explode')
+
+    hlwm.call(['load', layout])
+
+    assert is_subseq(layout.replace(' ', ''), hlwm.call('dump').stdout)
```
dnnr: Oof. I'm having a hard time understanding this.

Are you using is_subseq here only because the difference between a load and a subsequent dump is that the dump contains the `(clients ...)` components?

tests/test_load.py
```diff
@@ -0,0 +1,130 @@
...
+])
+def test_only_warn_invalid_winids(hlwm, layout):
+    p = hlwm.call(['load', layout])
+
+    assert p.stdout.startswith("Warning: Unknown window IDs")
+
+
+@pytest.mark.parametrize(
+    "running_clients_num,focus",
+    [(n, f) for n in [1, 3] for f in range(0, n)])
```
dnnr: FYI: `range(0, x)` can also be written as `range(x)`. But in case you put the `0` there on purpose, to make things more explicit, keep it! :+1:

tests/test_load.py
```diff
@@ -0,0 +1,130 @@
...
+        "(clients max:0 W)",
+        "(clients max:1 W W)",
+        "(split horizontal:0.9:0 (split vertical:0.5:1 (clients max:0) (clients grid:0)) (clients horizontal:0))",
+        "(split vertical:0.4:1 (clients max:2 W W W) (clients grid:0 W))",
+    ])
+def test_full_layouts(hlwm, layout):
+    clients = [hlwm.create_client() for k in range(0, layout.count('W'))]
+    for winid, _ in clients:
+        # replace the next W by the window ID
+        layout = layout.replace('W', winid, 1)
```
dnnr: Nice solution! :100:

tests/test_load.py
```diff
@@ -0,0 +1,130 @@
...
+    assert p.stdout == ''
+    assert layout == hlwm.call('dump').stdout
+
+
+@pytest.mark.parametrize("layout", [
+    "(clients horizontal:0 0234)",
+    "(clients vertical:0 0x2343)",
+    "(clients vertical:0 1713)",
+])
+def test_only_warn_invalid_winids(hlwm, layout):
```
dnnr: What's the "only" referring to here? Should the be no other warning (then I'd be missing an assertion for that)? Or is it meant as "*just* test the warning for invalid winids"? Then I'd simply call it `test_load_invalid_winids`

tests/test_load.py
t-wissmann: Yes. The output of contains the full tree but in `layout` we didn't specify the tree in its entirety.

tests/test_load.py
t-wissmann: I didn't know this! But I think here it makes reading easier because `n` comes from the other for loop.

tests/test_load.py
t-wissmann: Yes, it means 'only warn' but still return 0. I've renamed it!

tests/test_load.py
t-wissmann: I've added a comment to make this clearer in the test case. You're right that the test case as such is very hard to read.

dnnr: How would you feel about dealing with this by running the Python-equivalent of `s/\(clients [^)]+\)//g` on the `dump` output and then doing a regular string comparison? Would that work?

tests/test_load.py
dnnr: Aah, okay! I didn't even think of that meaning :smile:

tests/test_load.py
t-wissmann: No this would not work for e.g. `layout =  "(split vertical:0.3:0 (clients horizontal:0))"`. And if I did the replacement also in `layout` then I would test too little. Furthermore, the additional branches can also be `(split …)`-branches (this is the case for `layout = "(split horizontal:0.3:0)"` and `num_splits = 2`).

dnnr: Okay, got it (I think)!

t-wissmann: Sorry, for the lack of descriptive comments in the test case.

PR #390, opened 2019-04-03
Get rid of glib dependency

t-wissmann:
This rewrites the few remaining code parts that used glib. This
affects the xinerama code and hook_emit(), which now use vectors.


tests/test_monitor.py
```diff
@@ -100,3 +100,8 @@ def test_new_clients_appear_in_focused_monitor(hlwm):
     assert hlwm.get_attr('tags.by-name.tag2.client_count') == '1'
     assert hlwm.get_attr('tags.by-name.default.client_count') == '0'
     # TODO: Assert that client's winid is in tag2 (not yet possible)
+
+def test_detect_monitors_does_not_crash(hlwm):
```
dnnr: Seems reasonable to "test" `detect_monitors` this way, but there should be two blank lines above. ;-)

dnnr:
This is quite a huge change that looks like it could have been a couple of separate PRs. I didn't spot anything problematic, but might well have missed something.

t-wissmann:
Do you prefer to have it as separate PRs? I'm also fine with creating a merge commit (the commit history in this PR should be quite clean).

dnnr:
I wouldn't "demand" it, but I personally would have gone for smaller PRs. You'll probably get more thorough reviews that way.

That said, the whole thing doesn't look too bad, so feel free to go ahead with this PR if you don't see a particular need for a closer look here.

My main trouble with it is figuring out the relation between the implementation changes and the tests. With smaller steps, it's often much easier to tell whether there is anything not yet tested or if the existing tests are really sufficient.

t-wissmann:
I'm still used to the meanings 'branch' = 'feature' and 'commit' = 'one step towards a feature' and 'branch' = 'PR'. I can try to make smaller PRs in the future.

The-Compiler:
FWIW might be easier to review this PR commit by commit?

t-wissmann:
> FWIW might be easier to review this PR commit by commit?

Right, and reviewing commit-wise gives you meaningful commit messages.

dnnr:
In theory, yes, you could review it commit-by-commit (and if your branch has a clean history, you'd probably turn every commit into a PR so we'd end up doing exactly that anyway). But the tooling doesn't really support this. You can only comment a specific overall change, you cannot see if all intermediate commits would have a green CI build, you'd have to rebase all subsequent commits to incorporate a change suggestion somewhere in the history.

Fun fact: In Gerrit, you pretty much review each commit, but that's because the very foundation of Gerrit is to use commits as "unit to be reviewed" instead of branches.

The-Compiler:
> But the tooling doesn't really support this.

It doesn't really support multiple stacked PRs very well either (unless you're willing to wait until every single-commit PR is merged and continue based on that - and then you end up with some half-baked feature in the master branch).

> You can only comment a specific overall change

You can comment on individual commits of a PR just fine.

>  you cannot see if all intermediate commits would have a green CI build

I'd argue you usually don't care - I'd even argue that it's okay for individual commits in a PR to not work properly (like if you're breaking up a big refactoring or something, into logical human-readable units rather than "it compiles" units).

But if you do, the commit status is shown for every push in the commit list.

src/command.cpp
```diff
@@ -689,18 +688,13 @@ int complete_command(int argc, char** argv, Output output) {
...
 void complete_against_env(int argc, char** argv, int position,
                           Output output) {
-    GString* curname = g_string_sized_new(30);
     const char* needle = (position < argc) ? argv[position] : "";
     for (char** env = environ; *env; ++env) {
-        g_string_assign(curname, *env);
-        char* name_end = strchr(*env, '=');
-        if (!name_end) {
-            continue;
+        vector<string> chunks = ArgList::split(*env, '=');
```
dnnr: This is just a test comment to see how GH displays it.

dnnr:
> > But the tooling doesn't really support this.
> 
> It doesn't really support multiple stacked PRs very well either

True, but that's only an issue if the changes really depend on each other (or touch the same lines). That's probably not the case when doing something like "replace GLib code in five places".

> > You can only comment a specific overall change
> 
> You can comment on individual commits of a PR just fine.

Indeed! I think this changed recently. Before, when commenting on a single commit in a PR, that comment was not visible in a PR because it was literally a comment in the commit.

t-wissmann:
I've moved the bigger change sets to PRs #392 and #393 and so this PR won't compile until the other PRs are merged.

cmake/dependencies.cmake
```diff
@@ -1,8 +1,6 @@
 ## get project dependencies
 # Xlib
 find_package(X11 REQUIRED)
-# GLib (will be removed later)
-include(FindPkgConfig)
-pkg_check_modules(GLIB2 REQUIRED glib-2.0)
+#include(FindPkgConfig)
```
dnnr: Does this stay for any particular reason?

t-wissmann: all this cmake configuration is not clear to me, so I kept it because it might help at some point if we want to add new dependencies...

But actually you're right if you vote for dropping it.

t-wissmann: I've removed this line.

dnnr:
I guess the PR description could be updated now

t-wissmann:
I've updated the description, thanks!

PR #391, opened 2019-04-03
Explicitly use bash in tests for read -n

t-wissmann:
We use 'read -n 1' in the test cases, which is a bash specific option.
To let the test cases run if bash is not the default shell, explicitly
use bash.

### DIFF
--- a/tests/test_stack.py
+++ b/tests/test_stack.py
@@ -63,19 +63,19 @@ def test_stack_tree(hlwm):
   - 
     - Monitor 1 ("monitor2") with tag "tag2"
       - Focus-Layer
-        - Client <windowid> "sleep infinity"
+        - Client <windowid> "bash"
       - Fullscreen-Layer
       - Normal Layer
-        - Client <windowid> "sleep infinity"
+        - Client <windowid> "bash"
       - Frame Layer
         - Window <windowid>
         - Window <windowid>
     - Monitor 0 with tag "default"
       - Focus-Layer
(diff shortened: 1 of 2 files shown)


tests/test_stack.py
```diff
@@ -63,19 +63,19 @@ def test_stack_tree(hlwm):
   - 
     - Monitor 1 ("monitor2") with tag "tag2"
       - Focus-Layer
-        - Client <windowid> "sleep infinity"
+        - Client <windowid> "bash"
```
dnnr: I get the purpose of the PR, but what's the reason for this part?

t-wissmann: The change in the hlwm bridge affects the window title, which again affects this test case.

dnnr: Ah, I read this wrong. Yeah, makes perfect sense! :+1:

dnnr:
Whoopsie, already merged. Feels weird being on the other end of this automation.

t-wissmann:
Do you think we should change the rules for mergify?

dnnr:
Well, right now the process is basically *"if you're yet ready to have it auto-merged, put a WIP in the title"*. We could probably to change that to *"as soon as you want it auto-merged, add some label to the PR"* and that way turn it from opt-out to opt-in. I see benefits in both ways and don't mind the status quo enough to push for a switch myself. In this case, I was simply a bit surprised because I never did it before.

PR #392, opened 2019-04-04
Drop glib from the disjoin_rects algorithm

t-wissmann:
This drops the glib dependency in rectangle.cpp. The new algorithm is
much different than the original code, but makes the whole algorithm
simpler, shorter, and hopefully much easier to read.

### DIFF
--- a/src/rectangle.cpp
+++ b/src/rectangle.cpp
@@ -43,7 +43,7 @@ static RectangleVec disjoin_from_subset(Rectangle large, Rectangle center)
     // +-----------------------+
     // coordinates of the bottom right corner of large
     int br_x = large.x + large.width, br_y = large.y + large.height;
-    Rectangle (*r)(int,int,int,int) = Rectangle::fromCorners;
+    auto r = Rectangle::fromCorners;
     Rectangle top   = r(large.x, large.y, large.x + large.width, center.y);
     Rectangle left  = r(large.x, center.y, center.x, center.y + center.height);
     Rectangle right = r(center.x + center.width, center.y, br_x, center.y + center.height);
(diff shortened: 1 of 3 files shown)


src/rectangle.cpp
```diff
@@ -79,63 +41,64 @@ static RectList* insert_rect_border(RectList* head,
...
-    } else {
-        // element intersects with the head rect
-        auto center = intersection_area(head->rect, element->rect);
-        auto large = head->rect;
-        head->rect = center;
-        head->next = insert_rect_border(head->next, large, center);
-        head->next = insert_rect_border(head->next, element->rect, center);
-        g_free(element);
-        return head;
+    Rectangle (*r)(int,int,int,int) = Rectangle::fromCorners;
```
dnnr: Unless you prefer it like it this, I *think* you can just say `auto r = ...` here.

dnnr:
Not an issue with this PR but it caught my eye here: How would you feel about removing those opaque `RectangleVec` and `RectangleVecIdx` typedefs in favor of just writing it as it is? The latter could perhaps also be a `map<int, Rectangle>` but I haven't checked if that fits the way it is used.

t-wissmann:
I can't find `RectangleVecIdx`, where is it defined?

`RectangleVec` is used very rarely, so I think we can replace it with `std::vector<Rectangle>`.

dnnr:
`RectangleVecIdx` is defined in `x11-types.h`, right next to `RectangleVec`.

t-wissmann:
For `RectangleIdxVec`, I suggest that we should find a solution that is local in floating.h, because this is the only place where it is used. So e.g. something like
```
class Floating {
   using RectangleIdxVec = std::vector<std::pair<int, Rectangle>>;
   …
   static find_in_direction(...);
}
```

dnnr:
That would certainly be an improvement. :+1:

PR #393, opened 2019-04-04
Drop glib from ewmh and stack

t-wissmann:


src/ewmh.cpp
```diff
@@ -172,69 +170,36 @@ void ewmh_get_original_client_list(Window** buf, unsigned long *count) {
...
 }
 
-struct ewmhstack {
-    Window* buf;
-    int     count;
-    int     i;  // index of the next free element in buf
-};
+void ewmh_update_client_list_stacking() {
+    // First: get the windows currently visible
+    auto buf = make_shared<vector<Window>>();
```
dnnr: I'm having a hard time grasping the data flow here (might just be my own shortcoming, not an issue with the PR).

So, the monitor stack and the tag stacks get passed a shared pointer to this `buf`, and use it to push Windows into the vector, right? There are no references (as in: copies of the shared pointer) kept within those `toWindowBuf()` calls and stored beyond the scope of those calls? Ultimately, the vector is used for that `XChangeProperty` call (I'm guessing to tell X about the stacking order) and that's the end of its lifecycle, right?

t-wissmann: Yes, you're right! The `XChangeProperty` only provides the stacking order according to the EWMH protocol. The actual stacking order is changed via `XRestackWindows` in `Monitor::restack()`, involving the same data flow.

Originally, I hoped that there was some `outstream<Window>` datastructure. If you have a better solution than `shared_ptr<vector<Window>>` I'm happy to hear it!

Another option would be to pass some parameter `function<void(Window)> push_on_stack`, and so in `toWindowBuf()` we can call e.g. `push_on_stack(client->window)` without the need to see how this stack is implemented. What do you think?

dnnr: Okay, I see. It's tricky. First of all, I don't think a `shared_ptr` is appropriate here. Technically, it would be fine to just pass along a reference, but I can see how that doesn't feel right, because it looks a lot like `toWindowBuf` treats `buf` like it had ownership, which it doesn't.

That makes think that the root issue is with the interface offered by Stack, as you indicated yourself.

Maybe I'm missing something, but why shouldn't `toWindowBuf` just return a vector object, leaving it to the caller to concatenate the vectors together? It's not really efficient, but still within the same complexity class, so we usually don't care, right?

t-wissmann: Yes, we can make it return a vector. It's just a bit less convenient, because we need a bit more code in `toWindowBuf` (create it as a local variable and concatenate vectors).

I guess this makes the code clearer than the `push_on_stack` option, because it's a much more usual pattern.

t-wissmann: I've implemented returning of a Vector, and I don't think it is as beautiful as the `push_on_stack`, so at some point, I will make a separate PR that shows the difference.

src/stack.cpp
```diff
@@ -162,115 +162,59 @@ int print_stack_command(int argc, char** argv, Output output) {
...
-    data.missing = 0;
-    data.real_clients = real_clients;
-
+vector<Window> Stack::toWindowBuf(bool real_clients) {
+    vector<Window> result;
     for (int i = 0; i < LAYER_COUNT; i++) {
-        data.layer = (HSLayer)i;
         for (auto slice : top[i]) {
-            slice_to_window_buf(slice, &data);
+            vector_append(
```
dnnr: Maybe it's just a matter of personal preference, but I think I would have gone with just another `for` loop (calling `push_back`) instead of introducing a helper like that.

t-wissmann: I need this `vector_append` at multiple places (4 times). I don't think that it makes the code better if I replace the 4 calls of `vector_append` with a for loop each.

src/ewmh.cpp
dnnr: I wasn't sure if you wanted to do this *first*. If you want to go ahead for now and merge it like it is: Everything looks good to me!

t-wissmann: My suggestion is in PR #465.

PR #394, opened 2019-04-07
Fix most flake8 warnings in Python code

blarz:
Some unused imports, but mostly whitespace. Ignoring the "line too long" warning.


t-wissmann:
Great! Can you also add a script (in a separate PR) that runs these checks such that I can directly check it when writing further test cases?

tests/conftest.py
```diff
@@ -243,9 +243,9 @@ def __init__(self, tmpdir, env):
         autostart.chmod(0o755)
         bin_path = os.path.join(BINDIR, 'herbstluftwm')
         self.proc = subprocess.Popen([bin_path, '--verbose'], env=env,
-                bufsize=0,  # essential for reading output with selectors!
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE)
+                                     bufsize=0,  # essential for reading output with selectors!
+                                     stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE)
```
t-wissmann: Here, the indentation was "wrong" because I found the line length more important than the alignment of the parameters below each other. How is this handled in python generally? Is the following a solution solving both "violations"?

    self.proc = subprocess.Popen(
        [bin_path, '--verbose'], env=env,
        bufsize=0,  # essential for reading output with selectors!
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE)

blarz:
> Great! Can you also add a script (in a separate PR) that runs these checks such that I can directly check it when writing further test cases?

I will talk to @dnnr about this. I just used the flake8 plugin in my editor for now.
But adding a general flake8 config file and maybe add the check to CI job would certainly be nice!

tests/conftest.py
blarz: True, your suggestion is a valid solution for that check.
I updated the PR

blarz:
I missed a few more, I'll fix them tomorrow

t-wissmann:
Ok, just tell when you're finished!

tests/conftest.py
The-Compiler: According to [pep8](https://www.python.org/dev/peps/pep-0008/#indentation) (which most projects follow), these two possibilities are fine:

- Aligning with `(` (with content on the line containing the `(`)
- Hanging indent (indented by four spaces, without content on the line containing the `(`)

So yup, your proposal is fine.

The-Compiler:
> But adding a general flake8 config file and maybe add the check to CI job would certainly be nice!

FWIW I'd suggest setting up [tox](https://tox.readthedocs.io/en/latest/) which many projects use for this purpose - it allows doing e.g. `tox -e flake8`, sets up a virtualenv with all specified dependencies (flake8 itself, maybe plugins in the future), and runs the specified commands. Then Travis can also just be configured to run `tox -e flake8` as well.

ci-build.py
```diff
@@ -46,16 +45,14 @@
 build_env = os.environ.copy()
 build_env.update({
     'CC': args.cc,
-    'CXX': args.cxx,
-    })
+    'CXX': args.cxx})
```
dnnr: Is this really how we want it to be? I usually try to format all lines in a listing like this the same (hence the trailing comma even on the last entry), and I assumed this to be the generally preferred way™.

Did flake8 really suggest this change? I can't reproduce it in my environment.

ci-build.py
```diff
@@ -73,9 +70,7 @@
 
 if args.run_tests:
     tox_env = os.environ.copy()
-    tox_env.update({
-        'PWD': build_dir,
-        })
+    tox_env.update({'PWD': build_dir})
```
dnnr: Maybe just put `tox_env['PWD'] = build_id` here?

tests/conftest.py
```diff
@@ -222,7 +222,7 @@ def shutdown(self):
 @pytest.fixture
 def hlwm(hlwm_process):
     display = os.environ['DISPLAY']
-    #display = ':13'
+    # display = ':13'
```
dnnr: Does this serve any purpose, @t-wissmann? Maybe just remove it...

dnnr: (same goes for the similar line further below)

ci-build.py
blarz: That was one question I asked further down in this PR.
The other solution is having the closing bracket on a separate line indented with the beginning of the line opening the bracket, like @The-Compiler suggested.
I will adjust this tonight.

blarz: Please take a look at the way I did it now and tell me if that looks better

ci-build.py
blarz: Fixed

blarz:
The only three things still failing in flake8:

./www/compose.py:35:33: F821 undefined name 'basestring'
Not sure why it complains this is Python built in...

./tests/test_rules.py:184:9: W503 line break before binary operator
If I move the binary operator before the line break, it complains about this, too... Strange. Ignore it?

./tests/test_stack.py:65:4: W291 trailing whitespace
This is a deliberate trailing whitespace. But I cannot "noqa" it in line...

ci-build.py
dnnr: Yup, much better. I don't have a strong opinion on how to indent the closing brackets, as long as they're on a separate line. To my surprise, flake8 doesn't seem to care either.

My slight preference would be to indent them just like the lines before (i.e., not having them on the same level as the opening line). But I really don't mind either way.

blarz:
> The only three things still failing in flake8:
> 
> ./www/compose.py:35:33: F821 undefined name 'basestring'
> Not sure why it complains this is Python built in...
> 
> ./tests/test_rules.py:184:9: W503 line break before binary operator
> If I move the binary operator before the line break, it complains about this, too... Strange. Ignore it?
> 
> ./tests/test_stack.py:65:4: W291 trailing whitespace
> This is a deliberate trailing whitespace. But I cannot "noqa" it in line...

flake8 is quiet now, only complaining about lines being too long (which we will ignore in the tox.ini)

tests/test_stack.py
```diff
@@ -80,5 +80,5 @@ def test_stack_tree(hlwm):
         - Client <windowid> "bash"
       - Frame Layer
         - Window <windowid>
-'''
+'''  # noqa: W291 ignore trailing whitespace in this string
```
dnnr: The suppression would have been reasonable, but I took the opportunity to fix the underlying reason for the trailing space. As soon as #396 is merged, you can drop this noqa.

www/compose.py
```diff
@@ -2,12 +2,12 @@
 
 import sys
 import datetime
-import types
+from __builtin__ import basestring
```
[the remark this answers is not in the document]
blarz: I also messed up the print parentheses

tests/conftest.py
t-wissmann: I sometimes activate this lines to let the tests run in a `xephyr` (instead of a `xvfb`).

www/compose.py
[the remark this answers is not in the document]
t-wissmann: I think I've used python 2 because asciidoc / a2x use python 2, but fortunately, we don't import any asciidoc module, so it's very appropriate to move forward to python 3!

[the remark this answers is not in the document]
blarz: Now that #398 is merged, I'll update this PR. Thanks!

tests/conftest.py
blarz: I'll keep it then

tests/test_stack.py
blarz: Removed it

blarz:
I think this is good to go now

tests/conftest.py
```diff
@@ -123,7 +124,7 @@ def create_client(self, term_command='sleep infinity', title=None):
         command = ['xterm'] + title + ['-class', wmclass, '-e', 'bash', '-c', term_command]
 
         # enforce a hook when the window appears
-        self.call(['rule', 'once', 'class='+wmclass, 'hook=here_is_'+wmclass])
+        self.call(['rule', 'once', 'class=' + wmclass, 'hook=here_is_' + wmclass])
```
The-Compiler: Since we're already using fstrings anyways, this could be:

```suggestion
        self.call(['rule', 'once', f'class={wmclass}', f'hook=here_is_{wmclass}'])
```

instead.

tests/conftest.py
The-Compiler: FWIW it's quite straightforward to [add custom commandline options](https://docs.pytest.org/en/latest/example/simple.html) to pytest for your testsuite (basically, add a `pytest_addoption` hook adding the option, and here use the `config` fixture and check `config.getoption('--xephyr')` or so).

tests/conftest.py
```diff
@@ -343,7 +346,7 @@ def investigate_timeout(self, reason):
                             + " but hlwm still running") from None
         else:
             raise Exception("{} made herbstluftwm quit with exit code {}"
```
The-Compiler: With fstrings:

```suggestion
            raise Exception(f'{reason} made herbstluftwm quit with exit code {self.proc.returncode}' from None
```

tests/conftest.py
t-wissmann: I had the impression that one should avoid fstrings. Here, I find the non-fstring version more readable (also in the other changeset), so I'd keep it as it is.

PR #395, opened 2019-04-08
Enable flake8 in CI build

blarz:
Ignoring overlong lines and a PEP8 incompatible warning


blarz:
Needs #394 first

ci-build.py
```diff
@@ -77,6 +77,7 @@
         'PWD': build_dir,
         })
     sp.check_call(f'tox -e py37 -- -n auto -v -x', shell=True, cwd=build_dir, env=tox_env)
+    sp.check_call(f'tox -e flake8', shell=True, cwd=build_dir, env=tox_env)
```
dnnr: I'm not sure if flake8 should really be part of `--run-tests`... that doesn't make a lot of sense.

We should either rename the CLI argument (but I wouldn't know a good name) or introduce a `--flake8` argument.
I'd be fine with the latter, but this won't scale well *if* we are going to add many more things like that.

Alternatively, we could offer something like `--checks=iwyu,tests,flake8`. Unfortunately, argparse doesn't seem to offer that out of the box.

blarz: I added an extra flag for flake8 and extracted the whole tox environment preparation and execution

t-wissmann:
To be honest, I already feared that when I write 'document flake8 requirements in terms of a script' that this is read as 'check it automatically in the CI'. There is something I really dislike about our current ci setup, namely:

_If there is some non-entirely-standard whitespace placement in the test-system's python script, then not a single testcase is run on the actual herbstluftwm system_

I'm aware that this PR did not create this issue, but it's contributing to it and we should really get rid of this by moving all the code-style checks (whitespace in py, order of includes, usage of `using std::…` to a spearate build job.

tox.ini
```diff
@@ -21,3 +21,14 @@ deps =
...
+
+[testenv:flake8]
+deps =
+    flake8
+commands = flake8 .
+
+[flake8]
+ignore =
+    E501, # line too long
+    W503, # incompatible with pep8.. "Line break occurred before a binary operator"
```
The-Compiler: A bit of context: For years, pep8 required line breaks to be after binary operators, then [it changed](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) to instead require the opposite. Thus, flake8 (or rather pycodestyle, which flake8 uses) has both checks, but [disables](http://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore) W503 by default.

Setting a custom `ignore = E501` overrides the default ignores, which is why this needs to be re-added.

dnnr:
@t-wissmann You're raising a very valid issue. In automated quality gates, this is an often seen trade-off between two opposing goals:
 1. wanting checks to fail as early as possible (to reduce wait time and resource consumption)
 2. wanting to provide a full picture of everything problematic found in a change

Occasionally, the tooling offers something like "mark build as failed already, but continue running the remaining tests", but usually, you have to make a choice one way or the other.

I don't see any particular reason for us to stick to variant 1 (I just did it like that because it's a bit easier), so if I understand your concerns correctly, we should be able to simply resolve them like this: change ci-build.py to continue executing even in the case of failure, but still emit a non-zero exit status. I'll have to figure out something else for the "ci-build.py && ci-build.py" chain .travis.yml, but that won't be a blocker.

Would that be acceptable?

t-wissmann:
@dnnr: it's still not so convenient because one needs to scroll through the build log of "check nearly everytihng". What about my proposed solution (see #399) to move all the 'coding style' checks to a separate build job?

tox.ini
blarz: Thanks for the explanation!

PR #397, opened 2019-04-08
Add CMakeLists.txt.user to gitignore

blarz:


.gitignore
```diff
@@ -9,3 +9,6 @@ build*/
 /.cache
 /.tox
 __pycache__
+
+# ignore CMake generated files
```
dnnr: CMake is innocent here. This is Qt Creator's doing.

blarz: True. Updated the comment, thanks!

Issue #399, opened 2019-04-08
Move coding style CI checks to separate build job


t-wissmann:
cf. #395:

There is something I really dislike about our current ci setup, namely:

_If there is some non-entirely-standard whitespace placement in the test-system's python script, then not a single testcase is run on the actual herbstluftwm system_

We should really get rid of this by moving all the code-style checks (whitespace in py, order of includes, usage of using std::… to a spearate build job.

When creating a new PR, in a new branch, it's simply frustrating when the CI did not run any test cases because the includes weren't ordered alphabetically...

blarz:
This issue can be closed, right?

t-wissmann:
Right! Thanks for the reminder.

PR #400, opened 2019-04-09
Extract separate CI job for linting/static analysis

dnnr:


dnnr:
This ended up a bit more verbose in `.travis.yml` than I'd like it to be. I'm open to suggestions.

In any case, I think it's still an acceptable fix for #399.

t-wissmann:
Great, thanks a lot! This really makes my life easier when preparing new PRs!

I don't know how to reduce the redundancy from the .travis.yml. I also didn't know that IWYU requires compiled sources. In gitlab there is this notion of different stages that can pass intermediate results to each other. But I fear in travis we have to re-compile it, right?

dnnr:
> I also didn't know that IWYU requires compiled sources.

Oh, it doesn't. It just needs the generated build directory, hence the new `--compile` flag which makes the actual compilation step optional.

The significant waste/overhead (as in: duplicate operations) introduced by this PR is happening *outside* of `ci-build.py`: Booting the VM and doing the Docker stuff. But all things considered, I find this acceptable.

dnnr:
...and no, to my knowledge, there is no feasible way of passing intermediate results between jobs in Travis.

t-wissmann:
Ok Great! I think booting the VM is very acceptable!

PR #406, opened 2019-04-13
Add tests for urgent windows

t-wissmann:
This uses python-xlib in the tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -435,7 +435,7 @@ def x11():
 
             w.map()
             self.display.sync()
-            return w
+            return w, self.winid_str(w)
 
         def shutdown(self):
             try:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -379,7 +379,6 @@ class HlwmProcess:
                 raise Exception("herbstluftwm did not quit on sigterm"
                                 + " and had to be killed") from None
 
-
 @pytest.fixture(autouse=True)
 def hlwm_process(tmpdir):
     env = {
@@ -402,6 +401,33 @@ def running_clients(hlwm, running_clients_num):
     """
     return hlwm.create_clients(running_clients_num)
 
+@pytest.fixture()
+def x11():
+    from Xlib import X, display, Xutil
+    class X11:
+        def __init__(self):
+            self.display = display.Display()
+            self.screen = self.display.screen()
+            self.root = self.screen.root
+
+        def window(self, winid_string):
+            """return python-xlib window wrapper for a string window id"""
+            winid_int = int(winid_string, 0)
+            return self.display.create_resource_object('window', winid_int)
+
+        def make_window_urgent(self, winid_string):
+            """make window urgent, given a winid (as a string)"""
+            w = self.window(winid_string)
+            w.set_wm_hints( flags = Xutil.UrgencyHint )
+            self.display.sync()
+
+        def is_window_urgent(self, winid_string):
+            """check urgency of a given window id (as a string)"""
+            w = self.window(winid_string)
+            hints = w.get_wm_hints()
+            return bool(hints.flags & Xutil.UrgencyHint)
+
+    return X11()
 
 @pytest.fixture()
 def keyboard():
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -211,16 +211,6 @@ class HlwmBridge:
                 'waiting for hook triggered by client \"{}\"'.format(wmclass))
         return line[-1]
 
-    @contextmanager
-    def wait_for_hook(self, *hook_regex_list):
-        """context manager for herbstclient -w HOOK_REGEX_LIST"""
-        hc_wait = subprocess.Popen([self.HC_PATH, '-w'] + list(hook_regex_list))
-        # TODO: fix the race condition that the freshly forked hc_wait connects
-        # to hlwm too late
-        yield
-        hc_wait.wait(2)
-        assert hc_wait.returncode == 0
-
     def shutdown(self):
         for client_proc in self.client_procs:
             client_proc.terminate()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -434,6 +434,7 @@ def x11():
                 w.set_wm_hints( flags = Xutil.UrgencyHint )
 
             w.map()
+            self.display.sync()
             return w
 
         def shutdown(self):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -405,19 +405,48 @@ def x11():
             winid_int = int(winid_string, 0)
             return self.display.create_resource_object('window', winid_int)
 
-        def make_window_urgent(self, winid_string):
-            """make window urgent, given a winid (as a string)"""
-            w = self.window(winid_string)
-            w.set_wm_hints( flags = Xutil.UrgencyHint )
+        def winid_str(self, window_handle):
+            return hex(window_handle.id)
+
+        def make_window_urgent(self, window):
+            """make window urgent"""
+            window.set_wm_hints( flags = Xutil.UrgencyHint )
             self.display.sync()
 
-        def is_window_urgent(self, winid_string):
-            """check urgency of a given window id (as a string)"""
-            w = self.window(winid_string)
-            hints = w.get_wm_hints()
+        def is_window_urgent(self, window):
+            """check urgency of a given window handle"""
+            hints = window.get_wm_hints()
+            if hints is None:
+                return False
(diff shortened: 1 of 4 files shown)


tests/test_clients.py
```diff
@@ -35,3 +36,53 @@ def test_close_without_clients(hlwm):
...
     hlwm.create_clients(running_clients_num)
     hlwm.call('close')
+
+
+@pytest.mark.filterwarnings("ignore:tostring")
+@pytest.mark.parametrize("urgent", [True, False])
+def test_urgent_on_start(hlwm, x11, urgent):
+    command = []
+    if urgent:
+        # prepent urgent command
```
dnnr: Typo ("prepend")

tests/test_clients.py
```diff
@@ -35,3 +36,53 @@ def test_close_without_clients(hlwm):
...
+        command.append(r"echo -e '\a'")
+    command += [
```
dnnr: Why not use `command.append()` here, too?

dnnr: Ah you're still working on this anyway. My bad, nevermind.

tests/test_clients.py
t-wissmann: Luckily, I got rid of this part by using python-xlib directly.

tests/test_clients.py
t-wissmann: I expected this list to become longer... Also, I hoped that the race condition is not triggered in travis. I was wrong, and so I've replaced this part.

tests/conftest.py
```diff
@@ -365,7 +369,6 @@ def shutdown(self):
                 raise Exception("herbstluftwm did not quit on sigterm"
                                 + " and had to be killed") from None
 
-
```
dnnr: This was a load-bearing blank line, please keep intact :stuck_out_tongue:

tests/test_clients.py
```diff
@@ -1,4 +1,5 @@
 import pytest
+import time
```
dnnr: Unused

tests/conftest.py
```diff
@@ -388,6 +391,62 @@ def running_clients(hlwm, running_clients_num):
...
+            self.display.sync()
+
+        def is_window_urgent(self, window):
+            """check urgency of a given window handle"""
+            hints = window.get_wm_hints()
+            if hints is None:
+                return False
+            return bool(hints.flags & Xutil.UrgencyHint)
+
+        def create_client(self, urgent=False):
```
dnnr: Will this end up replacing `HlwmBridge.create_client()`? I'm just wondering if we could avoid confusion due to the same name, but OTOH, the name is fine in itself.

tests/conftest.py
```diff
@@ -388,6 +391,62 @@ def running_clients(hlwm, running_clients_num):
...
+            )
+
+            w.set_wm_name( 'Some Window' )
+            if urgent:
+                w.set_wm_hints( flags = Xutil.UrgencyHint )
+
+            w.map()
+            return w
+
+        def shutdown(self):
```
dnnr: This seems to have lost its purpose along the way. Remove?

tests/test_clients.py
```diff
@@ -35,3 +36,43 @@ def test_close_without_clients(hlwm):
...
     hlwm.call('close')
+
+
+@pytest.mark.filterwarnings("ignore:tostring")
+@pytest.mark.parametrize("urgent", [True, False])
+def test_urgent_on_start(hlwm, x11, urgent):
+    # first create a dummy window, such that the second
+    # window is not focused and thus keeps the urgent flag
+    winid_focus, _ = hlwm.create_client()
+    window = x11.create_client(urgent=urgent)
```
dnnr: For added brevity in the tests, maybe you could change `create_client` to return both the `window` and the `windid_str`?

t-wissmann: Right! This also makes it a bit more consistent with hlwm.create_client()!

tests/conftest.py
t-wissmann: I don't get it, but I've added it again.

tests/conftest.py
t-wissmann: Yes!

dnnr:
Oh my machine, this is triggering a memory leak (allocated at the `new Client` in ClientManager) while running `test_urgent_jumpto[False]`. The version it happens on is be05032.

tests/conftest.py
t-wissmann: Yes, I'd vote for using x11.create_client() in the long run (or actually: using it in hlwm.create_client()). I think this will also simplify the keymask-tests a lot.

dnnr: Nice, looking forward to it.

t-wissmann:
I think this is because the destructor of `ClientManager` does not delete the `Client` objects.

tests/conftest.py
t-wissmann: Now I get it and moved the empty line to line 371

tests/test_clients.py
```diff
@@ -35,3 +35,42 @@ def test_close_without_clients(hlwm):
...
+    assert hlwm.get_attr('clients.{}.urgent'.format(winid)) \
+        == hlwm.bool(urgent)
+
+
+@pytest.mark.filterwarnings("ignore:tostring")
+def test_urgent_after_start(hlwm, x11):
+    winid_focus, _ = hlwm.create_client()
+    winid, _ = hlwm.create_client()
+    assert hlwm.get_attr('clients.{}.urgent'.format(winid)) == 'false'
+    assert x11.is_window_urgent(x11.window(winid)) == False
```
dnnr: You can just say `assert not ...` here

tests/conftest.py
```diff
@@ -388,6 +392,53 @@ def running_clients(hlwm, running_clients_num):
...
+            """return python-xlib window wrapper for a string window id"""
+            winid_int = int(winid_string, 0)
+            return self.display.create_resource_object('window', winid_int)
+
+        def winid_str(self, window_handle):
+            return hex(window_handle.id)
+
+        def make_window_urgent(self, window):
+            """make window urgent"""
+            window.set_wm_hints( flags = Xutil.UrgencyHint )
```
dnnr: There should be no whitespace within the brackets here

tests/conftest.py
```diff
@@ -388,6 +392,53 @@ def running_clients(hlwm, running_clients_num):
...
+                return False
+            return bool(hints.flags & Xutil.UrgencyHint)
+
+        def create_client(self, urgent=False):
+            w = self.root.create_window(
+                50, 50, 300, 200, 2,
+                self.screen.root_depth,
+                X.InputOutput,
+                X.CopyFromParent,
+                background_pixel = self.screen.white_pixel,
```
dnnr: This being a keyword argument and not an assignment statement, there should be no spaces around the `=`

tests/test_clients.py
```diff
@@ -45,3 +45,42 @@ def test_close_without_clients(hlwm):
...
+    # make the client urgent:
+    x11.make_window_urgent(x11.window(winid))
+
+    assert hlwm.get_attr('clients.{}.urgent'.format(winid)) == 'true'
+    assert x11.is_window_urgent(x11.window(winid))
+
+
+@pytest.mark.parametrize("explicit_winid", [True, False])
+def test_urgent_jumpto(hlwm, x11, explicit_winid):
+    winid_old_focus, _ = hlwm.create_client()
```
dnnr: You don't seem to need the window ID. Maybe just ignore the entire return value and use the gained space for a comment long these lines?
```
hlwm.create_client()  # dummy client that gets the focus
```
(the same should be possible in the other two tests)

tests/test_clients.py
```diff
@@ -45,3 +45,42 @@ def test_close_without_clients(hlwm):
...
+
+    # make the client urgent:
+    x11.make_window_urgent(x11.window(winid))
+
+    assert hlwm.get_attr('clients.{}.urgent'.format(winid)) == 'true'
+    assert x11.is_window_urgent(x11.window(winid))
+
+
+@pytest.mark.parametrize("explicit_winid", [True, False])
+def test_urgent_jumpto(hlwm, x11, explicit_winid):
```
dnnr: Test name suggestions: `test_urgent_jumpto_makes_client_nonurgent` or `test_urgent_jumpto_resets_urgent_flag`

t-wissmann: I've renamed it!

tests/test_clients.py
t-wissmann: I've dropped these three variables.

PR #412, opened 2019-04-13
Add configuration file for CodeCov

dnnr:
Adding this file to the master branch is an attempt at getting CodeCov
to actually interpret the file.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+    precision: 1
+
+comment:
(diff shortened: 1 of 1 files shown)


dnnr:
My suspicion is that CodeCov behaves the same as Mergify, in that it only considers the file as it is present in the GitHub *default branch*, which is master for us.

t-wissmann:
I've just approved the installation request. Let's see if it helps!

PR #413, opened 2019-04-13
Add test for hlwm shutdown with running clients

dnnr:


dnnr:
I haven't looked close, but I think we certainly do have a couple of tests that work on the assumption that there are no already-running clients.

PR #415, opened 2019-04-13
Drop several unused includes in object.cpp

dnnr:


dnnr:
The compiler works in mysterious ways

PR #418, opened 2019-04-13
Kill existing clients before test case execution

t-wissmann:


tests/conftest.py
```diff
@@ -366,18 +367,39 @@ def shutdown(self):
                                 + " and had to be killed") from None
 
 
+def kill_all_existing_windows(show_warnings=True):
+    xlsclients = subprocess.run(['xlsclients', '-l'], stdout=subprocess.PIPE)
```
dnnr: I think you can use `check_output` instead of `run` and then you do not need the `assert`

dnnr: (alternatively, passing `check=True` should also work)

tests/conftest.py
```diff
@@ -366,18 +367,39 @@ def shutdown(self):
...
+    assert xlsclients.returncode == 0
+    clients = []
+    for l in xlsclients.stdout.decode().splitlines():
+        m = re.match(r'Window (0x[0-9a-fA-F]*):', l)
+        if m:
+            clients.append(m.group(1))
+    if len(clients) > 0:
```
dnnr: You could reduce the indendation here by saying `if clients and show_warnings:` here, and moving the `for` one level up (it doesn't care whether the list is empty).

dnnr:
Looks like the build failed because this happened (randomly?): `xlsclients:  unable to open display ":1009"`. Could this just be one of the display number race conditions that we sometimes see?

PR #424, opened 2019-04-13
Drop some unused #include "glib-backports.h"

dnnr:
Removing these reveals that floating.cpp is missing an include of
<climits>, which is added.


dnnr:
I've noticed that this is a subset of #390, but given that it is basically independent of actually removing the dependency, I thought it warrants a separate change.

PR #428, opened 2019-04-14
Fix includes in stack.h (drop memory, add set)

dnnr:
<memory> (things like shared_ptr) is not used, but <set> is.

### DIFF
--- a/src/stack.h
+++ b/src/stack.h
@@ -3,7 +3,7 @@
 
 #include <X11/X.h>
 #include <array>
-#include <memory>
+#include <set>
 #include <vector>
 
 #include "types.h"


dnnr:
Seems like GitHub doesn't like angle brackets in descriptions. Let's see how this ends up looking in the squash commit. :thinking:

t-wissmann:
Angle brackets in the description are fine. I was confused when I first saw
https://github.com/herbstluftwm/herbstluftwm/pull/421 but after hitting 'edit' I've noticed that a meaningful description is still there, and the final commit message seems fine: https://github.com/herbstluftwm/herbstluftwm/commit/5555f88527ab0dafcb611cfe76dd0627f9b61ea1

PR #430, opened 2019-04-14
CodeCov: Disable PR status checks for now

dnnr:
This is a rather new feature, but not yet very helpful because of our
known issues with flaky coverage.


dnnr:
I'm hoping that the master thing and editing the file in the CodeCov settings was a red herring, and everything was fixed with #427.

PR #431, opened 2019-04-16
Avoid iwyu fix_include output for unchanged files

t-wissmann:
Filter the output of fix_include by removing messages
of the pattern:

    >>> Fixing #includes in 'somefilename.cpp'
    No changes in file somefilename.cpp

This uses subprocess.run because the documentation of subprocess.check_call
says that it should not be combined with piped stdout (stdout=subprocess.PIPE).

### DIFF
--- a/ci-build.py
+++ b/ci-build.py
@@ -80,11 +80,14 @@ if args.ccache:
 
 if args.iwyu:
     # Check lexicographical order of #include directives (cheap pre-check)
-    fix_include = sp.run('fix_include --dry_run --sort_only --reorder /hlwm/*/*.{h,cpp,c}', shell=True, executable='bash', stdout=sp.PIPE, check=True)
+    fix_include = sp.run('fix_include --dry_run --sort_only --reorder /hlwm/*/*.{h,cpp,c}', shell=True, executable='bash', stdout=sp.PIPE)
     print(re.sub(
         r">>> Fixing #includes in '[^']*'[\n\r]*No changes in file [^\n\r]*[\n\r]*",
         '',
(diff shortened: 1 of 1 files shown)


t-wissmann:
See https://travis-ci.org/herbstluftwm/herbstluftwm/jobs/520641831 for the old version of the output. Note that both logs still show the summary

    IWYU edited 0 files on your behalf.

Which is fine in my opinion, and it shows that not the entire `fix_include` is hidden.

dnnr:
Great idea, and the change looks good to me, but have you verified how it looks when `fix_include.py` actually finds a problem? I can check it myself, but not right now.

t-wissmann:
I haven't verified how it looks at first. And I'm not sure whether I can entirely say what is done by `iwyu_tool` and what by `fix_includes`. Is it correct that `iwyu_tool` finds wrong/missing includes and that `fix_includes` only groups and orders them?

With the new commit in this PR, the output looks at follows if fix_include finds a problem: https://travis-ci.org/herbstluftwm/herbstluftwm/jobs/520780308

dnnr:
> I haven't verified how it looks at first. And I'm not sure whether I can entirely say what is done by `iwyu_tool` and what by `fix_includes`. Is it correct that `iwyu_tool` finds wrong/missing includes and that `fix_includes` only groups and orders them?

Pretty much, yeah.

> With the new commit in this PR, the output looks at follows if fix_include finds a problem: https://travis-ci.org/herbstluftwm/herbstluftwm/jobs/520780308

This looks totally fine to me, let's do it.

PR #432, opened 2019-04-19
Avoid wireAttributes() in Client

t-wissmann:
Also, group attributes together (even though some of them are not
attributes in winterbreeze yet, but only in master)

### DIFF
--- a/src/client.h
+++ b/src/client.h
@@ -22,24 +22,12 @@ public:
 
     Window      window_;
     Decoration  dec;
-    Rectangle   float_size_ = {0, 0, 100, 100};     // floating size without the window border
-    Attribute_<bool> urgent_ = {"urgent", false};
-    Attribute_<bool> fullscreen_ = {"fullscreen", false};
-    Attribute_<std::string> title_ = {"title", {}};  // or also called window title; this is never NULL
-    Slice* slice = {};
     Rectangle   last_size_;      // last size excluding the window border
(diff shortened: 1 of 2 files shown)


src/client.h
```diff
@@ -51,6 +39,21 @@ class Client : public Object {
...
+
+    // attributes:
+    Attribute_<bool> urgent_;
+    Attribute_<bool> fullscreen_;
+    Attribute_<std::string> title_;  // or also called window title; this is never NULL
+    Attribute_<std::string> window_id_str;
+
+    Attribute_<std::string> keyMask_; // keymask applied to mask out keybindins
+    Attribute_<bool> pseudotile_; // only move client but don't resize (if possible)
+    bool        ewmhrequests_ = true; // accept ewmh-requests for this client
```
dnnr: Why these spaces after `bool`?

t-wissmann: I've only reordered lines and this was from a previous alignment.

dnnr: Ah! I'd suggest cleaning that up while at it. It wouldn't make the diff any harder to read, imho.

t-wissmann: Ah, I've just noticed the "(keep)" in your question.
These lines will eventually be replaced by `Attribute_<bool>`, so let us please not worry about the spaces for the Moment.

dnnr: Alright, nevermind then.

PR #433, opened 2019-04-19
Move window property utility functions to XConnection

t-wissmann:


t-wissmann:
I don't get what mergify is complaining about

PR #434, opened 2019-04-19
Add pid attribute to Client

t-wissmann:


t-wissmann:
This includes #432.

tests/test_clients.py
```diff
@@ -84,3 +84,13 @@ def test_urgent_jumpto_resets_urgent_flag(hlwm, x11, explicit_winid):
...
+
+
+def test_client_with_pid(hlwm):
+    winid, proc = hlwm.create_client()
+    assert int(hlwm.get_attr('clients.focus.pid')) == proc.pid
+
+
+def test_client_without_pid(hlwm, x11):
+    _, winid = x11.create_client()
+    assert int(hlwm.get_attr('clients.focus.pid')) == -1
```
dnnr: Is it obvious (just not to me) why this client does not have a PID? Is that some "side-effect" of an x11 fixture client in contrast to xterm? Unless I'm missing something, this feels to me like there should be short explanation of *why* we expect this to be -1.

t-wissmann: It's a side effect because `x11.create_client()` simply does not set the pid.

In the long run, I've planned to add an optional parameter `pid` to `x11.create_client()` that accepts an arbitrary int or `None`, so with `x11.create_client(pid=None)` it should become obvious.

dnnr: Ah okay. That will certainly clear things up. You *could* already add that pid parameter, and put a `assert not pid, "PID setting not yet support`, but that might be more effort than actually implementing it (I have no idea tbh).

t-wissmann: I've just added a comment to the test case

t-wissmann: I've implemented the `pid`-argument in PR #443.

PR #435, opened 2019-04-19
Add tag attribute to Client

t-wissmann:


t-wissmann:
This includes/requires #432

dnnr:
There a conflicts that need to be resolved. (I really don't get why they don't indicate this more prominently in the PR list... there should not be a green check mark. :-1: )

t-wissmann:
I've resolved the conflicts.

src/client.cpp
```diff
@@ -695,3 +696,9 @@ void Client::clear_properties() {
     XDeleteProperty(g_display, window_, g_wmatom[WMState]);
 }
 
+//! name of the tag on which the client is
+string Client::tagName() {
+    // be save during initialization phase and don't assume
```
dnnr: Typo: safe

tests/test_clients.py
```diff
@@ -86,6 +86,18 @@ def test_urgent_jumpto_resets_urgent_flag(hlwm, x11, explicit_winid):
     assert hlwm.get_attr('clients.focus.urgent') == 'false'
 
 
+def test_client_tag_attribute(hlwm):
+    othertag = 'othertag'
+    winid, _ = hlwm.create_client()
+    hlwm.call(['add', othertag])
```
dnnr: Imho, it would be perfectly fine to just say `hlwm.call('add othertag')` here and ` == 'othertag'` at the end. We're doing thinks like that in many tests (with foo, bar, etc.) and it just keeps things simple.

src/client.cpp
t-wissmann: Fixed!

tests/test_clients.py
t-wissmann: I wanted to emphasize that here the actual name does not matter. I've inlined the definition of `othertag`, since you prefer it for simplicity.

dnnr: Thanks! I think the important part is to choose a name that clarifies the fact that it is "random but fixed". Then a shared variable doesn't add much more value. In this case, "othertag" does that perfectly (so would "sometag", "whatever", "foobar", etc.).

This PR now looks good to me, but there is another conflict to resolve.

t-wissmann: I've just resolved the conflict, and so you need to approve this PR again.

PR #436, opened 2019-04-19
Remove string_to_bool, use Converter<bool> instead

t-wissmann:


src/types.h
```diff
@@ -57,6 +57,9 @@ struct Converter {
      * 'Source' may be given relative to 'relativeTo', e.g. "toggle" for booleans.
      */
     static T parse(const std::string& source, T const* relativeTo);
+    static T parseDefault(const std::string& source, const T& relativeTo) {
```
ypnos: :rabbit: The naming of this method confuses me. First I thought it indicates that the second value should be used as a default when parsing fails. But then I realized it is not used like that. What does `default` stand for here? Or could this method also just share the name with existing `parse()`?

t-wissmann: It can't be called `parse()` because then the call to `Converter<bool>::parse()` is ambigous. What do you think about the names `parseWithDefault` or `parseRelativeTo`, or something else?

src/types.h
```diff
@@ -86,19 +93,24 @@ inline std::string Converter<bool>::str(bool payload) {
...
+}
+template<>
+inline bool Converter<bool>::parse(const std::string &payload, const bool& previous) {
+    if (payload == "toggle")
+        return !previous;
+    try {
+        return parse(payload);
+    } catch (std::invalid_argument&) {
+        throw std::invalid_argument("only on/off/true/false/toggle are valid booleans");
+    }
```
t-wissmann: That's a neat trick!

src/types.h
```diff
@@ -53,13 +53,20 @@ class Input : public ArgList {
...
+    /** Parse a text into the right type
+     * @brief overload that accepts a reference value to relate input string to
+     * @param source may be given relative to 'relativeTo', e.g. "toggle" for booleans.
+     * @param relativeTo previous/existing value
+     * @throws std::invalid_argument or std::out_of_range
+     */
+    static T parse(const std::string& source, const T& relativeTo) {
+        // note: specializations _do_ use relativeTo where appropriate
+        return parse(source);
+    };
```
t-wissmann: We only use the `realtiveTo` for booleans (so far), so this is a very sensible compromise! :-)

PR #438, opened 2019-04-21
Add ewmhnotify attribute to Client

t-wissmann:
Add and test the new attribute `ewmhnotify` of Client. This also
adds a new python dependency ewmh to the test suite (tox.ini)


t-wissmann:
This PR requires #432

src/client.h
```diff
@@ -105,6 +105,7 @@ class Client : public Object {
     void clear_properties();
     bool ignore_unmapnotify();
 
+    void updateEWMHState();
```
dnnr: I don't think we ever discussed a consistent style for this (or even the need for it), but just in case you're interested in establishing one: As far as I know, the most common way of handling acronyms/initialisms in camelCase is by lower-casing all but the first character. Here, it would be `updateEwmhState`.

My primary reason for recommending it is that it retains the role of upper-case characters as word separators. This really only gets important when there is more than one acronyms in a row (non-sensical example: `startEwmhGui` vs. `startEWMHGUI`).

t-wissmann: Seems like a reasonable argument to me. I've renamed it.

dnnr:
I don't see anything wrong with it, but I don't think I can fully judge it.

t-wissmann:
> I don't see anything wrong with it, but I don't think I can fully judge it.

If you're fine with the tests, I suggest that we move forward :-)

dnnr:
Fair enough! Approved with low confidence ;-)

PR #439, opened 2019-04-21
Rename ChildByIndex to IndexingObject

ypnos:
I always get confused when I see this class about what it does. Hope the proposed name is liked as well the old one.


dnnr:
Thanks for this change! I felt the same about the old name, and this is definitely an improvement. 👍

PR #440, opened 2019-04-22
Let call_xfail_no_output() assert no output

t-wissmann:
In call_xfail_no_output(), assert that the command has indeed no
output. I think the former reading of the command name was "make
no assertion about the output".

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -110,6 +110,7 @@ class HlwmBridge:
     def call_xfail_no_output(self, cmd):
         proc = self.unchecked_call(cmd)
         assert proc.returncode != 0
+        assert proc.stderr == ""
         return proc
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -110,6 +110,7 @@ def f(self2, reg):
     def call_xfail_no_output(self, cmd):
         proc = self.unchecked_call(cmd)
         assert proc.returncode != 0
+        assert proc.stderr == ""
```
dnnr: Didn't we, some time in the past, also assert that stdout is empty in an xfail call? That doesn't seem to be the case anymore. Not sure if we dropped it on purpose, though.

In any case, it's probably better to call `call_xfail` here instead of `unchecked_call`, since it is a specialization of `call_xfail`. Cleaning that up might well be done separately, so this PR looks good to me!

PR #443, opened 2019-04-24
tests: Explicitly set PID in x11.create_client()

t-wissmann:


tests/conftest.py
```diff
@@ -465,6 +465,12 @@ def create_client(self, urgent=False):
             if urgent:
                 w.set_wm_hints(flags=Xutil.UrgencyHint)
 
+            if pid is not None:
```
dnnr: You can just say `if pid` here (...is what The-Compiler would say :wink: )

t-wissmann: Is this really equivalent for `pid=0`?

dnnr: Oh, you're totally right. Disregard that.

t-wissmann: Being more explict: assume that I call `create_client` with `pid=0`, then `pid is not None` is `True` and we indeed want the x-property to be set. However `bool(pid)` evaluates to `False` and this would cause the xproperty to be left unset (which we don't want).

t-wissmann:
The mergify is broken again, and I don't get it.

PR #445, opened 2019-04-24
Replace plain enum with SplitAlign enum class

dnnr:
This also drops ALIGN_EXPLODE in favor of a temporary variable in the
split command handler.


dnnr:
I'm still a bit overwhelmed by the huge diff in #388, so I took the liberty to extract this self-sufficient change from it.

PR #447, opened 2019-04-26
ci-build.py: Do not assume /hlwm as repo root

dnnr:
This removes the rather hard dependency on executing in a Docker
environment where the repository is mounted at /hlwm.

Beware that the script modifies the ccache configuration. Passing a
custom folder to --ccache is highly recommended.

### DIFF
--- a/ci-build.py
+++ b/ci-build.py
@@ -32,11 +32,12 @@ parser.add_argument('--ccache', nargs='?', metavar='ccache dir', type=str,
                     const=os.environ.get('CCACHE_DIR') or True)
 args = parser.parse_args()
 
+repo = Path(__file__).resolve().parent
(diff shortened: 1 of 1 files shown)


t-wissmann:
> This removes the rather hard dependency on executing in a Docker
> environment where the repository is mounted at /hlwm.
> 
> Beware that the script modifies the ccache configuration. Passing a
> custom folder to --ccache is highly recommended.

I don't get the connection between this PR  and `--ccache`

dnnr:
The connection is that now that you can run ci-build.py directly, without Docker, it has become easier to inadvertently truncate your existing ccache to 500MB by accidentally passing `--ccache`. I just mentioned it as a reminder.

PR #450, opened 2019-04-27
Allow the monitor name to be empty

t-wissmann:
Monitor names are optional and a monitor with no name has the name
attribute empty. Only for add_monitor, it is not permitted to pass an
empty monitor name.

This is the current behaviour in master.


t-wissmann:
Names are optional for monitors, so we want to create monitors with names, but also without names.

PR #451, opened 2019-04-27
Move rename_monitor command to Monitor

t-wissmann:


t-wissmann:
For a better test coverage, we can wait until #450 is merged.

dnnr:
The change in itself looks perfectly fine to me.

But it makes me suspect there is some arbitrariness to where we put functions like this. At first, I briefly wondered why this isn't in MonitorManager. But I couldn't think of any strong reason to prefer *either* Monitor or MonitorManager.

I'm tending towards Monitor because keeping track of names is not strictly necessary to do a manager's job (which is keeping track of existences). That said, I think a manager should always have control of exactly one identifier for the objects in manages. But in this case (and probably most) that role is already fulfilled by the index not the name, right?

In any case, renaming tags is implemented in TagManager, so there is some inconsistency now. If you're sure of the pattern as you implemented it here, I suppose we should move it for tags as well?

t-wissmann:
My impression is that nearly all of the old code in monitor.cpp belongs to MonitorManager. For remaining/renaming functions, I'm not entirely sure. This particular implementation of `rename_monitor` is just a wrapper around the attribute setter, so I think `Monitor` is fine.

And I think for renaming tags, the code should look similar.

PR #452, opened 2019-04-27
Move unusedTag() to TagManager

t-wissmann:


dnnr:
Mergify is stuck again. It's saying "Required status check "continuous-integration/travis-ci/push" is pending." This might not even be an issue with Mergify but one with Travis. We could try to migrate to travis-ci.com (which is apparently using a newer GitHub API) and see if it helps. Here's more info on that: https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/
The migration is something that is supposed to be happening to all repos at some point. I've already done it for my own hlwm fork and haven't seen any issues.

PR #454, opened 2019-05-01
Do tag renaming validation in TagManager

t-wissmann:
This makes the tag renaming a bit more similar to the renaming of
monitors.


dnnr:
To be honest, I'm doubting the merit of this change. Maybe there was a misunderstanding in #451, but I thought that we consider the name to be something solely under the control of the object and not its manager?

If the only things that are relevant to a manager are the object identity (as represented by its memory address) and its index (as given by the manager's IndexingObject base class) then why would the manager treat the name attribute any different than all the other attributes of the objects it manages? This seems needlessly inconsistent to me.

t-wissmann:
> To be honest, I'm doubting the merit of this change. Maybe there was a misunderstanding in #451, but I thought that we consider the name to be something solely under the control of the object and not its manager?

It's hard for me to answer in this level of generality, because of course names should be unique, so the name can't be 'solely' under the control of the object. To be more explicit: I think the function `isValidTagName` fits best in `TagManager`. Do you agree with this?

> If the only things that are relevant to a manager are the object identity (as represented by its memory address) and its index (as given by the manager's IndexingObject base class) then why would the manager treat the name attribute any different than all the other attributes of the objects it manages? This seems needlessly inconsistent to me.

Well because the name attributes of all the tags need to be distinct, the other attributes don't have such a requirement.

To clear things up: Can you show me the part of the diff of this PR you disagree with?

dnnr:
Yeah right, I forgot about the uniqueness criterion. That *does* in fact make the name more special than other attributes. Sorry, I judged a bit too quickly and didn't actually look at what the validation does, and assumed that it's mostly about syntax validation.

Taking this into account, I'd say it's fine doing it like this. But just as an idea, let me propose an alternative that would separate the concerns more strictly (but I'm not sure if it adds unnecessary complexity):
 * Add a `bool isAvailableTagName(string)` to TagManager that provides only the part of the validation for which TagManager is really responsible.
 * Let HSTag retain the validation method, which is then a *composition* of syntax validation (which, in case of tags, is only the non-emptiness criterion) *and* asking TagManager whether the name is available.

Ultimately, that's not too different to the old code, but it does delegate the "already exists" check to TagManager instead of having HSTag iterating over the tag names, which is really not how it should be.

But again, if that doesn't sound any better to you, let's just do it like you did.

t-wissmann:
In the meantime, I've noticed another reason why I think also the non-emptiness check (and all the other checks in the case of `MonitorManager::isValidMonitorName`) belong to the manager: The manager needs to ensure that one can address the manager's children via the name, and so the manager should decide what names are valid and which names are not (because the syntactically interfere with other adressing methods like (relative or absolute) indices and because this is used in the `by-name` child object). This is the reason why `MonitorManager::isValidMonitorName` is so restrictive and this is why `""` is not a valid tag name. And I think the tag and the monitor shouldn't even be aware of the concrete rules regarding names because the restrictions only exist in order to make other code of `monitormanager.cpp`/`tagmanager.cpp` work.

dnnr:
These are very good points. You've convinced me that this is the way to go. :+1:

src/tagmanager.cpp
```diff
@@ -39,13 +39,26 @@ HSTag* TagManager::find(const string& name) {
...
 
+//! if the name is a valid tag name, return "", otherwise return an error message
+string TagManager::isValidTagName(string name) {
+    if (name.empty()) {
+        return "An empty tag name is not permitted";
+    }
+    if (find(name)) {
+        return "A tag with the name \"" + name + "\" already exists";
+    }
+    if (find_monitor_by_name(name.c_str())) {
```
dnnr: Just one more thing... what's the purpose of this? Am I missing something or is this just not doing anything?

t-wissmann: Looks like a bad copy&paste mistake from monitormanager. This shouldn't be here, and I've removed this! Thanks!

PR #456, opened 2019-05-01
Add tests fixture to log hooks

t-wissmann:
Add a hc_idle fixture that returns a list of hooks that have been
emitted so far.


t-wissmann:
For some reason this fails in travis but works locally here. I can't explain why a race condition drops hooks...

t-wissmann:
And now that I've  added debug output, the quantum decided to let the build pass...

t-wissmann:
The build just failed for another PR because of this hook issue. I think we should disable the hook tests / revert the commit until we understand it fully..

dnnr:
Agreed.

PR #457, opened 2019-05-01
Test that 'rename' emits tag_rename hook

t-wissmann:


t-wissmann:
This includes #454 and #455 and requires #456 to work.

PR #461, opened 2019-05-01
Move emit_hook to IpcServer

t-wissmann:
The hook communication is a core part of the ipc protocol and thus fits
perfectly into the IpcServer class. Note that on the client side, the
'hook' and 'call' are already grouped in a module ipc-server.c.

During this migration, some variable names changed:

  - g_event_window -> hookEventWindow_
  - last_property_number -> nextHookNumber_


t-wissmann:
This includes #460

src/ipc-server.h
```diff
@@ -2,22 +2,30 @@
...
     //! try to run an ipc request in the given window, return if there was one
     bool handleConnection(Window window);
+    //! send a hook to all listening clients
+    void emitHook(std::vector<std::string> args);
 
 private:
     XConnection& X;
+
+    Window hook_event_window_; //! window on which the hooks are announced
+    int next_hook_number_; //! index for the next hook
```
dnnr: Use camelCase here, as usual?

t-wissmann: Right! I've fixed this (also in this PR's description)

t-wissmann:
I'm merging this since this is the approved PR modulo obvious merge conflict resolutions.

PR #462, opened 2019-05-01
Turn ewmh_ functions into members of a Ewmh class

t-wissmann:
Make all the ewmh_-functions member functions of a new EWMH
class. Temporarily, this class has a singleton getter Ewmh::get()
that wraps Root:get() and avoids that we need to include root.h
whenever we need some ewmh functionality.

In principle, this only makes syntactic changes (name of functions,
and migration to a namespace with a this argument, ...) but does not
change the control flow itself.


src/client.cpp
```diff
@@ -166,7 +167,7 @@ void Client::window_unfocus_last() {
     if (lastfocus) {
         /* only emit the hook if the focus *really* changes */
         hook_emit_list("focus_changed", "0x0", "", nullptr);
-        ewmh_update_active_window(None);
+        Ewmh::get().updateActiveWindow(None);
```
dnnr: Is there any reason why it's `Ewmh::get().` here but `ewmh.` in the other places in this file?

src/ewmh.h
```diff
@@ -69,51 +67,66 @@ enum {
...
+    void windowUpdateWmState(Window win, WmState state);
 
-void window_update_wm_state(Window win, WmState state);
+    static Ewmh& get(); // temporary singleton getter
+
+private:
+    bool focusStealingAllowed(long source);
+    bool readClientList(Window** buf, unsigned long *count);
+    Root* root_ = nullptr;
+    XConnection& X;
```
dnnr: This should be a lower-case `x`, shouldn't it?

src/main.cpp
```diff
@@ -564,6 +564,10 @@ void event_on_configure(Root*, XEvent event) {
     }
 }
 
+void clientmessage(Root* root, XEvent* event) {
```
dnnr: Could be made `static`

src/client.cpp
t-wissmann: Because this function is static, whereas the other places are in non-static functions where we have the `ewmh` member available.

dnnr: Ah, right! That was my first thought, too, but apparently I forgot that the `static` would not be here but in the header, and then I assumed it to be non-static ;-)

src/ewmh.h
t-wissmann: Right now, I always use `X` in classes where `X` is actually used, because I think `X.display()` looks much better than `x.display()` or `x_.display()` or `X_.display()`. A lower-case `x` looks like the `x`-coordinate.

src/main.cpp
t-wissmann: I think it does not matter much, because I'm planning to cut the `main.cpp` into smaller pieces by moving the Xlib-specific main-loop to an extra class.

t-wissmann: But I've made it static for the moment.

src/ewmh.h
dnnr: Mhm, I see the issue. How do you think about `xConn`?

t-wissmann: I think `xConn.display()` in every second line looks neither helpful nor nice. What's the issue with `X` (or maybe with `X_`)? Is this only about obeying the law "Every variable name must be in camelCase" or is there any risk of confusion where the variable is used? As a compromise, we can name it `X_` to make it clear that `X_` is a member variable.

dnnr: It's really only about staying with the naming convention, so it's truly a minor concern. I can live with `X`/`X_`, but I wanted to check if there's maybe a third way that makes everybody happy.

Adding the underscore suffix would be appropriate regardless of the casing, but since we're breaking the rules anyway, just go with what you prefer here. ;-)

t-wissmann: I've named it `X_` to make it easier for the reader to see that this is a member variable, even though it looks a bit different because of the upper case X.

PR #463, opened 2019-05-02
Make valgrind-xephyr.sh more configurable

t-wissmann:
Make the autostart file and the debugger (e.g. gdb intead of valgrind)
configurable vial command line parameters.


dnnr:
I couldn't try this out, probably because my glibc wasn't built right (so, not your fault): https://gist.github.com/dnnr/d6e916424ee1ad8c81e50945e143d147

But it *looks* fine, so let me know if I should just approve it without having tested it.

t-wissmann:
As commented in the gist, I think valgrind is not compatible when the binary is built with the gcc-asan-flags. But you should be able to run the script with `--none` or `--gdb`.

dnnr:
I see, thanks! It does work with `--none` and `--gdb`, but removing the sanitizer flags does make valgrind work. Judging by this thread, I think this is an issue on my end: https://forums.gentoo.org/viewtopic-t-1069032-start-0.html

valgrind-xephyr.sh
```diff
@@ -10,6 +10,55 @@ die() {
...
 
+cmdname="$0"
+usage() {
+cat <<EOF
+Usage: $cmdname FLAGS
+
+Start a Xephyr and run herbstluftwm in it. Per default, valgrind
+is used. Other debug environments are:
+
+    --valgrind          Run in valgrind (default)
```
dnnr: Hmmm, the default doesn't work as advertised. Is the `|''` case supposed to achieve this? I'm not entirely sure but I wouldn't expect this `for` loop to run at all if there are no arguments.

I think you could move the `debugger() { valgrind "$@" }` in front of the loop and do nothing in the `--valgrind` case. Bash doesn't mind function redefinitions.

t-wissmann: Oh you're right! Originally I've only checked `$1` and later turned it into a loop. I've fixed this, thanks!

dnnr: Works now! :+1: But I think you forgot to remove the `|''` from the case

t-wissmann: Gnah, sorry. I've removed this. Thanks!

PR #464, opened 2019-05-03
Update Mergify config to work with travis-ci.com

dnnr:
On travis-ci.com, the PR status check appears to have a different "name"
than on travis-ci.org. Changing it like this might help.


dnnr:
Unfortunately, the only way to find out if it works seems to be merging it...

dnnr:
Ah, it's even mentioned here: https://doc.mergify.io/conditions.html (look for status-success).

t-wissmann:
Great!

PR #465, opened 2019-05-04
Construct stack via lambda function in toWindowBuf

t-wissmann:
Instead of letting Stack::toWindowBuf() return a vector of Window
elements, we now directly construct the stack, by passing a lambda
function addToStack(), which takes care of adding the element to the
stack as appropriate for the current scenario.

With this, the stack construction is more elegant:

  - In toWindowBuf(), we don't need to append the vectors that come out
    of recursive calls but can simply pass the addToStack() parameter to
    the reursive calls.
  - In Monitor::restack(), we don't need to remove the window of a
    focused fullscreen window from the stack, but we can modify the
    lambda function addToStack() such that a focused fullscreen window
    is never added to the stack at all.


src/stack.cpp
```diff
@@ -163,56 +164,57 @@ int print_stack_command(int argc, char** argv, Output output) {
     return 0;
 }
 
-static vector<Window> slice_to_window_buf(Slice* s, bool real_clients, HSLayer layer) {
+//! helpfer function for Stack::toWindowBuf() for a given Slice and layer. The
```
dnnr: Typo (Helper)

dnnr:
I'm happy with the code, but the names of things involved here continue to confuse me.

So, `toWindowBuf()` is method of `Stack` that exists to provide a list of all contained Windows, ordered by the stacking order, right? There is no buffering going on? (same goes for `slice_to_window_buf`)
And `addToStack` is the callback that is receiving the windows *from* the stack to add them *to* that list, right? It doesn't add anything to the `Stack`? Granted, there is not much difference between "list" and "stack", so the resulting list could well be referred to as a "stack", too, but why not avoid that ambiguity...

How would you feel about naming them like this?
 * `void Stack::getWindows(..., function<void(Window)> yield)`
 * `void Slice::getWindows(..., function<void(Window)> yield)` (assuming it's already turned into a Slice member, so maybe later/separately)

Naming the callback `yield` focuses on its purpose, discarding the irrelevant details on what happens with the callback parameter. The callback might not even do any "adding" at all (but printing, searching for a match, etc.), and toWindowBuf/getWindows doesn't have any knowledge of that. Not anymore, that is.

The only part I'm unsure of is the `get`, because it kind of suggests a non-void return value. But I'd still prefer it over alternatives like `{to,as}Window{List,Vector}`. At the very least, I'd like to get rid of the `Buf`.

t-wissmann:
All your observations on the functionality are right.

I agree that the `Buf` is inappropriate here, and I didn't notice that `addToStack` is misleading. So I think `yield` is fine and I think `getWindows` is fine. Do you think `extractWindows` is a better name? Make this emphasizes that something is constructed here and that the function is not just a read-only getter for some member variable.

dnnr:
Good idea, I like `extractWindows` even better :+1:

t-wissmann:
I've renamed it, and I've also renamed `addToStack` to `addToVector`.

PR #466, opened 2019-05-04
Remove hook_emit_list

t-wissmann:
Passing multiple parameters via a vector/initializer list is shorter,
more type safe, and makes it easier to pass std::string.


t-wissmann:
I've removed the unnecessary include of `stdarg.h`.

PR #467, opened 2019-05-04
Revert "Add tests fixture to log hooks (#456)"

t-wissmann:
This reverts commit 5bbd6682064835143f8b68493211e84e4249b21d.

There is some race condition that causes the hook logger to miss some
hooks. So we should revert this feature until we understand it fully.


t-wissmann:
To avoid issues in the other PR, I'm merging this directly.

PR #468, opened 2019-05-04
Add tests fixture to log hooks

t-wissmann:
Add a hc_idle fixture that returns a list of hooks that have been
emitted so far.


t-wissmann:
I think I understand the issue here, but I don't have a solution yet: The first hook in the test case is already emitted before the `hc --idle`-process has fully connected to the xserver and is listening for hooks.

t-wissmann:
Alright, here we go!

In the long run, we could think of adding a command line option to herbstclient that lets it print some message as soon as the connection to hlwm is established.

Independently from the present issue, python-bindings to herbstclient (or actually ipc-client/ipc-client.h) would be preferable. However, there seem to be many ways to set up python bindings. @The-Compiler: is this right? or is there a canonical way for setting up python bindings for C-header files?

tests/conftest.py
```diff
@@ -387,6 +388,60 @@ def shutdown(self):
...
+        number_sent = 0
+        while [] == select.select([self.proc.stdout], [], [], 0.1)[0]:
+            # while there hasn't been a message received, send something
+            number_sent += 1
+            self.hlwm.call(['emit_hook', 'hc_idle_bootup', number_sent])
+        # now that we know that self.proc is connected, we need to consume
+        # its output up to the last message we have sent
+        assert number_sent > 0
+        number_received = 0
+        while number_received < number_sent:
```
dnnr: So it's guaranteed that all emitted hooks are echoed by the process eventually? Isn't that the same assumption that proved to be wrong in the first place (i.e., an expected hook being emitted before `hc --idle` was even ready)?

Going by you explanation of the race condition, I would have thought that, for that exact same reason, you could *not* wait for `number_received == number_sent` here.

dnnr: (I've deleted a comment here, nevermind)

tests/conftest.py
```diff
@@ -387,6 +388,60 @@ def shutdown(self):
...
+            number_received = int(line[1])
+
+    def hooks(self):
+        """return all hooks since the last call of this function"""
+        # collect all hooks so far, so collect all up to the following hook:
+        self.hlwm.call(['emit_hook', 'hc_idle_logging_marker'])
+        hooks = []
+        while True:
+            line = self.proc.stdout.readline().decode().rstrip('\n').split('\t')
+            print("hook logger: received line: '{}'".format(line))
```
dnnr: This was only for debugging, right? Do we still need it? (If yes, maybe tweak the prefix a bit... `HcIdle: received hook: `)

t-wissmann: Right! I think I should remove it.

PR #469, opened 2019-05-04
Make IpcServer the counterpart of ipc-client.h

t-wissmann:
Make the interface of the IpcServer the precise counterpart of
ipc-client.h. This is also removes IpcServer's dependency on command.h.
Now, the IpcServer calls a callback for each request received from a
(herbst-)client. The result of the callback is then sent back to the
(herbst-)client.

Formerly, the IpcServer called a function from command.h directly. Now, the
only in-project dependencies of IpcServer are ipc-protocol.h and xconnection.h.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -232,6 +232,18 @@ unique_ptr<CommandTable> commands(shared_ptr<Root> root) {
     return unique_ptr<CommandTable>(new CommandTable(init));
 }
 
+//! wrapper around Commands::call()
+static std::pair<int,std::string> callCommand(const vector<string>& call) {
+    // the call consists of the command and its arguments
+    std::ostringstream output;
+    auto input =
+        (call.size() == 0)
+        ? Input("", call)
+        : Input(call[0], vector<string>(call.begin() + 1, call.end()));
+    int status = Commands::call(input, output);
+    return make_pair(status, output.str());
+}
+
 // core functions
 int quit() {
     g_aboutToQuit = true;
@@ -714,6 +726,7 @@ void createnotify(Root* root, XEvent* event) {
     // printf("name is: CreateNotify\n");
     if (root->ipcServer_.isConnectable(event->xcreatewindow.window)) {
         root->ipcServer_.addConnection(event->xcreatewindow.window);
+        root->ipcServer_.handleConnection(event->xcreatewindow.window, callCommand);
     }
 }
 
@@ -833,7 +846,7 @@ void propertynotify(Root* root, XEvent* event) {
     Client* client;
     if (ev->state == PropertyNewValue) {
         if (root->ipcServer_.isConnectable(event->xproperty.window)) {
-            root->ipcServer_.handleConnection(event->xproperty.window);
+            root->ipcServer_.handleConnection(event->xproperty.window, callCommand);
         } else if((client = get_client_from_window(ev->window))) {
             if (ev->atom == XA_WM_HINTS) {
                 client->update_wm_hints();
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -890,6 +892,17 @@ int main(int argc, char* argv[]) {
 
     init_handler_table();
     Commands::initialize(commands(root));
+    // now that we have initialized
```
dnnr: Wouldn't it simplify things a bit to turn this around and make the call handler a constructor parameter and move the current contents of the constructor into a `start()`/`startListening()`/`connect()` member that would be called after `Commands::initialize`?

That would remove the need for the call handler setter *and* for the default fallback handler with the HERBST_UNKNOWN_ERROR.

t-wissmann: I don't think that this makes things clearer. The design principle for me was: while Commands is not ready to be used yet, the ipc server should not have a handle on it. This is why the call to the setter is *right after* `Commands::initialize()`.

So I think it's conceptually bad to pass the callback to the IpcServer-constructor before the call of `Commands:initialize()`. (And this is regardless of the question whether the `IpcServer`-constructor already starts up the server or whether this is part of a `start()` function)

t-wissmann: If you want to avoid the setter or the default value, I'm fine with passing the `CallHandler` to every call of `addConnection()` and `handleConnection()`.

dnnr: > The design principle for me was: while Commands is not ready to be used yet, the ipc server should not have a handle on it.

That's a fair point. My thought about the current ctor comments was basically the same: Why start listening to calls immediately, if we *know* that we're not yet ready to handle calls anyway? 

I mean, the root case for all this seems to be the circular dependency Root->IpcServer->Commands->Root. Is it possible for the circle to be broken at the Root->IpcServer part (without just pointlessly shifting the issue elsewhere, of course)?

If not, maybe the best solution would be to do nothing at all in the IpcServer constructor and move both parts into explicitly called methods?

In any case, I'm just throwing around some ideas, I don't have any objections against your solution.

t-wissmann: Yes, the main issue is the relationship between Root, IpcServer, and Commands. We only need IpcServer in Root for `emit_hook`. But I wouldn't call it dependency, because in the present PR, IpcServer does not depend ony anything else (except XConnection...).

I'd like to emphasize again, that if we ove the server-bootup out of the IpcServer-constructor, then we don't solve any issues. Or saying it the other way round: booting up the entire server in the IpcServer constructor does not cause any issues at all.

Maybe I need to illustrate explicitly what I've meant with my last comment about `handleConnection()`.

dnnr: Ah, I think I see what you mean. The ctor code isn't the part that requires the callback to work. It's handleConnection that depends on it?

dnnr: Okay yeah, you're totally right about moving code out of the constructor is pointless.

But with regards to the circular dependency, am I getting this right that **if** `Root` didn't expect an `IpcServer*` as ctor argument, we could move the entire creation of IpcServer past the Commands::initialize() call and pass it the final callback handler as ctor parameter right there?

t-wissmann: Yes, absolutely right!

dnnr: And... do you think that would improve anything? At this point I'm not even sure myself.

t-wissmann: I've removed the member variable and the setter and added a parameter to handleConnection. Do you think, this makes the dependencies clearer? If not, I can revert the commit 5ffa988 again.

By the way: the design of handleConnection can be understood as a python-style context manager implemented using lambda functions.

t-wissmann: > And... do you think that would improve anything? At this point I'm not even sure myself.

I don't know, proprably not.

dnnr: I like the change! So now there is no chance of `Commands:call` being invoked prematurely because the very first potential calls of `createnotify` or `propertynotify` (the only two points-of-use of `callCommand`) are within the event loop, which comes way after `Commands::initialize`?

t-wissmann: Yes!

dnnr: Alright, let's do this. :-)

PR #473, opened 2019-05-05
Use a more plain stack for the monitor stack

t-wissmann:
For the monitor stack, we don't need all the layer handling. And
furthermore in a tag's stack, there can never be a slice of a monitor.
Hence, we remove all the monitor handling from Stack and Slice.

### DIFF
--- a/src/monitor.h
+++ b/src/monitor.h
@@ -10,7 +10,6 @@
 class HSTag;
 class MonitorManager;
 class Settings;
-class Slice;
 class Stack;
 
 class Monitor : public Object {
@@ -23,7 +22,6 @@ public:
 
     HSTag*      tag;    // currently viewed tag
     HSTag*      tag_previous;    // previously viewed tag
-    Slice* slice;  // slice in the monitor stack
     Attribute_<std::string>   name;
     Attribute_<unsigned long> index;
     DynAttribute_<std::string>   tag_string;
@@ -83,8 +81,6 @@ void ensure_monitors_are_available();
 void all_monitors_replace_previous_tag(HSTag* old, HSTag* newmon);
 
 void drop_enternotify_events();
-
-Stack* get_monitor_stack();
 
 void monitor_update_focus_objects();
 
--- a/src/monitor.h
+++ b/src/monitor.h
@@ -10,7 +10,6 @@
 class HSTag;
 class MonitorManager;
 class Settings;
-class Stack;
 
 class Monitor : public Object {
 public:
--- a/src/monitormanager.cpp
+++ b/src/monitormanager.cpp
@@ -11,7 +11,6 @@
 #include "ipc-protocol.h"
 #include "layout.h"
 #include "monitor.h"
-#include "plainstack.h"
 #include "settings.h"
 #include "stack.h"
 #include "tag.h"
@@ -28,7 +27,6 @@ MonitorManager* g_monitors;
 MonitorManager::MonitorManager()
     : IndexingObject<Monitor>()
     , focus(*this, "focus")
-    , monitorStack_(make_unique<PlainStack<Monitor*>>())
     , by_name_(*this)
 {
     cur_monitor = 0;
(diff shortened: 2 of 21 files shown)


t-wissmann:
In the next PR, we can replace a layer in a stack by `PlainStack<Slice*>`.

src/monitormanager.h
```diff
@@ -48,8 +49,11 @@ class MonitorManager : public IndexingObject<Monitor> {
     void unlock();
     std::string lock_number_changed();
 
+    void extractWindowStack(bool real_clients, std::function<void(Window)> addToStack);
+    void restack();
+
     int cur_monitor;
-    Stack* monitor_stack;
+    std::unique_ptr<PlainStack<Monitor*>> monitorStack_;
```
dnnr: Should this really keep being a public member, or are you planning on making it private later? (if it's the latter, I'd appreciate something like a `// TODO: make this private as soon as possible`

src/stack.h
```diff
@@ -23,11 +23,36 @@ extern const std::array<const char*, LAYER_COUNT> g_layer_names;
...
     SLICE_CLIENT,
     SLICE_WINDOW,
-    SLICE_MONITOR,
 } SliceType;
 
 class Client;
-class Monitor;
+
+template<typename T>
+class PlainStack {
```
dnnr: Does this need to be in here? I'd prefer it getting its own `plainstack.h`.

src/monitormanager.cpp
```diff
@@ -322,3 +324,24 @@ string MonitorManager::lock_number_changed() {
     return {};
 }
 
+//! return the stack of windows by successive calls to the given yield
+//function. The stack is returned from top to bottom, i.e. the topmost element
+//is the first element added to the stack.
```
dnnr: I'd say "is the first element yielded." here ;-)

src/monitormanager.h
t-wissmann: This should become private soon; I've added the comment.

src/stack.h
t-wissmann: I wasn't sure, but I think it's reasonable to move it. So yes, it's now in `plainstack.h`.

src/monitor.h
```diff
@@ -10,7 +10,6 @@
 class HSTag;
 class MonitorManager;
 class Settings;
-class Slice;
 class Stack;
```
dnnr: The `Stack` forward declaration can now be removed, too

src/stack.cpp
```diff
@@ -187,13 +180,6 @@ static void extractWindowsFromSlice(Slice* s, bool real_clients, HSLayer layer,
                 yield(s->data.window);
             }
             break;
-        case SLICE_MONITOR:
-            tag = s->data.monitor->tag;
```
dnnr: The `tag` local is now unused and can be removed

src/monitormanager.h
```diff
@@ -48,8 +49,11 @@ class MonitorManager : public IndexingObject<Monitor> {
     void unlock();
     std::string lock_number_changed();
 
+    void extractWindowStack(bool real_clients, std::function<void(Window)> addToStack);
+    void restack();
+
     int cur_monitor;
-    Stack* monitor_stack;
+    std::unique_ptr<PlainStack<Monitor*>> monitorStack_; // TODO: make this private
```
dnnr: The lifetime of `monitorStack_` exactly matches that of `MonitorManager`, so this could actually just be a plain old member instead of a unique_ptr.

I've made the change to confirm that it works, so if you agree with it, you can just cherry-pick this: https://github.com/dnnr/herbstluftwm/commit/c79b6626d589873cacc874ad5ca9861dce9775b9

t-wissmann: I've cherry-picked it, thanks! Originally, I wanted to avoid to include the rather long stack.h in monitormanager.h, but with the shorter plainstack.h, I think a member is better. Thanks!

PR #474, opened 2019-05-05
CI: Only run 'push' builds on master, winterbreeze

dnnr:


dnnr:
For PRs against winterbreeze, the branches are getting builds now as well. I'm assuming that Travis (unlike Mergify) respects the individual branch's .travis.yml, so making this change here too should help.

t-wissmann:
All right!

PR #475, opened 2019-05-05
Move stack command to MonitorManager

t-wissmann:
Also provide another constructor to CommandBinding for commands
that sit in objects and that don't need an Input parameter.

### DIFF
--- a/src/monitormanager.h
+++ b/src/monitormanager.h
@@ -49,6 +49,7 @@ public:
     void unlock();
     std::string lock_number_changed();
 
+    int printStack(Output output);
     void extractWindowStack(bool real_clients, std::function<void(Window)> addToStack);
     void restack();
 
--- a/src/monitormanager.h
+++ b/src/monitormanager.h
@@ -49,7 +49,7 @@ public:
     void unlock();
     std::string lock_number_changed();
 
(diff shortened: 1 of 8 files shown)


src/monitormanager.h
```diff
@@ -49,6 +49,7 @@ class MonitorManager : public IndexingObject<Monitor> {
     void unlock();
     std::string lock_number_changed();
 
+    int printStack(Output output);
```
dnnr: I don't think we've ever talked about it, but in many cases, we already use a `Command` suffix for methods like these. At times, I've really appreciated that clarification when reading the code, as it was more visible than noticing the fact that it has `Input` and/or `Output` as arguments. I think there is even a case where it's "necessary" in the sense that there is a non-command member `foo()` that does a thing and a `fooCommand()` that provides the command parsing interface for doing the same thing.

I'd suggest adopting that pattern for all command members.

t-wissmann: Sometimes I've tried to keep the name short where it's obvious that it's a command. I've renamed it to `stackCommand()` to make it obvious that this is the implementation of the command `stack`.

I agree with your suggestion to have the `…Command` suffix for all command member functions.

PR #476, opened 2019-05-06
Fix keyboard focus

t-wissmann:
The focus setting helper frame_focus_recursive() was called in many
places (and the focus was still broken when the first client showed up).
However, every such call is followed by a call to Monitor::applyLayout()
which sets the object link clients.focus. If the window is not visible
yet, XSetInputFocus() has no effect. So we need to reset the focus on
MapNotify events.

Hence, we now directly set the keyboard focus at this single place in
Monitor::applyLayout().

### DIFF
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -162,7 +162,6 @@ int FrameTree::removeFrameCommand() {
         // if parent was root frame
         root_ = newparent;
     }
-    frame_focus_recursive(parent);
     get_current_monitor()->applyLayout();
     return 0;
 }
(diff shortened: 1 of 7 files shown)


t-wissmann:
Thanks! I've just started using the winterbreeze branch on my machine in the office. Let us see what happens :-)

blarz:
Nice, thank you for this fix!
I have been using winterbreeze on my laptop for a couple of months now and I had the problem that opening a new browser window left the focus in the old window sometimes (should've opened an issue, I know...).

This PR fixes the problem for me.

t-wissmann:
Oh wow :-) I thought that winterbreeze was unusable before this fix. Did you notice any segfaults? I've noticed #482 yesterday.

blarz:
I did not experience any segfaults or ASAN triggers so far (running a Debug binary here).
As I do not have an external monitor connected to my laptop, I did not run into that bug you mentioned.

winterbreeze runs quite stable here, no problems so far (besides the focus bug, but that didn't really bother me that much).

PR #477, opened 2019-05-06
Focus dummy window if no client is focused

t-wissmann:
Previously, if no window is focused, the root windows was focused. This
had the effect that the keyboard input was sent to the window under the
mouse cursor (even though this windows was not focused).

We now create a dummy window outside of the screen. This window is used
already for EWMH's _NET_SUPPORTING_WM_CHECK, and now this window is
focused if no real client is focused.

Additionally, this fixes the analogue issue with focus_follows_mouse
which did not work previously when no client window is focused.


t-wissmann:
This includes #476.

PR #478, opened 2019-05-06
Focus dummy window if no client is focused

t-wissmann:
This is an adaption of a commit in the winterbreeze branch with the same
title.

Previously, if no window is focused, the root windows was focused. This
had the effect that the keyboard input was sent to the window under the
mouse cursor (even though this windows was not focused).

We now create a dummy window outside of the screen. This window is used
already for EWMH's _NET_SUPPORTING_WM_CHECK, and now this window is
focused if no real client is focused.

Additionally, this fixes the analogue issue with focus_follows_mouse
which did not work previously when not client window is focused.

### DIFF
--- a/src/clientlist.cpp
+++ b/src/clientlist.cpp
@@ -462,8 +462,7 @@ void client_window_unfocus_last() {
         client_window_unfocus(lastfocus);
     }
     hsobject_unlink_by_name(g_client_object, "focus");
-    // give focus to root window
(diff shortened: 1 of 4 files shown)


t-wissmann:
This is PR #477 adapted to the old c-code in master.

t-wissmann:
I don't have time to test it in practise because I prefer to make the winterbreeze-code base stable. So I'm discarding this PR... (For the same reason, I'll not adapt #480 to the old code in master)

PR #480, opened 2019-05-09
Check in herbstclient if herbstluftwm is running

t-wissmann:
Before sending commands or listening for hooks, check that herbstluftwm
is running. This is done by reading the window id of herbstluftwm's hook
window and then checking that this window still exists.

In the test suite, it is checked that herbstclient recognizes an absent
herbstluftwm. Previously, sending commands without herbstluftwm made
herbstclient hang.

In this new test case, we either don't start hlwm or we make it quit
abnormally, so we don't want the hlwm_process fixture to start and check
that hlwm shuts down gently. Hence, this removes the autouse-flag from
the hlwm_process fixture. This does not affect the other taste cases,
because they actively rely on hlwm which depends on hlwm_process.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -464,6 +464,8 @@ def kill_all_existing_windows(show_warnings=True):
 @pytest.fixture()
 def hlwm_spawner(tmpdir):
     """yield a function to spawn hlwm"""
+    assert os.environ['DISPLAY'] != ':0', 'Refusing to run tests on display that might be your actual X server (not Xvfb)'
+
     def spawn():
         env = {
             'DISPLAY': os.environ['DISPLAY'],
@@ -475,7 +477,7 @@ def hlwm_spawner(tmpdir):
 
 @pytest.fixture()
 def hlwm_process(hlwm_spawner):
-    assert os.environ['DISPLAY'] != ':0', 'Refusing to run tests on display that might be your actual X server (not Xvfb)'
+    """Set up hlwm and also check that it shuts down gently afterwards"""
     hlwm_proc = hlwm_spawner()
     kill_all_existing_windows(show_warnings=True)
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -242,7 +242,7 @@ class HlwmBridge:
         return "true" if python_bool_var else "false"
 
 
-@pytest.fixture
+@pytest.fixture()
 def hlwm(hlwm_process):
     display = os.environ['DISPLAY']
     # display = ':13'
@@ -432,7 +432,7 @@ class HcIdle:
             self.proc.wait(2)
 
 
-@pytest.fixture
+@pytest.fixture()
 def hc_idle(hlwm):
     hc = HcIdle(hlwm)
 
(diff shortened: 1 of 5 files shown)


t-wissmann:
Once this is merged, I'll re-do the changes in master. Together with #478, these will be the last changes in the old code base for the last non-winterbreeze release of herbstluftwm.

tests/conftest.py
```diff
@@ -461,7 +461,19 @@ def kill_all_existing_windows(show_warnings=True):
         subprocess.run(['xdotool', 'windowkill', c])
 
 
-@pytest.fixture(autouse=True)
+@pytest.fixture()
+def hlwm_spawner(tmpdir):
```
dnnr: Why not make use of `hlwm_spawner` in `hlwm_process`? Putting together the environment and creating the HlwmProcess is basically the same code in both fixtures now, isn't it?

t-wissmann: Bascially yes, I suspected that it makes `hlwm_process` more nested&complicated to read but not really shorter. I only wrote `hlwm_spawner` because I didn't know how to access the constructor `HlwmProcess()` from test cases directly.

I've made this change in another branch: https://github.com/herbstluftwm/herbstluftwm/commit/ed171c854076b32c1b5430cc250bef3033a9bccb Do you think this change is an improvement. I'm really not sure.

dnnr: I think it's pretty logical. It's making use of the building blocks that we provide.

Would there be any reason not to move the assert into hlwm_spawner as well?

tests/test_herbstclient.py
```diff
@@ -0,0 +1,43 @@
...
+def test_herbstclient_recognizes_hlwm_not_running(hlwm_spawner, x11, hlwm_mode, hc_parameter):
+    if hlwm_mode == 'never started':
+        pass
+    else:
+        # start hlwm and kill it in order to leave X in some inconsistent state
+        hlwm_proc = hlwm_spawner()
+        # ping that herbstclient still works here
+        hc = subprocess.run([HC_PATH, 'echo', 'ping'],
+                            stdout=subprocess.PIPE,
+                            text=True,
```
dnnr: Given my previous stance on it, I'm not really in a position to argue for supporting older Python versions, but note that using `text=True` here is the only part the would make us incompatible with Python 3.6 (i.e., require 3.7). `text=True` is a convenience alias for `universal_newlines=True`, so we *could* use that to stay compatible with 3.6 for a while longer.

tests/conftest.py
t-wissmann: The assert is only there to avoid that clients of an actual session are killed. But we also can assert this for hlwm_spawner, it does not really matter, I think.

So I made all these changes in the present PR

tests/test_herbstclient.py
t-wissmann: I didn't know this and thus changed it to `universal_newlines=True`.

dnnr:
I took the liberty to fix the other place where `text=` was used

t-wissmann:
Thanks for the `text=`-fix!

PR #481, opened 2019-05-09
Keep X11 connection open for entire test run

dnnr:


tests/conftest.py
```diff
@@ -488,13 +490,32 @@ def running_clients(hlwm, running_clients_num):
...
+    class X11Connection:
+        def __init__(self):
+            self.display = Xlib.display.Display()
+
+        def shutdown(self):
+            self.display.close()
+
+    x11_connection_ = X11Connection()
+    yield x11_connection_
+    x11_connection_.shutdown()
```
t-wissmann: Here, we don't really need the class, right? We could just define:
```
display = Xlib.display.Display()
yield display
display.close()
```
But we can also keep the class for the case that we want to put more attributes here.

dnnr: Oh, you're totally right! I was quite stuck in the pattern it seems... I'll change it

tests/conftest.py
```diff
@@ -488,13 +490,25 @@ def running_clients(hlwm, running_clients_num):
...
 
+@pytest.fixture()
+def x11(x11_connection):
+    """ Short-lived fixture for interacting with the X11 display and creating
+    clients that are automatically destroyed at the end of each test. """
     class X11:
-        def __init__(self):
-            self.display = display.Display()
+        def __init__(self, x11_connection):
+            self.display = x11_connection.display
```
t-wissmann: This now becomes `self.display = x11_connection`

dnnr: Oops, thanks for fixing it

PR #485, opened 2019-05-10
Use PlainStack in the Stack class

t-wissmann:
At the same time, call the member variable top_. This simplifies the
code in stack.cpp.

### DIFF
--- a/src/monitormanager.cpp
+++ b/src/monitormanager.cpp
@@ -377,7 +377,7 @@ int MonitorManager::stackCommand(Output output) {
     for (Monitor* monitor : monitorStack_) {
         vector<shared_ptr<StringTree>> layers;
         for (size_t layerIdx = 0; layerIdx < LAYER_COUNT; layerIdx++) {
-            auto layer = monitor->tag->stack->top_[layerIdx];
+            auto layer = monitor->tag->stack->slices_[layerIdx];
 
             vector<shared_ptr<StringTree>> slices;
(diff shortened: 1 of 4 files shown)


src/monitormanager.cpp
```diff
@@ -377,7 +377,7 @@ int MonitorManager::stackCommand(Output output) {
     for (Monitor* monitor : monitorStack_) {
         vector<shared_ptr<StringTree>> layers;
         for (size_t layerIdx = 0; layerIdx < LAYER_COUNT; layerIdx++) {
-            auto layer = monitor->tag->stack->top[layerIdx];
+            auto layer = monitor->tag->stack->top_[layerIdx];
```
dnnr: Maybe this is an opportunity for a different name altogether? I think we even discussed this some time ago... iirc, the reason that it is called `top` is because it used to be a pointer to the first entry of a linked list?

How about calling it `slices_` or `layers_`?

t-wissmann: I remember the discussion but I can't remember any conclusion. But `slices_` and `layers_` are fine for me. Maybe let's go for `slices_`?

dnnr: I don't think we came to any conclusion other than that we should rename it at some point. ;-) I'm fine with both, but `slices_` seems slightly more logical.

t-wissmann: I've renamed it.

PR #486, opened 2019-05-11
Migrate raise_monitor command

t-wissmann:
Move the implementation raise_monitor command to MonitorManager. Also
implement the completion using the Completion class. The old
implementation complete_against_monitors() was not deleted since it is
used for the completion of other commands.

Also add test cases for raise_monitor and its completion.


src/monitormanager.h
```diff
@@ -31,6 +32,7 @@ class MonitorManager : public IndexingObject<Monitor> {
     int list_monitors(Output output);
     int list_padding(Input input, Output output);
     int string_to_monitor_index(std::string string);
+    void completeMonitorName(Completion& complete);
```
dnnr: This looks and feels like an internal helper. Why not make it private?

t-wissmann: It should have the same visibility as `string_to_monitor_index` because every command that is allowed to take a monitor as an argument should be able to call `completeMonitorName`. (That's why the other completion helpers in other classes are (and should stay) public too, e.g. `RootCommands::completeAttributePath`)

dnnr: Okay, makes sense! :+1:

PR #489, opened 2019-05-12
 Clean up slice member functions and visibility

t-wissmann:
Remove the unused functions slice_destroy() and slice_highest_layer()
(the latter is unused in master as well). Also make
extractWindowsFromSlice() a member function of Slice. This allows us to
make nearly all member variables of Slice private.


t-wissmann:
This includes #485.

PR #490, opened 2019-05-12
 Make SliceType an enum class Slice::Type

t-wissmann:
The members of this type class are named ClientSlice and
WindowSlice because the names without the suffix lead to shadowing
clashes (-Werror=shadow).


t-wissmann:
This includes #489 and #485.

PR #499, opened 2019-05-19
Catch exceptions by stoi() and stoul() calls

t-wissmann:
This fixes the crash when non-integer values are passed as the
respective arguments.


t-wissmann:
Since this only fixes crashes, I'm merging this directly.

PR #506, opened 2019-05-24
Check in herbstclient if herbstluftwm is running

t-wissmann:
This adapts ad8b461b6a6c0b7ef14fe6a8d55a10dd9a35652f (#480) to the old
herbstclient code in master


t-wissmann:
This herbstclient passes all tests in the winterbreeze branch, so I'd say it's save to be used.

PR #511, opened 2019-05-26
Read monitor list from xrandr

t-wissmann:
Add monitor detection via xrandr. This also ads a cmake option to
disable this feature at compile time.

There is a separate xrandr event (see man 3 xrandr), which we don't ask
for at the moment, since listening to resolution changes (Root window
gets reconfigured) seems to be enough for the moment.


dnnr:
What's the reason for making this compile-time configurable? I'd prefer if we didn't enter ifdef hell unless we really need to.

t-wissmann:
it's another compile time dependency not everybody needs.

Related: somebody joined the irc and asked for xrandr support such that he can build hlwm without xinerama support to make the system xinerama-free.

And I can imagine that there are people that do it the other way round. I don't see the ifdef hell here.

dnnr:
Any ifdef is ifdef hell, imho. :wink: But if there's demand for it, so be it. :+1: Is there something fishy about xinerama or xrandr that I'm not aware of? I'd just disable it if I didn't want to use it.

t-wissmann:
I've made xrandr mandatory.

dnnr:
Looks good to me, but I haven't tested it. Do you want me to?

blarz:
Does the INSTALL file need an update, too? Or how specific is the list of dependencies?

t-wissmann:
> Looks good to me, but I haven't tested it. Do you want me to?

I didn't know which corner cases to test, any ideas are welcome. Do you have any ideas how to test it in the testsuite?

> Does the INSTALL file need an update, too? Or how specific is the list of dependencies?

Right, we should update INSTALL as well!

dnnr:
I got no idea... I assumed that manual testing is all we can do here.

t-wissmann:
If you can test it manually, that would be great!

t-wissmann:
It seems to work here, so let's merge it!

PR #516, opened 2019-05-31
Merge winterbreeze with master

t-wissmann:
This merges master with winterbreeze, such that at any point, we can
simply fast-forward merge winterbreeze into master.


t-wissmann:
Once all the tests pass, I will merge this manually from the command line.

PR #539, opened 2019-06-09
Use Converter for MouseCombo io

t-wissmann:
Implement Converter<MouseCombo> and use it for parsing and completion of
MouseCombo values.


t-wissmann:
This includes #538

PR #553, opened 2019-06-28

bmwiedemann:
cmake already uses `SOURCE_DATE_EPOCH` for its `TIMESTAMP` function.

Makefile could use it as well as documented in

Note: cmake codepath needs testing. I only tested the Makefile part.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ tar: doc
 
 doc/%.1 doc/%.7: doc/%.txt version.mk
 	$(call colorecho,DOC,$@)
-	$(VERBOSE) $(A2X) -f manpage -a "herbstluftwmversion=herbstluftwm $(VERSION)" -a "date=`date +%Y-%m-%d`" $<
+	$(VERBOSE) $(A2X) -f manpage -a "herbstluftwmversion=herbstluftwm $(VERSION)" -a "date=`date -u -r NEWS %Y-%m-%d`" $<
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -93,7 +93,7 @@ tar: doc
 
 doc/%.1 doc/%.7: doc/%.txt version.mk
 	$(call colorecho,DOC,$@)
-	$(VERBOSE) $(A2X) -f manpage -a "herbstluftwmversion=herbstluftwm $(VERSION)" -a "date=`date +%Y-%m-%d`" $<
+	$(VERBOSE) $(A2X) -f manpage -a "herbstluftwmversion=herbstluftwm $(VERSION)" -a "date=`date -u -r NEWS %Y-%m-%d`" $<
```
t-wissmann: When using a file's modification date, I think one should use `$<` because the modification date of `NEWS` is independent from the man page modification date.

t-wissmann: I'm wrong for released versions. I think it is fine as it is!

t-wissmann:
Your changes improve the build system, and I don't want to discard your contribution. However, the development in the master branch is paused in favor of the winterbreeze branch, and the entire content of master will be discarded soon. The winterbreeze branch is "ahead of master" (in the git language), so I'd like not to commit to the master branch (except for bug fixes).

You are welcome to do these changes in the winterbreeze branch (the resepctive file is `doc/CMakeLists.txt`; the entire build system is cmake based now).

Issue #554, opened 2019-06-29
Winterbreeze: Pseudotile rule does not apply to windows with multiple window types


olmokramer:
I'm trying out winterbreeze (commit cf7a170c) and most things work just as they did before. The rules I used pre-winterbreeze to make the gpg-agent pinentry window pseudotiled automatically do not work anymore, however. Any one of these rules worked pre-winterbreeze but does not get applied in winterbreeze:

```
herbstclient rule windowtype~'_NET_WM_WINDOW_TYPE_(DIALOG|UTILITY|SPLASH) focus=on pseudotile=on

herbstclient rule class~'[Pp]inentry' focus=on pseudotile=on
```

The former still does seem to work for some of the windows it should target, such as GIMP's splash window.

Following rule also still works (you can open it via the menu, 'File' -> 'Document Properties...'):

```
herbstclient rule class~'[Ii]nkscape' title~'Document Properties.*' focus=on pseudotile=on
```

olmokramer:
This also appears to happen to qutebrowsers "open file" dialog that opens, for example, when clicking the "Attach files ..." text below these GitHub textboxes. When I noticed this I also tried Chromium's "open file" dialog, and it _does_ get the pseudotile applied.

Upon investigating, I think `xprop` may have revealed (part of) the issue: both the pinentry and qb's "open file" dialog have

```
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG, _NET_WM_WINDOW_TYPE_NORMAL
```

while Chromium's "open file" dialog had

```
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG
```

I tried changing the rule to read like below, but none of the variations worked:

```
windowtype~'_NET_WM_WINDOW_TYPE_(DIALOG|UTILITY|SPLASH).*'

windowtype~'.*_NET_WM_WINDOW_TYPE_(DIALOG|UTILITY|SPLASH).*'

windowtype~'_NET_WM_WINDOW_TYPE_(DIALOG|UTILITY|SPLASH), _NET_WM_WINDOW_TYPE_NORMAL'
```

This does not explain why the `class~'[Pp]inentry'` rule did not work, though, unless the problem is that pseudotiling windows with multiple window types is broken, which appears not to be the case, as `herbstclient pseudotile on` from the terminal does work.

t-wissmann:
[...]

PR #555, opened 2019-06-30
Use first window type in rules if there are multiple

t-wissmann:
If a window has multiple window types, then use the first window type
entry. This establishes the behaviour of the 'windowtype' rule condition
from the old herbstluftwm codebase.

This should fix issue #554.


t-wissmann:
This fix definitely makes the winterbreeze behaviour more similar to the old herbstluftwm codebase ("master"). I assume that this fixes #554, but even if this PR does not fix this particular issue, I think this PR is worth merging.

PR #561, opened 2019-07-08
Run herbstluftwm with valgrind in test suite

t-wissmann:
Run the test suite with valgrind and let the test suite fail if there
are any errors reported by valgrind.


t-wissmann:
This will fail anyway until #559 is merged.

PR #563, opened 2019-07-08

bmwiedemann:
See https://reproducible-builds.org/ for why this is good

CMake already uses `SOURCE_DATE_EPOCH` for `TIMESTAMP`
The `SOURCE_DATE_EPOCH` variable is documented in

Also use UTC to be independent of timezone.

Note: needs testing


t-wissmann:
Great, thanks!

t-wissmann:
I've temporarily reverted this commit again, because this leads to build failures here: 68199bd

bmwiedemann:
Maybe the `STRING(TIMESTAMP` needs to move further up? needs testing.

PR #584, opened 2019-10-18
Man page: fix typo in conditions

dario23:


t-wissmann:
Thanks for the fix! Right now, I only accept PRs for bug fixes & crashes for master. So I have cherry-picked it to winterbreeze.

PR #588, opened 2019-11-04
Print `~` for regex rules instead of `=`

xaverh:
for consistency with the `rule` command


t-wissmann:
I want only critical bug fixes in master for the moment, and for winterbreeze (the next master), your proposed change has already been implemented in 56101aba0b9b5e8bd282bf308f024e5f14253988. Thank you anyway for your contribution!

PR #600, opened 2020-01-18
Completion update

phy1729:
Few completion updates.
* zsh completers are intended to be autoloaded by compinit rather than sourced.
* Third party completers should be installed into site-functions.

I also switched _herbstclient to use `${(f)...}` parameter expansion flag as it's more idiomatic zsh. I don't use herbstluftwm, so I would appreciate if someone could double check completion still works.


t-wissmann:
Thanks for your contribution. Can you please rebase your pull request to the `winterbreeze` branch? This is where the development currently takes place, and for the moment, there are only critical bug fixes that are merged to `master`.

phy1729:
Rebased.

I can't test the patch because I get an unrelated build error on OpenBSD.
```
[ 19%] Building CXX object src/CMakeFiles/herbstluftwm.dir/completion.cpp.o
/home/phy1729/herbstluftwm/src/completion.cpp:38:33: error: invalid operands
      to binary expression
      ('basic_ostream<char, std::__1::char_traits<char> >' and 'const char *')
        output_ << escape(word) << (shellOutput_ ? " \n" : "\n");
        ~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/memory:5393:1: note: candidate template ignored: could not
      match 'shared_ptr<type-parameter-0-2>' against 'const char *'
operator<<(basic_ostream<_CharT, _Traits>& __os, shared_ptr<_Yp> const& __p);
^
/usr/include/c++/v1/string:4249:1: note: candidate template ignored: could not
      match 'basic_string<type-parameter-0-0, type-parameter-0-1,
      type-parameter-0-2>' against 'const char *'
operator<<(basic_ostream<_CharT, _Traits>& __os,
^
/home/phy1729/herbstluftwm/src/completion.cpp:75:33: error: invalid operands
      to binary expression
      ('basic_ostream<char, std::__1::char_traits<char> >' and
      'const char [2]')
        output_ << escape(word) << "\n";
        ~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~
/usr/include/c++/v1/memory:5393:1: note: candidate template ignored: could not
      match 'shared_ptr<type-parameter-0-2>' against 'char const[2]'
operator<<(basic_ostream<_CharT, _Traits>& __os, shared_ptr<_Yp> const& __p);
^
/usr/include/c++/v1/string:4249:1: note: candidate template ignored: could not
      match 'basic_string<type-parameter-0-0, type-parameter-0-1,
      type-parameter-0-2>' against 'char const[2]'
operator<<(basic_ostream<_CharT, _Traits>& __os,
^
2 errors generated.
```

t-wissmann:
The completion seems to work here, thanks for the contribution!

> * Use DATADIR in FISHCOMPLETIONDIR so systems that use PREFIX = /usr/local get the fish completion installed into the right spot.

I don't get this comment, because your change oes not metion FISHCOMPLETIONDIR at all.

phy1729:
Looks like that commit was lost in the rebase from master to winterbreeze (as winterbreeze doesn't have a config.mk) and I neglected to update the description; sorry about that.

t-wissmann:
I see. So as it turns out I still need to handle the installation of the  fish completion in the winterbreeze branch. Thanks for the hint.

Issue #576, opened 2019-08-28
Positioning windows on configure requests should be monitor-relative


dnnr:
It seems that this was fixed in master (https://github.com/herbstluftwm/herbstluftwm/commit/ebed05c0a3faeed9084bb8355532b4f5cfbf2ca6), but still occurs in winterbreeze.

preisi:
I have ported the respective commit to my hlwm fork. However, I would like to verify its correctness before sending a pull request. 
Is there any way to easily test the behaviour of my changes?

t-wissmann:
Great, thanks!

You can run
```
xterm -geometry 50x10+2000+0
```
where the 2000 is some X-coordinate on a monitor on the right. And then the xterm should appear at a reasonable coordinate and not stick on the right edge of the monitor.

t-wissmann:
If you have an application that remembers its coordinates, then you can open and close and open it again. For every monitor, the application should re-appear at roughly the same coordinate (and in particular should not open at the right edge of the monitor).

For example, sonata is an application that saves and restores the window coordinates.

t-wissmann:
The above mentioned commit is not the only commit that implements the geometry-fixing. Most of the work is done by 9e7538f83eccfe0a5cad1bfd7da82e37e1590a76.

t-wissmann:
As it turns out I was wrong with all my comments above:
This Issue (and the commit ebed05c) is about re-interpreting the coordinates on a configure request of an existing client.

My comments (both about the `xterm`-command and the referenced 9e7538f) are about re-interpreting the requested coordinates of a *new* client window when it shows up for the first time. (and the latter issue is now implemented by #614 )

t-wissmann:
Solved by #615.

PR #601, opened 2020-01-23
[WIP][winterbreeze] Make position monitor-relative on configure request

preisi:
A client's floating window position is relative to the monitor the
client is on. However, the coordinates of a configure request is
relative to the root window. So convert the coordinates. If the client
is not visible, we fuzzily find the most appropriate monitor based on
the coordinates of the client.

Ported from ebed05c.

The code compiles but it does not yet show the expected behaviour -> WIP


t-wissmann:
For documentation purposes: this PR targets the issue https://github.com/herbstluftwm/herbstluftwm/issues/576

t-wissmann:
I'm taking a look at it right now.

t-wissmann:
I've adapted the commit to the winterbreeze sourcecode in #615 (together with a test case).

Thanks for taking a look at it!

PR #604, opened 2020-02-25
Make getInnerNeighbourIndex() a member of HSFrameLeaf

t-wissmann:


t-wissmann:
Yes :-) This really needs to be finished...

PR #605, opened 2020-02-26
New command: export

t-wissmann:
Add a new command 'export' which is a convenience wrapper around 'setenv'.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -26,6 +26,7 @@ Current git version
     dialog (e.g. a rofi menu or a notification) closes.
   * monitor detection via xrandr
   * list_rules now prints regex-based rule conditions with '~' instead of '='
+  * new command: export (convenience wrapper around setenv)
 
 Release 0.7.2 on 2019-05-28
 ---------------------------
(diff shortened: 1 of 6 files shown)


t-wissmann:
This command did not exist in master before, but I find it extremely useful  and reasonable and thus implemented it as a warm-up exercise to get used to the codebase and C++ again :-)

dnnr:
Maybe I'm missing something, but how is different from setenv? Does it add anything, or is it just supposed to be a convenience *alias*?

t-wissmann:
> Maybe I'm missing something, but how is it different from setenv? Does it add anything, or is it just supposed to be a convenience _alias_?

The difference is the `=` syntax:
```
herbstclient setenv FOO BAR
herbstclient export FOO=BAR
```
So it's the usual `export` syntax, just with `herbstclient` in front of it.

dnnr:
I see. How about making it more explicit in the man page that one is really just an alias of the other (with a different separator), and then explain the actual functionality in just *one* of the commands, exclusively?

From a user's point of view, I think I'd be quite confused to see two options that seem identical yet are described differently. By each command "pretending" that the other one doesn't exist, it suggests some orthogonality that is not there.

t-wissmann:
I thought I was doing this already. Now I've updated the description to make it really obvious that they're the same with different syntax.

dnnr:
Alright, I think that's a bit better. 👍

Issue #520, opened 2019-06-01
Adapt from master: mouse resize in tiling mode


t-wissmann:
Adapt ed2cd33769e19cc5ac03b88e353a149c94db971f to winterbreeze.

PR #606, opened 2020-03-01
Encapsulate mouse drag handling to own class

t-wissmann:
Encapsulate all the details what to do with mouse movemeng in a new
class MouseDragHandler. The responsibilities are now:

  - MouseManager has to get the events and mouse coordinates from X and
    pass them over to MouseDragHandler.
  - MouseDragHandler has to interpret mouse coordinate updates
    appropriately.
  - MouseDragHandler has to check whether the resources that are
    manipulated still exist and throw an DragNotPossible exception
    otherwise (these 'resources' are essentially the member variables of
    MouseDragHandler).

With this interface defined, MouseManager does not need the 'mode'
member anymore because the mode is implicit in a single other member
variable dragHandler_.

The benefit is that we can now cleanly implement mouse dragging features
for the tiling mode that may require other resources.


t-wissmann:
I'm a bit lost: IWYU demands that
```
/hlwm/src/mousemanager.h should remove these lines:
- #include "x11-types.h"  // lines 10-10
[...]
IWYU made suggestions to remove things in (see log above): /hlwm/src/mousemanager.h
```
However, `x11-types.h` defines `Point2D`, which is needed in `mousemanager.h`. Hence, if I remove the `#include` (i.e. if I undo c290b59), then the compilation fails. Is there anything obvious I am missing @dnnr or is the suggestions by IWYU simply wrong?

t-wissmann:
`Point2D` was only used in function declarations (not definitions), and I didn't know that in function declarations, a forwars declaration of the types involved are sufficient. So it's all fine now :-)

src/mousedraghandler.cpp
```diff
@@ -0,0 +1,173 @@
+#include "client.h"
+#include "monitormanager.h"
+#include "mouse.h"
+#include "mousedraghandler.h"
+#include "tag.h"
+#include "x11-utils.h"
+
+MouseDragHandler::MouseDragHandler(ClientManager*, MonitorManager*  monitors, Client* dragClient, MouseDragFunction function)
```
dnnr: Did you leave out the parameter name here on purpose? It seems a bit inconsistent (cf. MonitorManager).

There's also some double whitespace (in the header file as well)

src/mousemanager.h
```diff
@@ -41,31 +37,20 @@ class MouseManager : public Object {
...
     void mouse_initiate_zoom(Client* client, const std::vector<std::string> &cmd);
     void mouse_initiate_resize(Client* client, const std::vector<std::string> &cmd);
     void mouse_call_command(Client* client, const std::vector<std::string> &cmd);
-    /* some mouse drag functions */
-    void mouse_function_move(Point2D newCursorPos);
-    void mouse_function_resize(Point2D newCursorPos);
-    void mouse_function_zoom(Point2D newCursorPos);
 
 private:
     //! check whether we can continue dragging
```
dnnr: Is this comment still appropriate? Maybe just leave it out

dnnr: Ah you somehow made this class represent the boolean of dragging possibility? I guess I have to take closer look first...

src/mousedraghandler.cpp
t-wissmann: I don't need `ClientManager*` here, but I thought that I might need it at some point.. Since it's not much work, I've removed the unused `ClientManager*` from the constructor.

src/mousemanager.h
t-wissmann: This comment is obsolete now! Sorry for the confusion.

The MouseDragHandler class now also takes care of this boolean of dragging possibility. But this class does more than that, namely the actual window geometry changes during the drag.

src/mousedraghandler.cpp
```diff
@@ -0,0 +1,173 @@
...
+    if (monitors_->byIdx(dragMonitorIndex_) != dragMonitor_
+            || !dragMonitor_
+            || !winDragClient_
+            || winDragClient_->tag()->floating == false)
+    {
+        throw DragNotPossible();
+    }
+}
+
+/** finalize the what was started by the drag, do more than what is
```
dnnr: Extraneous "the"?

src/mousemanager.cpp
```diff
@@ -186,20 +171,18 @@ void MouseManager::mouse_stop_drag() {
 }
 
 void MouseManager::handle_motion_event(Point2D newCursorPos) {
-    if (!draggingIsStillSafe()) {
-        mouse_stop_drag();
+    if (!dragHandler_) {
```
dnnr: Would it make sense, semantically, to call `mouse_is_dragging` here (and in similar situations)?

t-wissmann: No, it would not make sense in my opinion, because the check here is less semantically, but more technically: it ensures that `dragHandler_` is not null when dereferencing it in the next line:
```
dragHandler_->handle_motion_event(newCursorPos);
```

src/mousedraghandler.cpp
t-wissmann: fixed.

t-wissmann: fixed.

t-wissmann:
I think it's OK as it is for the moment. I will need to modify the drag handler anyway when implementing #520.

PR #608, opened 2020-03-08
CI: Restrict IWYU output to remove suggestions

t-wissmann:


t-wissmann:
Finally, the output seems to be correct:
```
IWYU made suggestions to remove things in the following files:
===== /hlwm/src/mousemanager.h should remove: =====
- #include "decoration.h"  // lines 6-6

===== /hlwm/src/client.h should remove: =====
- #include "decoration.h"  // lines 8-8
- #include "ewmh.h"  // lines 9-9

===== /hlwm/src/tagmanager.h should remove: =====
- class Decoration;  // lines 10-10

After removing the above lines it might be necessary to add
additional forward declarations to make it build again.
```

dnnr:
Looks good to me! The include changes were only for testing, right?

t-wissmann:
It seems that one of the test cases is unstable:
```
tests/test_clients.py::test_urgent_on_start[True] 
```

PR #613, opened 2020-03-15
Fix y-coordinate of floating clients

t-wissmann:

PR #614, opened 2020-03-15
Make initial client positions monitor relative

t-wissmann:
When a new client window suggests its window size, it does so by
proposing (x,y)-coordinates relative to (0,0) of the root window. We
now detect within which monitor the new window lies and transform the
coordinates to be relative to that monitor.

If the user puts the clinet window on the monitor within the window's
requested coordinates are, then the client window is precisely at its
requested coordinates. The new test case checks this (both for one and
two monitors).


t-wissmann:
The test case `test_initial_client_position` will succeed as soon as #613 is merged.

Issue #609, opened 2020-03-08
Unstable test case for urgency


t-wissmann:
One of the urgency test cases is a bit unstable. It failed already twice today:
```
tests/test_clients.py::test_first_client_gets_focus 
tests/test_clients.py::test_client_lives_longer_than_hlwm 
[gw0] [  0%] PASSED tests/test_clients.py::test_first_client_gets_focus 
[gw1] [  0%] PASSED tests/test_clients.py::test_client_lives_longer_than_hlwm 
tests/test_clients.py::test_fullscreen_completion 
tests/test_clients.py::test_alter_fullscreen 
[gw0] [  0%] PASSED tests/test_clients.py::test_fullscreen_completion 
tests/test_clients.py::test_close[1] 
[gw0] [  0%] PASSED tests/test_clients.py::test_close[1] 
[gw1] [  0%] PASSED tests/test_clients.py::test_alter_fullscreen 
tests/test_clients.py::test_close_without_clients 
tests/test_clients.py::test_close[3] 
[gw1] [  1%] PASSED tests/test_clients.py::test_close_without_clients 
tests/test_clients.py::test_close[2] 
[gw0] [  1%] PASSED tests/test_clients.py::test_close[3] 
[gw1] [  1%] PASSED tests/test_clients.py::test_close[2] 
tests/test_clients.py::test_close[4] 
tests/test_clients.py::test_urgent_on_start[True] 
[gw0] [  1%] ERROR tests/test_clients.py::test_urgent_on_start[True] 
==================================== ERRORS ====================================
_________________ ERROR at setup of test_urgent_on_start[True] _________________
[gw0] linux -- Python 3.7.3 /hlwm/.tox/py37/bin/python
self = <Xlib.display._BaseDisplay object at 0x7ff2bb604908>, flush = None
event = None, request = -1, recv = None
    def send_and_recv(self, flush = None, event = None, request = None, recv = None):
        """send_and_recv(flush = None, event = None, request = None, recv = None)
    
        Perform I/O, or wait for some other thread to do it for us.
    
        send_recv_lock MUST be LOCKED when send_and_recv is called.
        It will be UNLOCKED at return.
    
        Exactly or one of the parameters flush, event, request and recv must
        be set to control the return condition.
    
        To attempt to send all requests in the queue, flush should
        be true.  Will return immediately if another thread is
        already doing send_and_recv.
    
        To wait for an event to be received, event should be true.
    
[...]

PR #628, opened 2020-03-21
Work around the TypeError in Xlib.display.Display()

t-wissmann:
While https://github.com/python-xlib/python-xlib/pull/160 is not merged,
try to avoid issue #609 by attempting to open the display multiple times
as long as no TypeError occurs anymore.


t-wissmann:
Of course, one successful build doesn't tell us anything about whether it worked, so let us see where this goes.

PR #648, opened 2020-03-23
Fix Debug build with gcc older than version 8

blarz:
-Wcast-function-type was introduced in gcc8. Make a debug build possible
for older versions by only enabling (in this case disabling...) the
warning only for certain gcc versions.

### DIFF
--- a/cmake/debugflags.cmake
+++ b/cmake/debugflags.cmake
@@ -24,10 +24,14 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     list(APPEND FLAGS
         # Additional warnings that don't come with Wall/Wextra:
         -Wduplicated-cond -Wduplicated-branches -Wlogical-op
-
-        # This is since in GCC-8 and triggers on known issues:
-        -Wno-error=cast-function-type
         )
+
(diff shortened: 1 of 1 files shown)


cmake/debugflags.cmake
```diff
@@ -24,11 +24,15 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     list(APPEND FLAGS
         # Additional warnings that don't come with Wall/Wextra:
         -Wduplicated-cond -Wduplicated-branches -Wlogical-op
-
-        # This is since in GCC-8 and triggers on known issues:
-        -Wno-error=cast-function-type
         )
 
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 8.0)
```
t-wissmann: Is it enough to check the compiler version here? Does this make sense for clang?

blarz: We are in the gcc specific block, so the version check should be enough at this point.

clang does not have Wcast-function-type (there is -Wbad-function-cast, but it is neither enabled by default nor via enabled by some other flag), so this should not concern clang for now.

t-wissmann: > We are in the gcc specific block, so the version check should be enough at this point.

I've missed that! Thanks!

PR #653, opened 2020-03-26
Update settings types in autostart and manpage

tiosgz:
Update default autostart to use boolean values and on the manpage description of `toggle` and `set` (allowed values for boolean settings).


t-wissmann:
Thanks!

PR #655, opened 2020-03-27
Add apply_rules command

t-wissmann:
The apply_rules command applies all rules to a specified already managed
client. This is useful for experimenting with rules and for the handling
of clients that change their WM_CLASS after the window appears.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,7 @@ Current git version
 
   * monitor detection via xrandr
   * detection of external panels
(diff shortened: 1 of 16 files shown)


t-wissmann:
in the test case of `test_close`, the command `xlsclients -l` returned a non-zero status during the shutdown when all the remaining clients are killed. I hope that this is not a race condition in our test suite...

PR #658, opened 2020-03-27
Allow to re-apply the rules to all clients

t-wissmann:
Add a switch --all to apply_rules that re-applies the rules to all
(managed) clients.


t-wissmann:
This implements feature request #283

PR #666, opened 2020-03-29
Relax condition in drag test

t-wissmann:


t-wissmann:
The CI run is quite unstable at the moment, but I can't reproduce it locally, so let us relax the condition in the test a bit.

blarz:
Can you check if  #668 fixes this possibly?

t-wissmann:
I guess that your change fixes it :-)

PR #669, opened 2020-03-30
More verbose and human readable xerror output

t-wissmann:
Also do two smaller fixes.


t-wissmann:
Example output:
```
herbstluftwm: fatal error
   resource id: 0x0
   request code: 53 "X_CreatePixmap" (minor code: 0)
   error code: 9 "BadDrawable (invalid Pixmap or Window parameter)"
   serial number: 454
```

PR #674, opened 2020-03-30
x11.create_client() wait for hlwm to process

t-wissmann:
In the x11.create_client() function, wait for hlwm to empty its X event
queue. By this we are sure that everything related to managing the
client is completed.

Hopefully, this makes the test case test_drag_move more stable. It
definitely fixes a race condition, but I'm unsure whether there is
another one.

Many thanks to @The-Compiler and @blarz who helped a lot with debuging
this unstable test case!


t-wissmann:
Let us see if it helps..

PR #677, opened 2020-03-31
Minor fixes in manpages

tiosgz:
There are also some things I wasn't brave enough to edit as I'm not sure about them/didn't know how exactly to do that:
* herbstclient has in its bugs that it waits endlessly, while this is not true now
* also detection via xrandr works (tutorial — pre-last paragraph; main manpage — detect_monitors)
* `complete` — now completion works also with zsh and fish ("because there is already completion for *bash*")
* `keymask` in RULES section deserves a link to its explanation in OBJECTS

### DIFF
--- a/doc/herbstclient.txt
+++ b/doc/herbstclient.txt
@@ -81,8 +81,6 @@ other::
 
 BUGS
 ----
-It waits endlessly for a response from *herbstluftwm* (there is no timeout yet).
-
 See the *herbstluftwm* distribution BUGS file.
 
 COMMUNITY
--- a/doc/herbstclient.txt
+++ b/doc/herbstclient.txt
@@ -48,7 +48,7 @@ OPTIONS
     Same as *--idle* but exit after first *--count* hooks.
 
 *-c*, *--count* 'COUNT'::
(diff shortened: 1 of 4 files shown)


t-wissmann:
Thanks a lot!

> There are also some things I wasn't brave enough to edit as I'm not sure about them/didn't know how exactly to do that:
You're right with all your comments, feel free to edit/delete the wrong text passages:l 
> * herbstclient has in its bugs that it waits endlessly, while this is not true now
> * also detection via xrandr works (tutorial — pre-last paragraph; main manpage — detect_monitors)
> * `complete` — now completion works also with zsh and fish ("because there is already completion for _bash_")


> * `keymask` in RULES section deserves a link to its explanation in OBJECTS
Right, something like: `(see the explanation in the <<OBJECTS,*OBJECTS section*>>)` or do you think a link to the attribute description is easier?

tiosgz:
> Right, something like: `(see the explanation in the <<OBJECTS,*OBJECTS section*>>)` or do you think a link to the attribute description is easier?

It is surely easier to find.

Issue #597, opened 2019-12-29
[Winterbreeze] Frame does not fill entire screen after removing subframe


cnotis82:
Steps to reproduce:
1) Initial frame(without clients) is split in 2 subframes
2) Remove one of subframes, the remaining frame does not fill entire screen.

blarz:
I can reproduce this. The remaining frame has a gap on the right and on the bottom.
Switching to another tag and back fixes this.

t-wissmann:
With the default autostart, I can reproduce it too!

t-wissmann:
It only appears with `smart_frame_surroundings` enabled.

blarz:
I think the problem lies in src/framedecoration.cpp.
`if (settings->smart_frame_surroundings() && !data.hasParent) { bw = 0;}`

The `data.hasParent` check maybe wrong.
If I understand `FrameTree::removeFrameCommand()` correctly,
in case we remove the second to last frame, the last frame becomes the new `root_` like this:
`root_ = newparent;`
Isn't the parent of `newparent` still referencing `root_`?

Not sure if that is the actual problem, I really don't know the code that good (nor do I understand weak and smart pointers fully to be honest).

t-wissmann:
Yes, this was the issue and @blarz fixed it in PR #678.

PR #678, opened 2020-03-31
Fix replacing the root frame

blarz:
When replacing the root frame, make sure it has no parent.
Otherwise smart_frame_surroundings draws frame surroundings...

Fixes bug #597

### DIFF
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -147,7 +147,6 @@ int FrameTree::removeFrameCommand() {
     }
     auto clientFocusIndex = frame->getSelection();
     auto parent = frame->getParent();
-    auto pp = parent->getParent();
     bool insertAtFront = (frame == parent->firstChild());
     auto newparent =
             insertAtFront ? parent->secondChild() : parent->firstChild();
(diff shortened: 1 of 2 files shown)


src/frametree.cpp
```diff
@@ -180,7 +180,7 @@ int FrameTree::removeFrameCommand() {
         pp->replaceChild(parent, newparent);
     } else {
         // if parent was root frame
-        root_ = newparent;
+        replaceNode(root_, newparent);
```
t-wissmann, line 183: Actually, we don't need to get `pp` at all. It should be sufficient to do
```
replaceNode(parent, newparent)
```
instead of the entire `if (pp) { … } else { … }` (and so we don't need the definition of `pp` at all)

src/frametree.cpp
```diff
@@ -577,6 +577,7 @@ void FrameTree::replaceNode(shared_ptr<HSFrame> old,
     auto parent = old->getParent();
     if (!parent) {
         assert(old == root_);
+        // replacement->parent needs to be nulled?
```
t-wissmann, line 580: Yes, definitely! The assertion `!(root_->parent)` must always hold (and is violated here).

src/frametree.cpp
blarz: I changed it, still does not work (I added an assertion and a test that triggers it)

src/frametree.cpp
blarz: Added the assertion

t-wissmann: Maybe I should have said it more explicitly: to fix the issue, I think we need to make the assertion hold, i.e. add
```
root_->parent = {};
```
instead of the assertion

blarz: D'oh. Ok,changed it, now it works. Thank you Thorsten!

Should I remove the test? Or is there a way to get the geometry of a frame, so I can verify it in the test?

blarz: Updated the test

PR #682, opened 2020-04-01
Add test case for mouse resizing in tiling mode

t-wissmann:


t-wissmann:
I've changed my mind and moved the exception handling to another PR.

PR #709, opened 2020-04-05
Add test for checkotherwm

blarz:


blarz:
Not sure where to put the test so I created a new suite..

PR #712, opened 2020-04-05
Test tag restore on wmexec

blarz:
Add some tags and do a wmexec, to see if the tags are still there after
restart. (Also: this is the first test testing wmexec)


blarz:
Thanks @t-wissmann for the help creating this tets!

Issue #713, opened 2020-04-05
Keymask affects other clients


tiosgz:
If I set `keymask` to a client, it affects all other clients with unset keymask until another client with set keymask is focused.

Also, keymask works the opposite way of what's described in manpage, i.e., it *blocks* matching hotkeys (but that's a problem of manpage, the way it works makes more sense to me).

t-wissmann:
Thanks for this very precise issue description! I've fixed the mistake in #714.

Regarding the documentation it seems that the documentation is actually right ([original patch](http://herbstluftwm.org/archive/msg00341.html)) but my implementation (for the winterbreeze rewrite) is the wrong way around. Hence, I have to eventually change the behaviour to make it match all previously released herbstluftwm stable versions.

I think the best way to have both is to add another attribute `keymask_inverse` such that a key press reaches the client only if
1. it matches `keymask` and
2. it does not match `keymask_inverse`

What do you think? Do you have a better name than `keymask_inverse`?

tiosgz:
> I've fixed the mistake in #714.

Thanks!
> I think the best way to have both is to add another attribute `keymask_inverse` such that a key press reaches the client only if
> 
> 1. it matches `keymask` and
> 2. it does not match `keymask_inverse`
> 
> What do you think?

This would be good. But I think you meant *HLWM* instead of *client*.
> Do you have a better name than `keymask_inverse`?

`(hot)keys_disabled`?

t-wissmann:
> Also, keymask works the opposite way of what's described in manpage, i.e., it _blocks_ matching hotkeys (but that's a problem of manpage, the way it works makes more sense to me).

For the record: by PR #715, what was called `keymask` in winterbreeze (git version of the past two weeks) is now called `keys_inactive` and has the opposite meaning of `keymask` (as described in the man page). For the moment `keymask` has no effect, but I'll reimplement the documented meaning of `keymask` soon.

Thanks again to @tiosgz for spotting this inconsistency!

PR #715, opened 2020-04-05
Introduce keys_inactive client attribute

t-wissmann:
The current winterbreeze implementation of keymask was wrong (negated).
Since this is proven to be useful, to block certain keybindings for
concrete clients, we call this feature 'keys_inactive'.

This makes 'keymask' currently an attribute without any effect. The next
step is then to bring back the old behaviour of 'keymask'

See also #713.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,7 @@ Current git version
   * detection of external panels
   * new command: apply_rules
   * new command: export (convenience wrapper around setenv)
+  * new client attribute: key_inactive (negation of keymask)
   * new command: drag (initiates moving/resizing a window by mouse)
   * if tags have been configured through EWMH before herbstluftwm starts (from
     a previous running window manager), then herbstluftwm re-uses these tags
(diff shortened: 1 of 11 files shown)


doc/herbstluftwm.txt
```diff
@@ -1415,11 +1418,17 @@ listed as follows:
...
-                            The default keymask is an empty string (""), which
-                            does not disable any keybinding.
+                            bindings (as they are printed by list_keybinds).
+                            While this client is focused, only bindings that
+                            match the expression will be active. Any other
+                            bindings will be disabled. The default keymask is an
+                            empty string (""), which does not disable any
+                            keybinding.
+|s - keys_inactive        | A regular expression that describes which
+                            keybindings are inactive for the client. If a key
```
tiosgz, line 1428: "for the client" → "while the client is focused"?

t-wissmann: Yes, thanks!

NEWS
```diff
@@ -10,6 +10,7 @@ Current git version
   * detection of external panels
   * new command: apply_rules
   * new command: export (convenience wrapper around setenv)
+  * new client attribute: key_inactive (negation of keymask)
```
tiosgz, line 13: You're missing an "s" here ;-)

t-wissmann: Thanks! I've fixed it directly in `master`.

PR #724, opened 2020-04-07
Add test for default autostart

blarz:
We should test if the default autostart from share/ works by running it
with bash -e.

### DIFF
--- a/tests/test_herbstluftwm.py
+++ b/tests/test_herbstluftwm.py
@@ -17,9 +17,10 @@ def test_herbstluftwm_already_running(hlwm):
 
 def test_herbstluftwm_default_autostart(hlwm):
     expected_tags = [str(tag) for tag in range(1, 10)]
-    # FIXME: Is there a more robust way to find the default autostart path?
     default_autostart = os.path.join(os.path.abspath(BINDIR), 'share/autostart')
-    subprocess.run(['bash', '-e', default_autostart], check=True)
+    env_with_bindir_path = os.environ.copy()
+    env_with_bindir_path['PATH'] = BINDIR + ":" + env_with_bindir_path['PATH']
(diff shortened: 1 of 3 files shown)


tests/test_herbstluftwm.py
```diff
@@ -13,3 +13,18 @@ def test_herbstluftwm_already_running(hlwm):
...
+def test_herbstluftwm_default_autostart(hlwm):
+    expected_tags = [str(tag) for tag in range(1, 10)]
+    # FIXME: Is there a more robust way to find the default autostart path?
+    default_autostart = os.path.join(os.path.abspath(BINDIR), '..', 'share/autostart')
+    autostart = subprocess.Popen(['/usr/bin/env', 'bash', '-e', default_autostart],
+                                 stderr=subprocess.PIPE,
+                                 stdout=subprocess.PIPE,
+                                 universal_newlines=True)
+    autostart.wait(2)
+    assert autostart.returncode == 0
```
The-Compiler, line 27: This can probably be simplified to `subprocess.run(['bash', '-e', default_autostart], check=True)`

blarz: You're right, your way is so much simpler. Updated the PR accordingly. Thank you!

t-wissmann:
By the way, this does not check that the panel is working. For the default panel, one needs to install dzen... But I'm fine with not installing dzen2 into our CI system at the moment, because I hope that the users install another panel that is more powerful than then panel.sh from the herbstluftwm repository.

The-Compiler:
> because I hope that the users install another panel that is more powerful than then panel.sh from the herbstluftwm repository.

Might be a good idea to somehow make [barpyrus](https://github.com/t-wissmann/barpyrus) into a more official hlwm-panel?

PR #748, opened 2020-04-10
Fix narrowing warning in ewmh.cpp

blarz:
Also enable the narrowing warning in debug builds again, now this is
fixed.

### DIFF
--- a/src/ewmh.cpp
+++ b/src/ewmh.cpp
@@ -460,6 +460,9 @@ bool Ewmh::isFullscreenSet(Window win) {
 }
 
 void Ewmh::setWindowOpacity(Window win, double opacity) {
+    /* Based on the EWMH proposal
+     * https://mail.gnome.org/archives/wm-spec-list/2003-December/msg00035.html
+     */
     long long_opacity = 0xffffffff * CLAMP(opacity, 0, 1);
     X_.setPropertyCardinal(win, g_netatom[NetWmWindowOpacity], { long_opacity });
 }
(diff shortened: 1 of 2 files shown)


blarz:
I am not sure if this is correct, but .setPropertyCardinal wants a vector of longs, so it makes sense.

src/ewmh.cpp
```diff
@@ -460,10 +460,10 @@ bool Ewmh::isFullscreenSet(Window win) {
 }
 
 void Ewmh::setWindowOpacity(Window win, double opacity) {
-    uint32_t int_opacity = std::numeric_limits<uint32_t>::max()
+    long long_opacity = std::numeric_limits<long>::max()
                             * CLAMP(opacity, 0, 1);
 
-    X_.setPropertyCardinal(win, g_netatom[NetWmWindowOpacity], { int_opacity });
```
t-wissmann, line 466: I think this is wrong because the opacity value in the window property should not depend on the size of `long`. But I can't find the documentation for `_NET_WM_WINDOW_OPACITY`.

t-wissmann: The [`compton-trans`](https://github.com/chjj/compton/blob/master/bin/compton-trans#L206) utility sets the value to
```
opacity=$((opacity * 0xffffffff / 100))
```
So maybe we can simply multiply by `0xffffffff` instead of the limit of whatever type.

blarz: Thanks, changed it and added a comment. Is this correct now?

t-wissmann: yes, I think so, thanks!

PR #750, opened 2020-04-10
Fix Travis check in Mergify config

blarz:


blarz:
See https://doc.mergify.io/conditions.html#generic-status-check

Issue #732, opened 2020-04-08
cycle_all (Alt-Tab) should cycle through all windows


t-wissmann:
The cycle_all command should traverse both tiled and floated clients.

PR #752, opened 2020-04-11
Separate cycle_all implementation from command

t-wissmann:
This is another step towards #732. Here, the implementation of the
cycle_all command is separated from the command, and even the wrapping
is now configurable. So now, it's easy to let cycle_all also traverse
the floating windows.

The test cases still fully cover the functionality as it is at the
moment.

### DIFF
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -400,20 +400,73 @@ int FrameTree::cycleAllCommand(Input input, Output output) {
     if (delta == 0) {
         return 0; // nothing to do
     }
+    CycleDelta cdelta = (delta == 1) ? CycleDelta::Next : CycleDelta::Previous;
+    bool succeeded = cycleAll(cdelta, skip_invisible);
+    if (!succeeded) {
+        // we need to wrap. when cycling forward, we wrap to the beginning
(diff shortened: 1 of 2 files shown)


t-wissmann:
This can wait until travis is online again.

src/frametree.cpp
```diff
@@ -400,20 +400,73 @@ int FrameTree::cycleAllCommand(Input input, Output output) {
...
+        // we need to wrap. when cycling forward, we wrap to the beginning
+        CycleDelta rewind = (delta == 1) ? CycleDelta::Begin : CycleDelta::End;
+        cycleAll(rewind, skip_invisible);
+    }
+    // finally, redraw the layout
+    get_current_monitor()->applyLayout();
+    return 0;
+}
+
+//! go to the specified frame. Return true on succes, return false if
```
blarz, line 415: typo succes

src/frametree.cpp
```diff
@@ -400,20 +400,73 @@ int FrameTree::cycleAllCommand(Input input, Output output) {
...
+//! the end is reached (this command never wraps). Skipts covered windows
```
blarz, line 416: Skips

Issue #710, opened 2020-04-05
Move new clients to empty frames


t-wissmann:
Implement a frame index 'E' that denotes a frame that is empty.
```
hc rule focus=on index=E
```
This rule places new (tiled) clients in an empty frame (if there is one). Bonus exercise:
```
hc rule focus=on index=0E
```
Puts new clients in an empty frame in the first subtree of the root.

 This task involves:
* extend `shared_ptr<HSFrame> FrameTree::lookup(const string& path)` in frametree.cpp
* extend the man page
* extend NEWS
* add a test (in test_layout.py)

t-wissmann:
Implemented by PR #753.

PR #753, opened 2020-04-11
Frame index for empty frame

t-wissmann:
A new frame index 'e' specifier that finds an empty frame as close as
possible to the focused frame.

This implements #710.


t-wissmann:
I'm not satisfied yet how the empty frame is picked. If there are many empty frames, currently the frame most top left is picked, but this is a bit surprising.

t-wissmann:
> I'm not satisfied yet how the empty frame is picked. If there are many empty frames, currently the frame most top left is picked, but this is a bit surprising.

Now, it's better: I take the frame that is as close as possible in the tree and geometrically.

PR #758, opened 2020-04-13
Fix using std checker

blarz:
The reporoot directory was not correctly set,
so CI build just complained about

  grep: /hlwm/ci/src/*.cpp: No such file or directory

Fix this and all found occurrences.

### DIFF
--- a/src/framedata.cpp
+++ b/src/framedata.cpp
@@ -1,8 +1,11 @@
 #include "framedata.h"
 
+#include <regex>
+
 #include "completion.h"
 
 using std::string;
+using std::regex;
 
 static const char* g_align_names[] = {
     "vertical",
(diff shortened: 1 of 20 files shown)


src/framedata.cpp
```diff
@@ -1,8 +1,11 @@
 #include "framedata.h"
 
+#include <regex>
+
 #include "completion.h"
 
 using std::string;
+using std::regex;
```
blarz, line 8: I think this was an accident, will remove again

src/rules.cpp
```diff
@@ -10,6 +10,8 @@
 #include "utils.h"
 #include "xconnection.h"
 
+using std::regex;
+using std::regex_error;
 using std::string;
```
t-wissmann, line 15: Why `using std::regex`? I'd prefer not `use` it globally but to keep the `std::` prefix here..

t-wissmann: Ah, it's my mistake: there is:
```
regexstr.cpp:using std::regex;
```
So it's in the scope of another PR

PR #759, opened 2020-04-13
Explicitly provide add() function in TilingResult

t-wissmann:
The operator[] should not have side-effects (I ran into very unexpected
behaviour when calling 'res[client]' multiple times without know that it
adds element to the vector). For clarity, make it a function named
'add()'.

### DIFF
--- a/src/tilingresult.h
+++ b/src/tilingresult.h
@@ -22,7 +22,6 @@ public:
 class TilingResult {
 public:
     TilingResult() = default;
-    TilingStep& operator[](Client* client);
     void add(Client* client, const TilingStep& client_data);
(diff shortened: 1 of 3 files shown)


src/tilingresult.h
```diff
@@ -23,6 +23,7 @@ class TilingResult {
 public:
     TilingResult() = default;
     TilingStep& operator[](Client* client);
```
dnnr, line 25: Remove this? I wonder why the build doesn't fail. Is it because it is never called? 🤔

t-wissmann: Right, thanks!
Yes, the build doesn't fail because nothing links to it.

PR #760, opened 2020-04-13
Simplify reporoot definition in CI script

dnnr:
Assuming that the magic %*/ construct serves no special purpose, using
/../ is more self-explanatory and (as we just learned) less error-prone.


blarz:
Yes please, finding out what the /* magic does was kind of confusing.

PR #765, opened 2020-04-13
Use empty() for checking container size

blarz:
Found by clang-tidy


blarz:
I personally find it nicer to read, but I'd understand if this is not something we want.

PR #766, opened 2020-04-13
Add braces around statements

blarz:
Found with clang-tidy and formatted with clang-format-diff


blarz:
Also up to discussion but I always feel the need to put braces around single if/while statements regardless of their shortness..

src/layout.cpp
```diff
@@ -493,6 +515,52 @@ void HSFrameSplit::swapChildren() {
     swap(a_,b_);
 }
 
+int frame_change_fraction_command(int argc, char** argv, Output output) {
```
blarz, line 518: Messed up the rebase..

PR #769, opened 2020-04-13
Add clang-tidy to CI Dockerfile (disco)

dnnr:
Doing this separately from adding clang-tidy to the CI pipeline will
make it easier to identify the impact clang-tidy has on the CI duration.


blarz:
So clang-tidy-8 and clang-7 work together? (Just asking to be sure)

dnnr:
> So clang-tidy-8 and clang-7 work together? (Just asking to be sure)

I didn't even notice that version difference, but I also didn't see any problems with it.

blarz:
> > So clang-tidy-8 and clang-7 work together? (Just asking to be sure)
> 
> I didn't even notice that version difference, but I also didn't see any problems with it.

Ok then.

PR #773, opened 2020-04-13
Fix diverging parameter name and definitions

blarz:
Found by clang-tidy

### DIFF
--- a/src/frametree.h
+++ b/src/frametree.h
@@ -34,7 +34,7 @@ public:
     bool focusClient(Client* client);
     //! focus a frame within its tree
     static void focusFrame(std::shared_ptr<HSFrame> frame);
-    bool focusInDirection(Direction dir, bool externalOnly);
+    bool focusInDirection(Direction dir, bool external_only);
     //! return a frame in the tree that holds the client
     std::shared_ptr<HSFrameLeaf> findFrameWithClient(Client* client);
 
--- a/src/frametree.h
+++ b/src/frametree.h
@@ -34,7 +34,7 @@ public:
     bool focusClient(Client* client);
     //! focus a frame within its tree
     static void focusFrame(std::shared_ptr<HSFrame> frame);
-    bool focusInDirection(Direction dir, bool external_only);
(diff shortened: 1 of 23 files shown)


blarz:
Please discuss. Maybe some parameters should be named after how they are defined and vice versa.

src/frametree.h
```diff
@@ -34,7 +34,7 @@ class FrameTree : public std::enable_shared_from_this<FrameTree> {
     bool focusClient(Client* client);
     //! focus a frame within its tree
     static void focusFrame(std::shared_ptr<HSFrame> frame);
-    bool focusInDirection(Direction dir, bool externalOnly);
```
dnnr, line 37: I think we agreed on camelCase as the canonical style (even though it's not consistently used yet)

src/utils.h
```diff
@@ -87,7 +87,7 @@ void set_window_double_border(Display *dpy, Window win, int ibw,
 // returns the unichar in GSTR at position GSTR
 
 size_t      utf8_string_length(const std::string& str);
-std::string utf8_string_at(const std::string& str, size_t offset);
```
dnnr, line 90: Looks to me like `offset` would be the better name here

dnnr:
> Please discuss. Maybe some parameters should be named after how they are defined and vice versa.

I might find some more nitpicks, but for most of the cases, I mind the mismatch more than the naming. Except for my two comments, this looks good to me.

src/utils.h
blarz: Makes sense!

src/frametree.h
blarz: Fixed

blarz:
> > Please discuss. Maybe some parameters should be named after how they are defined and vice versa.
> 
> I might find some more nitpicks, but for most of the cases, I mind the mismatch more than the naming. Except for my two comments, this looks good to me.

Updated it. Yeah not all names are optimal, maybe other clang* tools will yell at us later.

blarz:
'readability-inconsistent-declaration-parameter-name' could be added to the clang-tidy PR, now too

PR #774, opened 2020-04-13
CI: Use ancient GCC to build for 32bit

dnnr:


.travis.yml
```diff
@@ -87,6 +87,8 @@ jobs:
                 --volume=$HOME/.ccache-gcc-4.8:/.ccache:rw
                 --env CC=gcc-4.8
                 --env CXX=g++-4.8
+                --env CXXFLAGS=-m32
```
blarz, line 90: As Werror is disabled in debugflags.cmake and we explictly enable it in build.py (which is not used here), we should add it here too to catch actual build errors.

blarz: Also we are not doing a Debug build here, apparently. Not sure if this is really desired?

t-wissmann: Not sure about that. For me it's important that herbstluftwm builds for people with old systems. Hence, I wouldn't care so much about `-Werror` and debug build.

dnnr: I tend to agree with @t-wissmann here. On the other hand, it probably doesn't hurt either?

dnnr: In any case, it'd be separate PR. :-)

blarz: Ok then!

PR #777, opened 2020-04-13
Fix bash completion

tiosgz:
With the previous `herbstclient-completion` it did not really complete for me.


blarz:
How this fix it? Was the filename incorrect?

tiosgz:
Yeah (or at least on my system (Manjaro))

tiosgz:
Oh! I should have tested if it works…

t-wissmann:
I think the name in the git repository was correct: `share/herbstclient-completion` says what the file is good for (and `share/herbstclient-bash-completion` would be better, maybe even a separate directory for completions). However `share/herbstclient` is a misleading filename, in my opinion. So I think CMake should be configured in such a way that it changes the filename during installation (I don't know how to do this); maybe this is even obsolete, if the filepath is `shell-completions/herbstclient` (maybe even `shell-completions/herbstclient.bash`)

tiosgz:
I'm not sure if I can do anything more than renaming it to `herbstclient.bash`, which seems to work (as you can see I'm even so naive that I thought just renaming the file is enough). Could you help me please?

tiosgz:
OK, so I somehow managed to do it. However, I don't know how the file/folder system works, so I cannot move it into a separate directory.

I'm sorry for all those file renames, I guess I could have done it more systematically.

t-wissmann:
Looks good, thanks! Don't worry about the renames. all commits of this PR are squashed into a single commit anyway.

blarz:
This is merged now, but do we need to update the docs, now, too? faq.txt and INSTALL

t-wissmann:
yes. faq.txt only mentioned the old install path. And I think INSTALL needs a reorganization anyway.

Issue #780, opened 2020-04-14
Failed build on i386 (FreeBSD)


uzsolt:
http://beefy5.nyi.freebsd.org/data/121i386-default/531650/logs/herbstluftwm-0.8.0.log
```
/wrkdirs/usr/ports/x11-wm/herbstluftwm/work/herbstluftwm-0.8.0/src/ewmh.cpp:146:42: warning: format specifies type 'unsigned long' but the argument has type 'size_t' (aka 'unsigned int') [-Wformat]
    fprintf(file, "EWMH: %lu desktops:", numberOfDesktops);
                         ~~~             ^~~~~~~~~~~~~~~~
                         %zu
/wrkdirs/usr/ports/x11-wm/herbstluftwm/work/herbstluftwm-0.8.0/src/ewmh.cpp:151:44: warning: format specifies type 'unsigned long' but the argument has type 'std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >::size_type' (aka 'unsigned int') [-Wformat]
    fprintf(file, "%lu managed clients: ", original_client_list_.size());
                   ~~~                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
                   %u
/wrkdirs/usr/ports/x11-wm/herbstluftwm/work/herbstluftwm-0.8.0/src/ewmh.cpp:466:66: error: non-constant-expression cannot be narrowed from type 'uint32_t' (aka 'unsigned int') to 'long' in initializer list [-Wc++11-narrowing]
    X_.setPropertyCardinal(win, g_netatom[NetWmWindowOpacity], { int_opacity });
                                                                 ^~~~~~~~~~~
/wrkdirs/usr/ports/x11-wm/herbstluftwm/work/herbstluftwm-0.8.0/src/ewmh.cpp:466:66: note: insert an explicit cast to silence this issue
    X_.setPropertyCardinal(win, g_netatom[NetWmWindowOpacity], { int_opacity });
                                                                 ^~~~~~~~~~~
                                                                 static_cast<long>( )
/wrkdirs/usr/ports/x11-wm/herbstluftwm/work/herbstluftwm-0.8.0/src/ewmh.cpp:478:45: error: non-constant-expression cannot be narrowed from type 'WmState' to 'long' in initializer list [-Wc++11-narrowing]
    X_.setPropertyCardinal(win, WM_STATE, { state, None });
                                            ^~~~~
[...]

PR #781, opened 2020-04-14
Make WM_STATE an enum class

t-wissmann:
The enum class also forces us to explicitly cast the state in the call
to setPropertyCardinal(), and this hopefully fixes the compilation error
in #780.


blarz:
Looks good to me!

PR #782, opened 2020-04-14
Enable some clang-analyzer-unix checks

dnnr:
These checks used to fail when initially adding clang-tidy to the pipeline.
The underlying issues have probably been fixed in the mean-time.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -16,7 +16,7 @@
 #   # Checks: '*,-abseil-*,-android-*,-objc-*,-fuchsia-*,-mpi-*,-performance-*,-llvm-include-order'
 #   ###
 #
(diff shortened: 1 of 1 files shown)


blarz:
Explain?

dnnr:
> Explain?

I was confused, too. See description, now

dnnr:
Added another check...

blarz:
I think the malloc one was in #767 maybe?

dnnr:
> I think the malloc one was in #767 maybe?

It was indeed!

blarz:
Does simply removing the checks from the '-' list also add them automatically?
Or do we need to add them explicitly?

blarz:
> Does simply removing the checks from the '-' list also add them automatically?
> Or do we need to add them explicitly?

Ah just looked at the list of enabled checks in the build output, the checks are enabled.

dnnr:
> Does simply removing the checks from the '-' list also add them automatically?
> Or do we need to add them explicitly?

This is not really obvious. Technically, I'm not enabling those checks, but I'm allowing them to run if they are in the default set of checks (which they were when introducing clang-tidy).

PR #783, opened 2020-04-14
Enable clang-tidy check StackAddressEscape

dnnr:


blarz:
That one is tricky complains about exit_on_error and another variable being passed around in a static struct.

dnnr:
> That one is tricky complains about exit_on_error and another variable being passed around in a static struct.

Would making `exit_on_xerror` also `static` be a valid solution? I wonder why `long_options` is static in first place 🤔

dnnr:
I think removing that `static` is the right way: #784

t-wissmann:
Do I understand it correctly that these checks are enabled because they are deleted from the blacklist?

dnnr:
> Do I understand it correctly that these checks are enabled because they are deleted from the blacklist?

Exactly!

PR #785, opened 2020-04-14
Increase timeouts for herbstclient termination

t-wissmann:
With more clang analyzer checks, it seems that the timeout does not
suffice anymore...


blarz:
> With more clang analyzer checks, it seems that the timeout does not
> suffice anymore...

Does this really depend on the clang-tidy checks? Is the load too high during tests because of it?

tests/test_herbstclient.py
```diff
@@ -85,7 +85,7 @@ def test_herbstclient_wait(hlwm, num_hooks_sent, num_hooks_recv):
         hlwm.call('emit_hook nonmatch nonarg')
         hlwm.call('emit_hook matcher somearg')
 
-    proc.wait(3)
+    proc.wait(5)
```
blarz, line 88: Should we extract the timeout to a variable used in both tests?

t-wissmann:
> > With more clang analyzer checks, it seems that the timeout does not
> > suffice anymore...
> 
> Does this really depend on the clang-tidy checks? Is the load too high during tests because of it?

I don't know. Aftere the clang-tidy checks were introduced, the timeout was reached by the following two commits, but it didn't appear again afterwards.

Maybe we can dismiss this PR and see whether the issues shows up again..

tests/test_herbstclient.py
t-wissmann: not sure whether it's an improvement

blarz: It's not that important to me, either ;)

blarz:
> > > With more clang analyzer checks, it seems that the timeout does not
> > > suffice anymore...
> > 
> > 
> > Does this really depend on the clang-tidy checks? Is the load too high during tests because of it?
> 
> I don't know. Aftere the clang-tidy checks were introduced, the timeout was reached by the following two commits, but it didn't appear again afterwards.
> 
> Maybe we can dismiss this PR and see whether the issues shows up again..

Let's see if it happens again, I guess the number of PRs will stay high in the following days ;)

PR #789, opened 2020-04-14
Update FAQ and INSTALL

tiosgz:
Aside from (again) updating bash completion paths, update the FAQ.


INSTALL
```diff
@@ -84,8 +84,8 @@ valgrind-xephyr.sh from your existing X session:
     mkdir -p ~/.config/herbstluftwm
     cp share/autostart ~/.config/herbstluftwm/
 
-5. add the share/herbstclient-completion to your
-   /usr/share/bash-completion/completions folder or source it in your bashrc
+5. add the share/herbstclient to your /usr/share/bash-completion/completions
```
t-wissmann, line 87: Isn't it `herbstclient-completion.bash` now? Maybe `copy` is better than `add` by the way

www/faq.txt
```diff
@@ -91,13 +91,11 @@ the according (internal) command directly is called.
 
 Q: How can I let single clients float?
 --------------------------------------
-Not at all. You don't need it. You have the power of manual tiling, so
-there is no need to place clients manually by dragging. Even better:
-You don't even need to place them manually. Use a rule to place special
-dialogs automatically when they appear. See the GIMP-Example for a good
-example.
```
t-wissmann, line 98: Great catch! Maybe add a `Since herbstluftwm 0.8.0, …` at the beginning.

INSTALL
tiosgz: Right, I forgot it's still in the pre-build dir structure. Thanks.

www/faq.txt
tiosgz: Can be now?

t-wissmann: 'Now' is imprecise, and I don't see the any disadvantage of being precise here :). If you don't want to start he answer with 'Since' you can also add `(since 0.8.0)` at the end of the answer.

tiosgz: I meant now = after the latest commit. Though it's also not completely precise…

t-wissmann: Now it's a bit misleading. Now, it sounds like the shortcut method also works for hlwm older than 0.8.

tiosgz: And now it sounds like a history lesson :-P

t-wissmann: I think it's good now, thanks!

PR #790, opened 2020-04-14
First draft of keyboard resizing in floating mode

t-wissmann:
This makes the resize command also work for clients in floating mode at
resizes the window into the specified direction to the next visible edge
on the screen. If growing is not possible anymore, then the window
shrinks again.

The current implementation does not work entirely for windows with
sizehints, such as terminals: if the resize delta is within the size
granularity, the current algorithm thinks that a resize is not possible
instead of looking for the next edge.


t-wissmann:
I think we can merge this and improve it later.

PR #792, opened 2020-04-14
Add safety parenthesis around macro arguments

dnnr:


src/utils.h
```diff
@@ -25,7 +25,7 @@ int MOD(int x, int n);
 #define DBGDO(X) { if(1) std::cerr << "DBGDO: " << (#X) << " = " << (X) << std::endl; }
 
 // control structures
-#define FOR(i,a,b) for (int i = (a); i < (b); i++)
+#define FOR(i,a,b) for (int (i) = (a); (i) < (b); (i)++)
```
t-wissmann, line 28: compilation says:
```
./src/utils.h:28:29: error: unnecessary parentheses in declaration of ‘i’ [-Werror=parentheses]
   28 | #define FOR(i,a,b) for (int (i) = (a); (i) < (b); (i)++)
      |                             ^
../src/monitormanager.cpp:80:5: note: in expansion of macro ‘FOR’
   80 |     FOR (i,0,size()) {
      |     ^~~
```

dnnr: Oof. clang is happy with it, but gcc actually complains about it being unnecessary. I'll revert this fix for now...

dnnr:
> By the way, I hope to get rid of most of these macros anyway.

That's good to hear. Usually, when they're easy to remove, I do it right away.

PR #794, opened 2020-04-14
Enable clang-tidy check bugprone-macro-parentheses

dnnr:


dnnr:
We'll need to get rid of the `FOR` macro before enabling this check, because GCC disagrees with the suggestion made by clang-tidy.

PR #796, opened 2020-04-14
Keyboard resizing of sizehints clients

t-wissmann:
As a follow up to #790 / b3e6419, make the keyboard resizing of floated
clients also work for floated clients with sizehints. The code is
getting a bit messy, but I leave the clean-up for future PRs.


t-wissmann:
I'm merging it because I'd like to test it in my working setup and because I wanted this for a very long time. (I'll also add tests at some point, hopefully soon).

PR #798, opened 2020-04-15
Get rid of FOR macro

t-wissmann:
This replaces the FOR loop macro, by either foreach loops or by explicit
for loops. Sometimes, we need indices in foreach loops, so we need to
explicitly count the index there. (I find this easier to read than an
explicit for-loop with iterators).

### DIFF
--- a/src/floating.cpp
+++ b/src/floating.cpp
@@ -105,7 +105,7 @@ int find_rectangle_right_of(RectangleIdxVec rects, int idx) {
     int cy = RC.y + RC.height / 2;
     int write_i = 0; // next rectangle to write
     // filter out rectangles not right of RC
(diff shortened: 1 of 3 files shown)


t-wissmann:
* For the index counting, we can consider using a library like https://github.com/ignatz/pythonic that provides a python-style `enumerate(…)`
* One occurence is not yet replaced: the `detect_monitors` command. I think its better to migrate it to the new command and completion syntax in a separate PR.

dnnr:
>     * For the index counting, we can consider using a library like https://github.com/ignatz/pythonic that provides a python-style `enumerate(…)`

The way you can do it in Python is certainly cleaner, but I don't think this is worth pulling in a 3rd-party helper.

PR #801, opened 2020-04-16
Migrate detect_monitors command

t-wissmann:
Make the detect_monitors command a member of the MonitorManager, and
convert it and its completion to the new command format.


tests/test_monitor.py
```diff
@@ -127,11 +127,25 @@ def test_new_clients_appear_in_focused_monitor(hlwm):
...
 
-def test_detect_monitors_does_not_crash(hlwm):
+@pytest.mark.parametrize("arg", ['-l', '--list-all', '--no-disjoin'])
+def test_detect_monitors_does_not_crash(hlwm, arg):
     # I don't know how to test detect_monitors properly, so just check that
     # it does not crash at least
+    hlwm.call(['detect_monitors', arg])
+
+
+def test_detect_monitors_root_window(hlwm):
```
dnnr, line 137: I don't quite get this test name. Isn't it more something `test_detect_monitors_has_effect_on_monitors_list` or `test_detect_monitors_yields_correct_monitors_list`?

tests/test_monitor.py
```diff
@@ -127,11 +127,25 @@ def test_new_clients_appear_in_focused_monitor(hlwm):
...
+    list_monitors_before = hlwm.call('list_monitors').stdout
+    hlwm.call('add othertag')
+    hlwm.call('set_monitors 80x80+5+5 80x80+85+5')
+    assert hlwm.get_attr('monitors.count') == '2'
+
+    # check that detect monitors restores the one monitor
     hlwm.call('detect_monitors')
 
+    assert hlwm.get_attr('monitors.count') == '1'
```
dnnr, line 146: Could we make assumptions about the Xvfb monitor size and verify it here?

t-wissmann:
This removes the last usage of the `FOR` macro. Once this is merged, #794 should succeed.

tests/test_monitor.py
t-wissmann: This is what I initially intended to do (when writing the test case name). But then I decided to do the full testing in a separate PR (at least with Xephyr, one can simulate multiple screens).

tests/test_monitor.py
t-wissmann: Yes, (see other comment). I've updated the name.

tests/test_monitor.py
dnnr: Alright, nice!

dnnr: Oh, and I totally missed that you basically already did what I had in mind by verifying that it matches the before-state. Sorry...

PR #804, opened 2020-04-16
Fix and check for implicit pointer-to-bool convs

dnnr:


blarz:
Needs a merge from master :)

src/types.cpp
```diff
@@ -36,7 +36,7 @@ void Input::replace(const string &from, const string &to)
 template<> void Converter<bool>::complete(Completion& complete, bool const* relativeTo)
 {
     complete.full({ "on", "off", "true", "false" });
-    if (relativeTo) {
+    if (relativeTo != nullptr) {
         complete.full("toggle");
```
t-wissmann, line 40: What's bad about `if (relativeTo)`? I'm wondering that this is really the only place in the source where I'm doing this.

dnnr: I was surprised as well that this is the only place, but beware that this is only for bool pointers, not any kind of pointer.

Other than usual, they don't have a comprehensive reasoning for this check on the docs. I suppose that it is considered a risky pattern because one might too easily confuse whether the pointer is checked for nullptr or the bool is checked for its value.

t-wissmann: I see thanks! I indeed missed the 'only to bool' part :-) The check makes perfect sense to me, and it also makes sense that this is the only place where a pointer to a boolean is passed around.

:+1:

Issue #803, opened 2020-04-16
One window in frame always shows border


bernerha:
With version 0.8 the border around a single window in a single frame always get displayed.
The version before behaved different. The border only get showed if more than one window was visible in a frame.

Is this a bug or a fix?

Regards

t-wissmann:
You're right, the setting `smart_window_surroundings` is ignored in 0.8.0. This is indeed a bug, thanks for the report!

blarz:
Weird, shouldn't tests/test_layout.py::smart_window_surroundings test exactly that?

t-wissmann:
> Weird, shouldn't tests/test_layout.py::smart_window_surroundings test exactly that?

Took me a while to see it: this test only checks that the `window_gap` is applied correctly but does not check the `border_width`.

t-wissmann:
Fixed by #807.

bernerha:
Nice, thank you. Do you plan to create a release at any time?

t-wissmann:
Yes, there are two open pull requests with bugfixes which I will merge soon. And then, I think it's the perfect occasion for another release.

PR #807, opened 2020-04-16
Activate smart_window_surroundings again

t-wissmann:
In the TilingResult, explicitly mention which clients should have a
minimal decoration, and later pass that to Client::resize_tiling().

This solves #803.

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -350,7 +350,11 @@ TilingResult HSFrameLeaf::computeLayout(Rectangle rect) {
             layoutResult = layoutHorizontal(rect);
             break;
     }
-    if (!smart_window_surroundings_active) {
+    if (smart_window_surroundings_active) {
+        for (auto& it : layoutResult.data) {
+            it.second.minimalDecoration = true;
+        }
+    } else { // if (!smart_window_surroundings_active)
         // apply window gap: deduct 'window_gap' many pixels from
         // bottom and right of every window:
         for (auto& it : layoutResult.data) {
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -354,7 +354,7 @@ TilingResult HSFrameLeaf::computeLayout(Rectangle rect) {
         for (auto& it : layoutResult.data) {
             it.second.minimalDecoration = true;
         }
-    } else { // if (!smart_window_surroundings_active)
+    } else {
         // apply window gap: deduct 'window_gap' many pixels from
         // bottom and right of every window:
         for (auto& it : layoutResult.data) {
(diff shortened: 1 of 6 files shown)


src/layout.cpp
```diff
@@ -350,7 +350,11 @@ TilingResult HSFrameLeaf::computeLayout(Rectangle rect) {
             layoutResult = layoutHorizontal(rect);
             break;
     }
-    if (!smart_window_surroundings_active) {
+    if (smart_window_surroundings_active) {
+        for (auto& it : layoutResult.data) {
+            it.second.minimalDecoration = true;
+        }
+    } else { // if (!smart_window_surroundings_active)
```
blarz, line 357: Is the commented out if statement really necessary?

tests/test_layout.py
```diff
@@ -352,10 +352,12 @@ def test_focus_directional_2x2grid(hlwm, client_focused, direction):
...
+@pytest.mark.parametrize("border_width", [0, 7])
+def test_smart_window_surroundings(hlwm, x11, border_width):
     hlwm.call('set_layout vertical')
     hlwm.call('set frame_gap 0')
-    hlwm.call('set window_border_width 0')
+    hlwm.call('attr theme.minimal.border_width 0')  # only until #806 is merged
+    hlwm.call(f'set window_border_width {border_width}')
     hlwm.call('set frame_border_width 0')
     hlwm.call('set frame_padding 0')
     hlwm.call('set smart_window_surroundings on')
```
blarz, line 363: In #803 you said that this test did not test the border width here. Should we add that now, so regressions are detected?
Checking that no border is drawn around the single window.

src/layout.cpp
t-wissmann: Not sure, probably not.. I can remove it

tests/test_layout.py
t-wissmann: This happens already in line 373 and 374. However, before this PR and before #806, there was really no border drawn because:
  * the usual border width was set to 0
  * smart_window_surroundings was not respected, so the usual border width was applied in line 372.

Since I now also set the window border width to 7, the tests in line 373 and 374 will fail on a regression.

t-wissmann: I've updated the coded and now:
* I also test different values for `theme.minimal.border_width`
* so we know that really the 'minimal decoration' is used for the client here
* so we are independent from PR #806 (the correct initial value is tested there and not here anyway)

src/layout.cpp
blarz: Thanks!

tests/test_layout.py
blarz: Awesome, thank you!

PR #808, opened 2020-04-16
Rename Ewmh member variables

t-wissmann:
Make the member variable names fit into the naming scheme of camelCase
with underscore-suffix. The rename was done using qtcreator's
refactoring tool.

Also remove the WM_STATE member variable, because the atom it holds is
already contained in the wmatom_ array.


t-wissmann:
This is the follow up of #787

PR #810, opened 2020-04-17
Test monitor detection via xinerama

t-wissmann:
For this, we also need to install Xephyr in the docker image in which we
run tests.


t-wissmann:
I'm a bit confused: the CI fails if I close the pipe after reading the display number from it. The CI fails because the Xfvb process can't write to the file descriptor anymore according to the Xfvb output (which itself is reasonable). However, I'm wondering *why* the Xvfb process tries to write something to the pipe after the display name has been written already. In the CI log, I see that the display number has been successfully read from the pipe, so actually there's no need to write something more to the pipe...

Locally, this problem does not arise....

t-wissmann:
For the record: With the `Xvfb` server, the following is the stdout output of the failing test:
```
Running: Xephyr -nolisten tcp -screen 500x450x8 -screen 300x400x8 +extension XINERAMA -extension RANDR -displayfd 26
Xephyr is using the display ":0"
----------------------------- Captured stderr call -----------------------------
_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
(EE) 
Fatal server error:
(EE) Cannot write display number to fd 26
(EE) 
herbstluftwm: XOpenDisplay() failed
```

t-wissmann:
I made a mistake in the loop that reads from the pipe. But I still can't explain why it caused the CI to fail originally. Now, it seems to work with the `os.close()`, so I think it's fine as it is.

blarz:
This looks good to me! 

The only thing I thought would make this PR even better if we'd use more f-strings instead of .format() in `MultiscreenDisplay`. But that is just my personal taste, you don't need to change that.

t-wissmann:
I always thought that f-strings are considered unclean, so I'm avoiding them in low-level code and only using them in high-level code (e.g. `hlwm.get_attr(f'clients.{winid}.floating)`)

blarz:
> I always thought that f-strings are considered unclean, so I'm avoiding them in low-level code and only using them in high-level code (e.g. `hlwm.get_attr(f'clients.{winid}.floating)`)

Ok, I'm no Python expert but I thought they were "the way to go".
But as I said, it's just my personal taste. I'll approve this change :)

PR #812, opened 2020-04-17
Make cycle_all traverse also floating clients

### LINKED ISSUE #732: cycle_all (Alt-Tab) should cycle through all windows

t-wissmann:
There is no need to adjust the description of the cycle_all command in
the man page because it is still/again true that cycle_all traverses all
windows and frames on a tag.

This fixes #732.

### DIFF
--- a/src/tag.cpp
+++ b/src/tag.cpp
@@ -298,8 +298,9 @@ int HSTag::cycleAllCommand(Input input, Output output)
         FrameTree::CycleDelta cdelta = (delta == 1)
                 ? FrameTree::CycleDelta::Next
                 : FrameTree::CycleDelta::Previous;
-        bool succeeded = frame->cycleAll(cdelta, skip_invisible);
-        if (!succeeded) {
+        bool focusChanged = frame->cycleAll(cdelta, skip_invisible);
+        if (!focusChanged) {
+            // if frame->cycleAll() reached the end of the tiling layer
             if (floating_clients_.empty()) {
                 // we need to wrap. when cycling forward, we wrap to the beginning
                 FrameTree::CycleDelta rewind = (delta == 1)
@@ -310,10 +311,10 @@ int HSTag::cycleAllCommand(Input input, Output output)
                 // if there are floating clients, switch to the floating layer
                 floating_focused = true;
                 if (delta == 1) {
-                    // wrap to last client
+                    // wrap (forward) to first client
                     floating_clients_focus_ = 0;
                 } else {
-                    // wrap to last client
+                    // wrap (backward) to last client
                     floating_clients_focus_ = floating_clients_.size() - 1;
                 }
             }
--- a/src/tag.cpp
+++ b/src/tag.cpp
@@ -282,16 +282,46 @@ int HSTag::cycleAllCommand(Input input, Output output)
     if (delta == 0) {
         return 0; // nothing to do
     }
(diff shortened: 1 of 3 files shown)


src/tag.cpp
```diff
@@ -282,16 +282,46 @@ int HSTag::cycleAllCommand(Input input, Output output)
...
+                // we need to wrap. when cycling forward, we wrap to the beginning
+                FrameTree::CycleDelta rewind = (delta == 1)
+                            ? FrameTree::CycleDelta::Begin
+                            : FrameTree::CycleDelta::End;
+                frame->cycleAll(rewind, skip_invisible);
+            } else {
+                // if there are floating clients, switch to the floating layer
+                floating_focused = true;
+                if (delta == 1) {
+                    // wrap to last client
```
blarz, line 313: first?

tests/test_layout.py
```diff
@@ -299,6 +299,24 @@ def test_cycle_all_empty_frame(hlwm, delta):
     assert layout == hlwm.call('dump').stdout
 
 
+@pytest.mark.parametrize("floating_clients", [1, 2])
+@pytest.mark.parametrize("delta", [1, -1])
+def test_cycle_all_with_floating_clients(hlwm, delta, floating_clients):
```
blarz, line 304: "all" is only floating clients here, right? Should we add non-floating parameterization, too, just to be sure?

src/tag.cpp
```diff
@@ -282,16 +282,46 @@ int HSTag::cycleAllCommand(Input input, Output output)
...
+        } else {
+            floating_clients_focus_ = static_cast<size_t>(newIndex);
+            floating_clients_[floating_clients_focus_]->raise();
+        }
+    } else {
+        FrameTree::CycleDelta cdelta = (delta == 1)
+                ? FrameTree::CycleDelta::Next
+                : FrameTree::CycleDelta::Previous;
+        bool succeeded = frame->cycleAll(cdelta, skip_invisible);
+        if (!succeeded) {
```
blarz, line 302: Add a comment here, what this means if cycleAll did not succeed? At least I had to think for a second what it means.
"End of tiling clients list, switch to float or rewind".
Or maybe a different name for the result would make it clearer.

tests/test_layout.py
t-wissmann: I think this is mostly covered because we have the empty frame here (the `None`). We can also add another testcase that mixes floating and tiling clients, but I'm not sure whether it tests something meaningful or just burns cpu time in the CI.

src/tag.cpp
t-wissmann: Is `focusChanged` clearer?

src/tag.cpp
t-wissmann: yes

tests/test_layout.py
t-wissmann: I've added a new test case `test_cycle_all_traverses_tiling_and_floating` to cover multiple tiling and floating clients.

blarz: Awesome, thank you!

src/tag.cpp
blarz: It is indeed, thank you!

tests/test_layout.py
blarz: Thank you!

PR #816, opened 2020-04-17
Make sure to init all builtins, pointers, records

dnnr:


src/framedecoration.h
```diff
@@ -15,7 +15,7 @@ class FrameDecorationData {
     bool visible = false;
     bool hasClients = false; // whether this frame holds clients at the moment
     bool hasParent = false;
-    Rectangle geometry;
+    Rectangle geometry = {};
```
t-wissmann, line 18: Wouldn't it be easier to add a default constructor in `Rectangle`?

dnnr: Not sure! I tried (just pushed) it, and curiously, it leads to `SnapData` needing its members initialized as well, because `Rectangle` is not a record type anymore.

These rules sure are funny.

src/frametree.cpp
```diff
@@ -820,12 +832,9 @@ int FrameTree::splitCommand(Input input, Output output)
...
     auto availableModes = SplitMode::modes(align_explode, align_auto);
-    SplitMode m;
-    for (auto &it : availableModes) {
-        if (it.name[0] == splitType[0]) {
-            m = it;
-        }
-    }
+    SplitMode m = *std::find_if(
+            availableModes.begin(), availableModes.end(),
+            [=](const SplitMode &x){ return x.name[0] == splitType[0]; });
```
t-wissmann, line 837: What happens here if `std::find_if` returns `availableModes.end()`?

t-wissmann:
This looks very good, thanks! It's really satisfying to know that all memory is really initialized before it's used :-) (there were bugs once due to insufficient initialization)

src/frametree.cpp
dnnr: Ooh, good catch! I've added a test for this in #825, which would have caught this blunder.

dnnr: (Oh and, in case your question wasn't rhetorical: it actually crashes...)

t-wissmann: Ah, I strongly suspected that but wasn't sure :-)

PR #822, opened 2020-04-18
Switch to bionic as Travis build environment

dnnr:


dnnr:
> Shouldn't matter much since the most happens in the docker image anyway, right?

Correct. The primary reason is to stay up to date and not be surprised by a depreciation of the xenial base image.

PR #823, opened 2020-04-18
Remove git from Docker images used for CI builds

dnnr:
This used to be necessary, but it is not anymore. Leaving it out, should
speed up the build a bit, due to smaller image sizes.


blarz:
Why is it a draft?

dnnr:
> Why is it a draft?

Just wanted to make sure the build passes

PR #824, opened 2020-04-18
Fix complaints by -Wnull-dereference, enable in CI

dnnr:


src/mousedraghandler.cpp
```diff
@@ -253,11 +253,18 @@ MouseResizeFrame::MouseResizeFrame(MonitorManager *monitors, shared_ptr<HSFrameL
...
+    if (!neighbour) {
         throw DragNotPossible("No neighbour frame in the direction of the cursor");
     }
+
     dragFrame_ = neighbour->getParent();
     auto df = dragFrame_.lock();
+
+    if (df == nullptr) {
+        // We need to check this explicitly because of -Wnull-dereference
+        throw DragNotPossible("No neighbour frame in the direction of the cursor");
```
blarz, line 265: That message isn't quite correct is it? It is the neighbours parent that is failing here?

src/frametree.cpp
```diff
@@ -475,8 +475,10 @@ bool FrameTree::focusInDirection(Direction direction, bool externalOnly)
     shared_ptr<HSFrame> neighbour = curframe->neighbour(direction);
     if (neighbour) { // if neighbour was found
         shared_ptr<HSFrameSplit> parent = neighbour->getParent();
-        // alter focus (from 0 to 1, from 1 to 0)
-        parent->swapSelection();
+        if (parent) {
+            // alter focus (from 0 to 1, from 1 to 0)
+            parent->swapSelection();
+        }
         return true;
```
blarz, line 482: Is this correct even in the case where !parent? Or should the return true be inside the if(parent) block?

t-wissmann: If it has a neighbour, the neighbour must have a parent. So the only purpose of `if (parent)` is to silence the compiler warning.

src/mousedraghandler.cpp
t-wissmann: This shouldn't happen anyway. The message is printed to the user by the `drag` command, so the only more helpful message is something like: `"The neighbour frame has no parent, this is a bug, please report"`.

src/frametree.cpp
blarz: Ok, just wanted to be sure. Thanks for your explanation!

src/mousedraghandler.cpp
dnnr: That makes sense, I've changed the message.

blarz: @dnnr just updated the message, resolving this.

blarz:
> Looks good to me, this was the last -W exception, very nice!

Ah, it is not the last one (there is `no-sign-compare`), but one that bothered me quite much :D

PR #825, opened 2020-04-18
Add test for invalid argument to "split" command

dnnr:


t-wissmann:
Thanks for the testcase. I've updated the whitespace btw to satisfy flake8

dnnr:
> Thanks for the testcase. I've updated the whitespace btw to satisfy flake8

Thanks! My vim-integrated Python linting seems to be broken...

PR #828, opened 2020-04-18
Fix 'bring' if the client is on the same tag

### LINKED ISSUE #818: Command 'bring' is broken
Bring works in 7.2 but not 8.0 or current git

blarz:
Basically reverts changes from 41ec3f1d7afcdaabacf6a9e46a8b141a8f9e96c5
but considers floating clients, too. Add some tests, of course.

Fixes #818

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -439,9 +439,8 @@ int frame_current_bring(int argc, char** argv, Output output) {
     auto frame = tag->frame->root_->frameWithClient(client);
     if (!client->is_client_floated() && !frame->isFocused()) {
         frame->removeClient(client);
-        tag->frame->focusedFrame()->insertClient(client);
+        tag->frame->focusedFrame()->insertClient(client, true);
     }
-    focus_client(client, false, false, true);
     return 0;
 }
 
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -436,6 +436,12 @@ int frame_current_bring(int argc, char** argv, Output output) {
     }
     HSTag* tag = get_current_monitor()->tag;
     global_tags->moveClient(client, tag, {}, true);
+    auto frame = tag->frame->root_->frameWithClient(client);
+    if (!client->is_client_floated() && !frame->isFocused()) {
+        frame->removeClient(client);
+        tag->frame->focusedFrame()->insertClient(client);
+    }
+    focus_client(client, false, false, true);
     return 0;
 }
 
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -441,6 +441,7 @@ int frame_current_bring(int argc, char** argv, Output output) {
         frame->removeClient(client);
         tag->frame->focusedFrame()->insertClient(client, true);
     }
+    focus_client(client, false, false, true);
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


blarz:
Not sure if I handle the floating clients correctly, seems to work though (added a test).
Also not sure if I added the test cases to the correct suites.

src/layout.cpp
```diff
@@ -436,6 +436,12 @@ int frame_current_bring(int argc, char** argv, Output output) {
     }
     HSTag* tag = get_current_monitor()->tag;
     global_tags->moveClient(client, tag, {}, true);
+    auto frame = tag->frame->root_->frameWithClient(client);
+    if (!client->is_client_floated() && !frame->isFocused()) {
+        frame->removeClient(client);
+        tag->frame->focusedFrame()->insertClient(client);
+    }
+    focus_client(client, false, false, true);
```
t-wissmann, line 444: Instead of `focus_client()` I think it's enough to pass `focus = true` to `insertClient()`.

t-wissmann: I hope it was OK that I've implemented this directly.

t-wissmann:
> Not sure if I handle the floating clients correctly, seems to work though (added a test).
> Also not sure if I added the test cases to the correct suites.

Thanks for fixing the issue! I usually put all test cases that test a command together in the same file. So all test cases that test `bring` should be in the same file. Similarly, all test cases that test the `load` command also should be in the same file. Since there are test cases that use the `load` command to test something else or some other command, there are test cases in other files to that mention `load`. (Maybe this confused you?)

But I don't think we should live this so strict (and I don't care too much). the most important think is thtat if one passes `-k test_bring` to pytest, that then the `bring` command is tested :-)

src/layout.cpp
blarz: This does seem to break it for me again. The brought client does not get focus and is still dislayed in a different frame. If I cycle through them, then it's frame gets updated. I'll remove your change again and update the test accordingly.

src/layout.cpp
```diff
@@ -436,6 +436,11 @@ int frame_current_bring(int argc, char** argv, Output output) {
     }
     HSTag* tag = get_current_monitor()->tag;
     global_tags->moveClient(client, tag, {}, true);
+    auto frame = tag->frame->root_->frameWithClient(client);
+    if (!client->is_client_floated() && !frame->isFocused()) {
+        frame->removeClient(client);
+        tag->frame->focusedFrame()->insertClient(client, true);
+    }
```
blarz, line 443: We need to add the focusClient back

t-wissmann: I think I only missed `tag->needsRelayout_.emit()`. I'll try it in a separate PR.

blarz:
> > Not sure if I handle the floating clients correctly, seems to work though (added a test).
> > Also not sure if I added the test cases to the correct suites.
> 
> Thanks for fixing the issue! I usually put all test cases that test a command together in the same file. So all test cases that test `bring` should be in the same file. Similarly, all test cases that test the `load` command also should be in the same file. Since there are test cases that use the `load` command to test something else or some other command, there are test cases in other files to that mention `load`. (Maybe this confused you?)
> 
> But I don't think we should live this so strict (and I don't care too much). the most important think is thtat if one passes `-k test_bring` to pytest, that then the `bring` command is tested :-)

I think what made me split was that the `bring` command was tested with two different "client types", tiling and floating, and I thought it would fit better to test floating stuff in the floating "suite".

PR #829, opened 2020-04-18
Enable sanitizers in CI builds again

blarz:


blarz:
I know this caused issues in #559, so I'd be ok with this PR being rejected ;)

t-wissmann:
So what's the advantage of these sanitizers? ;-)

blarz:
> So what's the advantage of these sanitizers? ;-)

Find more bugs!

- Running our CI build tests with ASAN enabled and experience no crashes, indicates that we do handle pointers and memory quite well ;)
- LSAN already detected some minor leaks (see #706). Although I thought there were plans to run tests within valgrind (that could make running tests quite  slow, LSAN has less runtime overhead AFAIK)

I don't want to rush this in. Maybe we could have some example PRs demoing LSAN and ASAN functionality first.

blarz:
BTW, I am running my hlwm git build on my laptop with sanitizers enabled in day to day use ;)

t-wissmann:
Of course, I also want to find bugs as soon as possible. However, after #559 it didn't give me much confidence, and I don't want to make the same mistakes again.

Maybe we can pick one of the bugs ([this was the fatal](https://github.com/herbstluftwm/herbstluftwm/pull/559/commits/146a0aa4b0e1482d219e5001687003fe54312c89)) and see whether it is still hidden by the sanatizer. (Maybe it's now discovered by the new clang-tidy checks?) 

By the way the valgrind PR #561 was never merged because there were further race-conditions discovered, but I'm not sure whether they are in the herbstluftwm code or only in the test cases.

dnnr:
I want the sanitizers, but I also agree that the bug masking thing was nasty. My memory is foggy here, but I think I tried a lot to find out *how* the bug was masked, and didn't get anywhere. The best case would be if we could find the reasons and fix them (maybe via some ASAN options). The second-best would be, as Thorsten said, ensuring that we'd now find the same bug through different tools.

blarz:
Making this a draft PR for now, as it is exactly that: a PR that sparks a discussion ;)

blarz:
This really seems to trigger a LSAN warning with monitor detection code.
Can't figure out where, though.

t-wissmann:
The relevant output is:
```
----------------------------- Captured stdout call -----------------------------
Running: Xephyr -nolisten tcp -screen 500x450x8 -screen 300x400x8 +extension XINERAMA -extension RANDR -displayfd 29
Xephyr is using the display ":0"
hlwm started
Client command ['chain', ',', 'add', 'tag1', ',', 'add', 'tag2', ',', 'add', 'tag3'] succeeded (no output)
Client command ['get_attr', 'monitors.count'] succeeded with output:
1
Client command ['detect_monitors', '--no-disjoin'] succeeded (no output)
Client command ['detect_monitors', '--list-all'] succeeded with output:
xrandr:
xinerama: 500x450+0+0 300x400+500+0
Client command ['get_attr', 'monitors.count'] succeeded with output:
2
Client command ['monitor_rect', '0'] succeeded with output:
0 0 500 450
Client command ['monitor_rect', '1'] succeeded with output:
500 0 300 400
----------------------------- Captured stderr call -----------------------------
_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
../src/monitordetection.cpp: 55: no xrandr available
../src/monitordetection.cpp: 55: no xrandr available
../src/main.cpp: 466: Interrupted by signal 15
.Cannot listen for hooks
./src/xconnection.cpp: 35: Closing display
=================================================================
==18639==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f482a61aae8 in malloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dae8)
    #1 0x7f482a3aa9eb  (/lib/x86_64-linux-gnu/libXrandr.so.2+0x29eb)
SUMMARY: AddressSanitizer: 32 byte(s) leaked in 1 allocation(s).
=========================== short test summary info ============================
FAILED tests/test_monitor_detection.py::test_detect_monitors_xinerama[Xvfb-screens0]
```
Interestingly, hlwm does not appear in the call stack of the malloc of the leaked memory...

blarz:
The leak seems to be in xrandr and there is not much we can do about it, besides making LSAN ignore it (as do other [projects like chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=119677)).

In the `test_monitor_detection.py` we use the `MultiscreenDisplay` fixture, which enables xinerama and disables xrandr for Xvfb/Xephyr. So the only interaction with xrandr is the query of `XRRQueryExtension` in https://github.com/herbstluftwm/herbstluftwm/blob/master/src/monitordetection.cpp#L49 to check if xrandr is even available.

I suspect that the call to [`XRRFindDisplay`](https://gitlab.freedesktop.org/xorg/lib/libxrandr/-/blob/master/src/Xrandr.c#L352) creates a leak, at least I see some XMalloc https://gitlab.freedesktop.org/xorg/lib/libxrandr/-/blob/master/src/Xrandr.c#L300 that is probably not freed?
That file hasn't been touched in years, the only recent commits to xrandr are the ones about moving to Gitlab ;) The last relevant commit was 2017, [fixing a memory leak](https://gitlab.freedesktop.org/xorg/lib/libxrandr/-/commit/87227e5fc79750d3eccc3c3482a3c5b3f2af2e90)

**tl;dr** xrandr leaks, we should ignore it in LSAN, [chromium does it too](https://chromium.googlesource.com/external/webrtc/+/d7de1209ae60c2e38e38aced8864589b1f2891c5/tools/lsan/suppressions.txt#29). I just didn't get [the suppression to work for the tests](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#suppressions). I'll look into that, but I think it's just a problem in regards to passing env variables to the right processes.

tests/conftest.py
```diff
@@ -21,6 +21,16 @@
 
 BINDIR = os.path.join(os.path.abspath(os.environ['PWD']))
 
+# List of environment variables copied during hlwm process creation:
+# * LSAN_OPTIONS: needed to suppress warnings about known memory leaks
+COPY_ENV_WHITELIST = ['LSAN_OPTIONS']
+
+
+def extend_env_with_whitelist(environment):
```
blarz, line 29: Should I extract this into a separate PR?

ci/build.py
```diff
@@ -124,6 +124,9 @@ def tox(tox_args, build_dir):
     tox('-e flake8', build_dir)
 
 if args.run_tests:
+    # Suppress warnings about known memory leaks:
+    os.environ['LSAN_OPTIONS'] = f"suppressions={repo}/ci/lsan-suppressions.txt"
```
blarz, line 128: Is this the best place to set this variable?

blarz:
PR runs green now, but I had to jump through some hoops to get the LSAN_OPTIONS environment variable to every place I need.

tests/conftest.py
t-wissmann: It's fine as it is! :)

ci/build.py
t-wissmann: fine for me.

t-wissmann:
Great, thanks! I'm happy to merge this PR (Is it on purpose that this PR is still in 'draft' state?)

tests/conftest.py
blarz: Alright, thanks! :)

ci/build.py
blarz: Discussed it with @dnnr too, should be fine.

blarz:
> Great, thanks! I'm happy to merge this PR (Is it on purpose that this PR is still in 'draft' state?)

No, just wanted to discuss the two remaining questions I had.
Now they're resolved, I'll unWIP it. Thanks!

blarz:
Just testing some commit message related stuff, ignore the wip label ;)

Issue #815, opened 2020-04-17
0.8.0 panel detection config option?


ZoeFiri:
I use lemonbar and my bar is configured to show at the bottom of my screen which I have padded for it, although unfortunately the new 0.8.0 update's panel detection feature appears to not be aware of where the bar is located which creates a large pad at the top of my screen. This is increasingly problematic since I use lemonbar to create visual notifications which only last for a couple seconds, but will pad the top of my screen even after they're gone. Is there a way to simply disable this? I've included an image for reference.
![image](https://user-images.githubusercontent.com/27149384/79593919-c4ea4980-80a1-11ea-8ccc-1208ee86f258.png)

ypnos:
So are these temporary panels that show notifications located at the top or at the bottom?
In other words, is the padding correct for these notifications?

Apart from that it looks like two/three issues:

- Automatic padding is always applied, even if padding was previously set manually
- No option to disable automatic padding
- Unregister seems not to work for your panel

As a "simple" workaround right now what you can do is use herbstclient to reset the padding in the script that shows your notification (after the panel appearance).

t-wissmann:
With the lemonbar here, `lemonbar -b` correctly sets up the pad at the bottom.
Do you have your lemonbar script somewhere public such that I can try it? If not, how are you calling lemonbar? with `-b` or simply with the geometry?

I think it's reasonable to add a setting that allows disabling the panel detection, just to be on the save side.

Nice bar by the way :-) I didn't recognize that it's `lemonbar`!

t-wissmann:
By the way, lemonbar also provides an option `-d` with which you can disable the panel detection.

For completeness, I also added a setting for controlling the detection in the PR #832.

ZoeFiri:
@t-wissmann haha thanks, it's kinda a hacky combination of a lot of lemonbar instances, which may be why it doesn't allocate things right. I'm calling lemonbar with geometry for more finely tuned control.
@ypnos neither. the "notification" shows up briefly at the center of my screen and disappears shortly after. I'll try the hc pad after workaround, thanks.

t-wissmann:
> @t-wissmann haha thanks, it's kinda a hacky combination of a lot of lemonbar instances, which may be why it doesn't allocate things right. I'm calling lemonbar with geometry for more finely tuned control.

[...]

PR #832, opened 2020-04-19
New setting 'auto_detect_panels'

t-wissmann:
This helps working around other problems with panels, such as the one in
issue #815.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1074,7 +1074,7 @@ auto_detect_monitors (Boolean)::
     If set, detect_monitors is automatically executed every time a monitor is
     connected, disconnected or resized.
 
-auto_detect_monitors (Boolean)::
+auto_detect_panels (Boolean)::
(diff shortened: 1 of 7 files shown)


doc/herbstluftwm.txt
```diff
@@ -1074,6 +1074,11 @@ auto_detect_monitors (Boolean)::
     If set, detect_monitors is automatically executed every time a monitor is
     connected, disconnected or resized.
 
+auto_detect_monitors (Boolean)::
```
blarz, line 1077: _panels?

t-wissmann: Thanks!

blarz:
Doesn't seem to break anything for me, with the default panel and the default setting.
Any other ways to test this?

t-wissmann:
> Doesn't seem to break anything for me, with the default panel and the default setting.
> Any other ways to test this?

We don't have test cases for the panel detection at all (we should have some in the long run). I also tested it locally with `lxpanel` within a Xephyr session, and it seems to be OK.

PR #834, opened 2020-04-19
Don't try this at home: not initializing Decoration members

t-wissmann:
Let us see what the CI detects.


t-wissmann:
Clang-tidy detects that the member variables aren't initialized, so I'm happy :-)

PR #835, opened 2020-04-19
Enable clang-tidy check for null function pointers

blarz:
See

 https://clang.llvm.org/docs/analyzer/checkers.html#core-callandmessage

for more info.

Fixes the following (unlikely) situation:

/home/simon/git/herbstluftwm/src/rootcommands.cpp:108:18: error: Called C++ object pointer is null [clang-analyzer-core.CallAndMessage,-warnings-as-errors]
    auto child = root->child(attr_path.first);
                 ^
/home/simon/git/herbstluftwm/src/rootcommands.cpp:72:9: note: Assuming the condition is false
    if (!p.empty()) {
        ^
/home/simon/git/herbstluftwm/src/rootcommands.cpp:72:5: note: Taking false branch
    if (!p.empty()) {
    ^
/home/simon/git/herbstluftwm/src/rootcommands.cpp:78:9: note: Assuming 'o' is null
    if (o && new_value.empty()) {
        ^
/home/simon/git/herbstluftwm/src/rootcommands.cpp:78:11: note: Left side of '&&' is false
    if (o && new_value.empty()) {
          ^
/home/simon/git/herbstluftwm/src/rootcommands.cpp:83:20: note: Calling 'RootCommands::getAttribute'
    Attribute* a = getAttribute(path, output);
                   ^
/home/simon/git/herbstluftwm/src/rootcommands.cpp:108:18: note: Called C++ object pointer is null
    auto child = root->child(attr_path.first);


blarz:
I know it is esoteric.. Not sure if we want this.

t-wissmann:
> I know it is esoteric.. Not sure if we want this.

Yes it is... I'm not quite convinced by the diff. Does it help if we change the `Object* root` into a `Object& root`?

dnnr:
I really don't get why the check triggers here. I'd expect that putting the nullptr check into the constructor helps as well, but it doesn't.

The good news is: Turning `root` into a reference seems to help, and it is more correct™ anyway.

blarz:
Thanks! Made it a reference.

src/rootcommands.h
```diff
@@ -23,7 +23,7 @@ class RootCommands {
     /** This object has no dependencies to other hlwm modules, and so the
      * 'root' pointer held by this class is just an Object-pointer.
```
blarz, line 24: Maybe this should be updated, too

dnnr: This also makes it sound like the causalities are reversed? I'd more describe it like "This module shall have minimal dependencies, therefore the reference to root just has the Object type instead of Root."

t-wissmann: Both directions are 'correct': rootcommands has no dependencies, so Object suffices, and it was designed that way (on purpose).
Yes, please update it while you're at it anyway. :)

blarz: Thanks! I updated the comment based on the suggestion by @dnnr

PR #839, opened 2020-04-19
Update fish completion script

### LINKED ISSUE #797: Fish completion inserts spaces for partial completions
In the fish shell, typing `herbstclient attr clien`<kbd>TAB</kbd> completes `clien` to `clients. ` with a space at the end. The problem is, that the fish completion file uses `complete` instead of `complete_shell`. Only `complete_shell` makes a distinction between 'full' and 'partial' argument suggestions.

However, I couldn't find out how to use `complete_shell` in the fish completion config, because I couldn't find out how to make the above mentioned distinction in a fish completion function.

ypnos:
Avoid spaces for partial completions. Fixes #797 with a workaround.
Also only complete herbstclient options where appropriate.

I would appreciate any feedback on the changes.

A small present for 0.8 release :smile_cat:

### DIFF
--- a/share/herbstclient.fish
+++ b/share/herbstclient.fish
@@ -1,19 +1,23 @@
 # fish completion for herbstclient
 
-function _get_herbstluftwm_completion
-    set -l tokens (commandline -op)
+function _first_hlwm_command
+    # Find first hlwm command in given tokens (e.g. commandline -o)
+    # Print token position if found; exit status 0 if found, else 1
 
-    # Delete "herbstclient"
-    set -e tokens[1]
-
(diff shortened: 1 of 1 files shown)


t-wissmann:
Thanks a lot! Here, it works great and is a great improvement over the fish completion we had before!

I think I can understand the completion function, but I didn't spot it where the hacky part is. (I've only noticed that `herbstclient rule windowtype=`<kbd>Tab</kbd> inserts a space instead of doing nothing as in bash/zsh, but I think we shouldn't care at all about such a minor thing).

PR #840, opened 2020-04-20
Never compile a regex for the empty string

t-wissmann:
An empty string is not allowed in the Posix grammar for extended regular
expressions. On openbsd, this led to a run time error on startup
because we initialize the keymask regexes with the empty string.

In the keymask code, we are using the empty string anyway as a special
case for an 'inactive' or 'unset' regular expression, so there is no
need to compile the regex for "".

Thanks to Lucas <lucas@sexy.is> and Florian Viehweger
<openbsd@out-of-creativity.de> for spotting and reporting this issue.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -13,6 +13,9 @@ Current git version
   * keybind now checks that the bound command exists.
   * cycle_all (Alt-Tab) now also traverses floating clients
   * new setting 'auto_detect_panels' controlling the panel detection algorithm.
+  * Bug fixes:
+    - Handle the regular expression "" correctly, since the grammar of extended
+      regular expressions does not allow "".
 
 Release 0.8.0 on 2020-04-09
 ---------------------------
--- a/src/keymanager.cpp
+++ b/src/keymanager.cpp
@@ -257,6 +257,8 @@ KeyManager::KeyMask::KeyMask()
 bool KeyManager::KeyMask::allowsBinding(const KeyCombo &combo) const
 {
     if (regex_.empty()) {
+        // an unset keymask allows every binding, regardless of
+        // the 'negated_' flag
         return true;
     } else {
         bool match = regex_.matches(combo.str());
--- a/src/regexstr.cpp
+++ b/src/regexstr.cpp
@@ -6,14 +6,20 @@ RegexStr::RegexStr()
 {
 }
 
-RegexStr RegexStr::fromStr(const string &source)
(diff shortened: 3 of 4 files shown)


NEWS
```diff
@@ -13,6 +13,9 @@ Current git version
   * keybind now checks that the bound command exists.
   * cycle_all (Alt-Tab) now also traverses floating clients
   * new setting 'auto_detect_panels' controlling the panel detection algorithm.
+  * Bug fixes:
+    - Handle the regular expression "" correctly, since the gramar extended
```
blarz, line 17: grammar

t-wissmann:
I don't know how to test it, except by adding a openbsd build to the CI.

blarz:
> I don't know how to test it, except by adding a openbsd build to the CI.

But you could add a test testing that you get an error when trying to do something with an empty regex?

t-wissmann:
> But you could add a test testing that you get an error when trying to do something with an empty regex?

I don't understand. The issue on openbox was that `std::regex("", …)` throws an exception. Obviously this does not happen in our CI build because otherwise *all* test cases would fail. So I don't see what I can test with our current CI setup. (Note that e.g. the regex `keymask` is `""` per default, so we've been checking the empty regex all the time)

blarz:
> > But you could add a test testing that you get an error when trying to do something with an empty regex?
> 
> I don't understand. The issue on openbox was that `std::regex("", …)` throws an exception. Obviously this does not happen in our CI build because otherwise _all_ test cases would fail. So I don't see what I can test with our current CI setup. (Note that e.g. the regex `keymask` is `""` per default, so we've been checking the empty regex all the time)

Ah ok, sorry I got that wrong then, nevermind :)

t-wissmann:
> Is it still wip?

I wanted to wait until tomorrow for the OpenBSD-guys to react. Apart from this, the PR is ready to be approved/merged IMO

blarz:
> > Is it still wip?
> 
> I wanted to wait until tomorrow for the OpenBSD-guys to react. Apart from this, the PR is ready to be approved/merged IMO

Ok, let's wait then. Other than that this looks good to go.

The-Compiler:
Looks like they're happy: [BUG: An empty regex is not allowed in the POSIX grammar.](https://lists.schokokeks.org/pipermail/hlwm/2020-April/000306.html)

PR #841, opened 2020-04-20
'resize' command: make delta parameter optional

t-wissmann:
Recently the 'resize' command became also functional in the floating
mode, where the 'FRACTIONDELTA' parameter is not used at all. In the
tiling mode, the parameter is most likely '0.02' for most people. Hence,
it makes sense to make the parameter optional, so one can now run e.g.

   herbstclient resize up

to make the focused client (or frame) grow into the upward direction.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,7 +9,8 @@ Current git version
   * new setting 'hide_covered_windows' to improve the appearance when used with
     a compositor.
   * resize floating windows with the same command ('resize') as in tiling mode
-    and thus the same keybindings as in tiling mode.
+    and thus the same keybindings as in tiling mode. Therefore, the
+    'fractiondelta' parameter to the 'resize' command is now optional.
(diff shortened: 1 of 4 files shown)


blarz:
In the default autostart, resizestep is 0.05 FWIW

t-wissmann:
Oh, right. Hm any opinions on that?
My impression is that `0.05` is quite (too?) coarse..

blarz:
> Oh, right. Hm any opinions on that?
> My impression is that `0.05` is quite (too?) coarse..

We could change the default.
It is only a one line change to get the old delta back if users want it more coarse.

t-wissmann:
> > Oh, right. Hm any opinions on that?
> > My impression is that `0.05` is quite (too?) coarse..
> 
> We could change the default.
> It is only a one line change to get the old delta back if users want it more coarse.

I've updated it as a separate PR in #842.

PR #850, opened 2020-04-21
MouseManager: do not print empty debug message

t-wissmann:
If there is no error message, do not print a debug message


src/mousemanager.cpp
```diff
@@ -122,7 +122,9 @@ bool MouseManager::mouse_handle_event(unsigned int modifiers, unsigned int butto
         return true;
     }
     string errorMsg = (this ->* (b->action))(client, b->cmd);
-    HSDebug("can not start drag: %s\n", errorMsg.c_str());
+    if (!errorMsg.empty()) {
+        HSDebug("can not start drag: %s\n", errorMsg.c_str());
```
dnnr, line 126: While at it, you could change this to `Cannot` ;-)

PR #852, opened 2020-04-22
Mergify: Take commit message from first PR body

t-wissmann:
Usually, when handing in a pull request, it consists of one commit with
a meaningful commit message. Then smaller commits with very short commit
titles like "Update comment", "Moar", "Remove whitespace" follow. Usually,
these messages do not have any meaning in the squashed commit that is
created when merging the PR. So they don't belong to the commit message.
Instead, with "commit_message: title+body" we can actively edit the
commit message before the PR is merged.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -9,3 +9,4 @@ pull_request_rules:
           merge:
               method: squash
               strict: smart
+              commit_message: title+body


t-wissmann:
See https://doc.mergify.io/actions.html

Unfortunately, I don't know whether this wraps the commit message to the usual width of git commit messages or whether the mark down source code of the comment is taken as-is.

dnnr:
I'm confused. The mergify docs state that _`title+body` means to use the title and body from the pull request itself as the commit message._ That does not sound like how you describe it.

The way I'm reading this is that for us, the only difference between `title+body` and `default` is whether to keep the PR number at the end of the first line (which is put there by GitHub).

dnnr:
Maybe we should try this in a personal fork of the repo first?

t-wissmann:
When you open a PR for a branch with only one new commit, then it's commit message is used for the PR body. If somebody opens a PR for a branch with more than one new commit, I expect the PR author to write a meaningful PR message body.

t-wissmann:
> Maybe we should try this in a personal fork of the repo first?

Maybe yes. Do you have mergify&travis set up already?

dnnr:
> When you open a PR for a branch with only one new commit, then it's commit message is used for the PR body. If somebody opens a PR for a branch with more than one new commit, I expect the PR author to write a meaningful PR message body.

Yeah, but that process is happening during the _opening_ of the PR (and maybe subsequent modification), not during mergify's part in _merging_ the PR.

Yup, I can take a look at it this evening.

t-wissmann:
> > When you open a PR for a branch with only one new commit, then it's commit message is used for the PR body. If somebody opens a PR for a branch with more than one new commit, I expect the PR author to write a meaningful PR message body.
> 
> Yeah, but that process is happening during the _opening_ of the PR (and maybe subsequent modification), not during mergify's part in _merging_ the PR.

Yes, that's precisely what I want to achive.

  1. Open a PR for one commit with a meaningful message
  2. Update the PR _and the first comment of the PR_ until the referees and the CI are happy
  3. As soon as it's approvied (not only the changeset but also the commit
  message!) Mergify makes a commit for the approved changes and with the approved commit message.

Or am I missing something here?

My plain wish is: I want to prevent meaningless/redundant commit message bodies like these:
ac69b438f44b5b8a6137fcd1bc2c07b47d9b6332 f5b301dd332a795c97666d6c23e17d22873505c4 3f3e60a73536947984a52de20d4baec66b824660 fe33b15577ab4ec55c4c05618cae3935ae4f4657 7974b7ac96dc141caa34323258e1621767a08c0e be75e086a08a354e807943d9171c26a2cc7d0f54

If you are right and the only change of this PR is that the PR-number is ommitted from the first line, I don't see what's so dangerous about trying it in the main repo ;-)

dnnr:
There is not much danger besides some messed up commits. I'd just prefer doing such experiments outside of master. ;-) Beware that you'll probably need to merge any config changes first and *then* verify them on a new PR.

Okay, if the branch shortlog being added to the squash message is the thing you want to get rid of, I fully agree. I wasn't even aware of that anymore. If I had, I'd have always squashed my branches before getting them merged. (Maybe I even did that last year but simply forgot about it.)

That said, I wouldn't underestimate the value of having the PR number in the commit message. I regularly make use of that back-reference, and I don't think there's a real alternative available. Maybe `title+body` even retains the PR number, it's not really clear. We should try!

t-wissmann:
I've tried it on https://github.com/t-wissmann/herbstluftwm/pull/1 and it seems that the PR number is ommitted from the commit message title. I wonder why this is the case, because I can't find anything about this in the documentation: https://doc.mergify.io/actions.html#actions does not even mention the word "number".

dnnr:
As an alternative approach, we could try to make use of this feature: https://doc.mergify.io/actions.html#commit-message

t-wissmann:
> As an alternative approach, we could try to make use of this feature: ~https://doc.mergify.io/actions.html#commit-message~ anchoring is broken, I'm referring to the `## Commit Message` trick in the description markdown

I've already tried that in https://github.com/t-wissmann/herbstluftwm/pull/3, it also does not include the PR number..

dnnr:
Ugh. Well, another way would be to enforce a squashed branch as a CI check. Not sure if that's worth the hassle.

In my opinion, keeping the PR number in the message is a bit more important than getting rid of the annoying shortlog the body.

t-wissmann:
> Ugh. Well, another way would be to enforce a squashed branch as a CI check. Not sure if that's worth the hassle.
> 
> In my opinion, keeping the PR number in the message is a bit more important than getting rid of the annoying shortlog the body.

I agree (with both statements) :-( So I fear I have to close this PR and wait for better times to come...

PR #854, opened 2020-04-22
Update fish completions

ypnos:
Sorry my last PR was a bit buggy and I wasn't available while the release happened. I am still getting feedback to my [questions over at fish](https://github.com/fish-shell/fish-shell/issues/6922).

It works well for the typical use cases so I think it's not too bad.
However here are some fixes and improvements. Not a complete wrap yet for me. It might stay the way it is but I am still thinking about possible improvements.


t-wissmann:
Cool! Is this PR still work in progress or is it ready to be tested&merged?

ypnos:
I will let you know when I believe it is ready.

t-wissmann:
OK, take your time!

ypnos:
The stuff fish people write in the other thread is confusing/makes no sense but I also don't feel like spending more time discussing with them right now to eventually maybe find ways to improve corner cases.

So my work is done here for now.

t-wissmann:
> The stuff fish people write in the other thread is confusing/makes no sense but I also don't feel like spending more time discussing with them right now to eventually maybe find ways to improve corner cases.
> 
> So my work is done here for now.

Okey, thanks! I'll test it. Can I remove the 'WIP' from the title? (You can do so as well)

ypnos:
Sorry that didn't cross my mind. Is not WIP anymore.

t-wissmann:
Thanks! Currently, I'm a bit confused because I don't know how I can check that the new completion file is active. I've copied it to `/usr/share/fish/vendor_completions.d/` and restarted fish. So far it seems to work, thanks a lot!

Is there any command/completion that works now that didn't work before or does this pr simply clean up the fish completion code?

ypnos:
This is how you can swap out the completions:

```sh
complete -c herbstclient -e; and source herbstclient.fish
```

The changes of this PR are:

- A regression introduced in 73c64e7 would lead to false completions
when hc is completed in a chain of commands (e.g. `echo hi; and hc <tab>`)
- Another regression introduced in last PR would send herbstclient options to hlwm for competion
- Options to herbstclient are completed even if hlwm command is already there, in front of command, (e.g. `hc <tab> detect_monitors`); a slight improvement after previous PR introduced that they are _not_ offered where not allowed (e.g. `hc detect_monitors -<tab>`)
- hlwm commands are not offered for completion after options that don't take commands (-v|-h|--help|-w|--wait|-i|--idle)

t-wissmann:
> This is how you can swap out the completions:
> 
> ```shell
> complete -c herbstclient -e; and source herbstclient.fish
> ```

Alright, thanks!

> * Options to herbstclient are completed even if hlwm command is already there, in front of command, (e.g. `hc <tab> detect_monitors`); a slight improvement after previous PR introduced that they are _not_ offered where not allowed (e.g. `hc detect_monitors -<tab>`)

Here, `hc detect_monitors -<kbd>tab</kbd>` still lists them. (Which would be fine for me anyway..). Maybe I did something wrong:
[asciicast](https://asciinema.org/a/326370)

tiosgz:
> Maybe I did something wrong

I think it's your history being completed.

ypnos:
Unfortunately I can't explain this behavior. Here is how it looks for me:

fish

t-wissmann:
> > Maybe I did something wrong
> 
> I think it's your history being completed.

I'm not talking about the history completions (the gray stuff) but that `detect_monitors` has e.g. `-c`.

t-wissmann:
> Unfortunately I can't explain this behavior. Here is how it looks for me:
> 
> fish

Strange.. does it depend on the fish version or some other setting? I have fish version 3.1.0 here. But anyway, I don't care too much about this detail because it works overall (especially for the attribute paths). So I guess we should simply merge it if it's fine for you :-)  (maybe fish sources another file for me I'm not aware of...)

tiosgz:
> > > Maybe I did something wrong
> > 
> > 
> > I think it's your history being completed.
> 
> I'm not talking about the history completions (the gray stuff) but that `detect_monitors` has e.g. `-c`.

Sorry, I didn't realise it links to a video.

ypnos:
I use 3.1.1 but I would be surprised if there is a relation to fish version. However I am already surprised at your result. Does it work differently with 0743465775e9df0f6ec473dac2599acb9055f101?

t-wissmann:
> I use 3.1.1 but I would be surprised if there is a relation to fish version. However I am already surprised at your result. Does it work differently with [0743465](https://github.com/herbstluftwm/herbstluftwm/commit/0743465775e9df0f6ec473dac2599acb9055f101)?

It seems I'm getting inconsistent behaviour for [0743465], when I tried it first, only the detect_monitors flags were completed (as it should be). But on the second try, I had the issue again that also `-c` and the other herbstclient flags were suggested. Is fish caching/compiling the files when sourcing? (or I am confusing the files all the time....)

t-wissmann:
I'm afk for some time. My overall impression is that I'm still doing something wrong and we shouldn't waste too much time on it if it clearly improves the situation for you as a fish user knowing what your doing (compared to me source files that I don't really understand)

ypnos:
I agree that this is ready to be merged.

All completion definitions are additive, so if you don't clear out existing completions for a command they will be preserved even if sourcing a new file.

PR #855, opened 2020-04-22
Make sizehints client attributes writable

t-wissmann:
Make the client attributes sizehints_tiling and sizehints_floating
writable. Also add a testcase checking that these and similar attributes
are writable.


NEWS
```diff
@@ -9,6 +9,8 @@ Current git version
       passed. (Affects list_padding move_monitor rename_monitor lock_tag
       unlock_tag)
     - Fix the type of WM_STATE
+    - The cient attributes sizehints_tiling and sizehints_floating are now
```
blarz, line 12: *client

t-wissmann: [x] fixed, thanks!

PR #856, opened 2020-04-22
close_(and/or)_remove only close in floating mode

### LINKED ISSUE #853: close_and_remove closes wrong window for single window floating clients
Steps to reproduce:

- 2 windows in tiling mode

- Make one of them floating

- Execute "close_and_remove" on floating window closes the window in tiling, the floating window cannot be closed.

t-wissmann:
For floating clients (regardless of single window or the entire tag),
the commands close_and_remove and close_or_remove do not make sense
because either the window is in no frame (single window floating) or
because the frame is not visisble (tag is floated). Hence these commands
must behave like the ordinary 'close' if the focused client is floated.

On the development side, this causes that we need to move both commands
from the FrameTree to the HSTag class. This commit also adds test cases
to verify both the new behavior and the usual behavior in tiling mode.

This fixes #853.


t-wissmann:
> The floating window handling changes reach into every command ;)

That's why I feared it for so long.

PR #858, opened 2020-04-23
Enable theme setting tight_decoration again

t-wissmann:
The tight_decoration attribute specifies whether the window sizehints
affect the decoration window or only the window content itself. Other
than in the old code base of v0.7.2, we now always center the window if
the size hints change the window size.

This only affects tiled clients with sizehints_tiling enabled. A visual
demonstration of the tight_decoration setting can be found in the
pull-request for this commit:
https://github.com/herbstluftwm/herbstluftwm/pull/858

Effectively, this commit only removes a redundant "if (!false)" and
replaced one "if (false)" by "if (scheme.tight_decoration())". The rest
is documentation and tests.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -149,23 +149,25 @@ void Decoration::resize_outline(Rectangle outline, const DecorationScheme& schem
 
     auto tile = inner;
     client_->applysizehints(&inner.width, &inner.height);
-    if (!false) { // formely: if (!tight_decoration)
-        // center the window in the outline tile
-        // but only if it's relative coordinates would not be too close to the
-        // upper left tile border
-        int threshold = settings_.pseudotile_center_threshold;
-        int dx = tile.width/2 - inner.width/2;
-        int dy = tile.height/2 - inner.height/2;
-        inner.x = tile.x + ((dx < threshold) ? 0 : dx);
-        inner.y = tile.y + ((dy < threshold) ? 0 : dy);
-    }
+
+    // center the window in the outline tile
+    // but only if it's relative coordinates would not be too close to the
+    // upper left tile border
+    int threshold = settings_.pseudotile_center_threshold;
+    int dx = tile.width/2 - inner.width/2;
+    int dy = tile.height/2 - inner.height/2;
+    inner.x = tile.x + ((dx < threshold) ? 0 : dx);
+    inner.y = tile.y + ((dy < threshold) ? 0 : dy);
 
     //if (RECTANGLE_EQUALS(client->last_size, rect)
     //    && client->last_border_width == border_width) {
     //    return;
     //}
 
-    if (false) { // formely: if (tight_decoration)
+    if (scheme.tight_decoration()) {
+        // updating the outline only has an affect for tiled clients
+        // because for floating clients, this has been done already
+        // right when the window size changed.
         outline = scheme.inner_rect_to_outline(inner);
     }
     last_inner_rect = inner;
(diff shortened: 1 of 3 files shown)


t-wissmann:
A demonstration of the difference between `tight_decoration=on` and `tight_decoration=off`:
tight-decoration

blarz:
> Effectively, this commit only removes a redundant "if (!false)" and
> replaced one "if (false)" by "if (scheme.tight_decoration())". The rest
> is documentation and tests.

Why was it "falsed out" in the first place?

Also: Do we need to add it to the default autostart? (Are most of the theme related attributes listed there?)

src/decoration.cpp
```diff
@@ -149,23 +149,25 @@ void Decoration::resize_outline(Rectangle outline, const DecorationScheme& schem
...
+    // center the window in the outline tile
+    // but only if it's relative coordinates would not be too close to the
+    // upper left tile border
+    int threshold = settings_.pseudotile_center_threshold;
+    int dx = tile.width/2 - inner.width/2;
+    int dy = tile.height/2 - inner.height/2;
+    inner.x = tile.x + ((dx < threshold) ? 0 : dx);
+    inner.y = tile.y + ((dy < threshold) ? 0 : dy);
 
     //if (RECTANGLE_EQUALS(client->last_size, rect)
```
blarz, line 162: What is this about?

src/decoration.cpp
```diff
@@ -149,23 +149,25 @@ void Decoration::resize_outline(Rectangle outline, const DecorationScheme& schem
 
     auto tile = inner;
     client_->applysizehints(&inner.width, &inner.height);
-    if (!false) { // formely: if (!tight_decoration)
```
blarz, line 152: Was the if(!tight_decoration) removed on purpose?

t-wissmann:
> > Effectively, this commit only removes a redundant "if (!false)" and
> > replaced one "if (false)" by "if (scheme.tight_decoration())". The rest
> > is documentation and tests.
> 
> Why was it "falsed out" in the first place?

That happend during the migration to the new object system in 12d9e839f45cdb94d273a4ce33fb0fd9dedf55b1 where I replaced `tight_decoration` by `false` as a compromise to make progress. But then, I forgot to replace the `false` back to `tight_decoration`. That is what this PR does. But now I realize, that it makes sense to keep the first block (in `if (!false)`) activated, even if `tight_decoration` is enabled. (That is, I don't replace `false` by `tight_decoration` here).

> Also: Do we need to add it to the default autostart? (Are most of the theme related attributes listed there?)

I don't think it's necessary, because this is just a minor setting which most people (maybe even all?) shouldn't care about. I'm already unhappy, that the documentation is so long, because I fear that it draws too much attention.

src/decoration.cpp
t-wissmann: The `// if (RECTANGLE_EQUALS…`?  I originally intended to save unnecessary redraws if the geometry does not change. (But with the "full" window decorations we have, it's not enough to check for the rectangle and the border width...).

src/decoration.cpp
t-wissmann: (For the record: long explanation in https://github.com/herbstluftwm/herbstluftwm/pull/858#issuecomment-618831793)

blarz: Thanks!

src/decoration.cpp
blarz: Ok. I just question commented out code when I see some flying by during reviews ;)

src/decoration.cpp
blarz: Thanks!

blarz:
Thanks for the explanation, I think this is good to go then!

PR #860, opened 2020-04-23
Trick ccache into caching clang-tidy calls

dnnr:


.travis.yml
```diff
@@ -59,9 +59,10 @@ jobs:
               - docker run -u $UID --privileged -t --rm
                 --volume=/etc/passwd:/etc/passwd:ro
                 --volume=$PWD:/hlwm:rw
+                --volume=$HOME/.ccache-eoan-clang:/.ccache:rw
                 --volume=$HOME/.tox-cache:/hlwm/.tox:rw
                 eoan sh -c '
-                    /hlwm/ci/build.py --build-dir=/hlwm/build-clang-9 --cxx=clang++-9 --cc=clang-9 --build-type=Debug --check-using-std --iwyu --clang-tidy --flake8
```
blarz, line 64: Now that --clang-tidy is not used here anymore, shall we remove it?
Or keep it for now (for local use maybe?)

ci/clang-and-tidy.sh
```diff
@@ -0,0 +1,58 @@
...
+#   * the source file is always the last argument
+#   * the source file name matches *.cpp
+#   * the compile_commands.json in the build dir corresponds the current build
+#   * clang-tidy gets its config string passed via -config=
+#   * if a different clang or clang-tidy is used, this script is also modified
+#     (because ccache wouldn't detect any change of mtimes)
+###
+
+set -o errexit
+# set -o nounset
```
blarz, line 25: Why is this commented out?

ci/clang-and-tidy.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Perform the regular compiler invocation:
+progname=$(basename "$0")
+if [[ $progname == clang-* ]]; then
+    clang-9 "$@"
+elif [[ $progname == clang++-* ]]; then
+    clang++-9 "$@"
```
blarz, line 32: Seeing that we will maybe switch to clang-10 soon, maybe extract CLANG_VERSION variable?

blarz:
Did you verify it by introducing a clang-tidy triggering error?

ci/clang-and-tidy.sh
```diff
@@ -0,0 +1,58 @@
...
+# The purpose of this is to reduce CI check times by tricking ccache into
+# serving as a cache for clang-tidy as well. The base assumption for this is
+# inputs for clang-tidy match those of the compiler, so a ccache hit implies
+# that this file has passed the clang-tidy checks before.
+#
+# For simplicity, this script makes a couple of assumptions, that hold within
+# the our controlled CI environment:
+#   * there is only ever one source file per compiler invocation
+#   * the source file is always the last argument
+#   * the source file name matches *.cpp
```
blarz, line 17: or .c?

.travis.yml
```diff
@@ -12,11 +12,11 @@ branches:
 
 jobs:
     include:
-        - name: "Build with most recent GCC & Clang, run tests"
```
blarz, line 15: it still the most recent gcc AFAIK

.travis.yml
dnnr: Yeah, I'd rather keep it, might be useful.

ci/clang-and-tidy.sh
dnnr: Because I'm a dumbass :-) Fixed.

ci/clang-and-tidy.sh
dnnr: Well, this is tricky. See the comment at the top: In a normal setup, ccache knows to separate cache entries made with different compilers based on the mtime of the compiler binary. With clang-and-tidy.sh, we're using a content hash instead of the mtime. That still means that we need the content to change when the compiler changes. Putting the program name with its version in the file is an approximation of that. Ideally, we'd put the compiler binary's hash in the script and verify it every time.

Btw, this reminds me that I'll still need to handle the .clang-tidy file. Right now, changing the config doesn't make it check everything again (which it shoud).

.travis.yml
dnnr: Yes, but then I'd have to say the same about clang, and mentioning it didn't feel as necessary to me anymore. Contrasted with the ancient GCC, it's quite obvious that we're using recent toolchains.

ci/clang-and-tidy.sh
dnnr: Absolutely, thanks

dnnr:
> Did you verify it by introducing a clang-tidy triggering error?

At some point, yes. Maybe this is a good opportunity to start a draft PR that does it and keep it permanently open. (Only helps *after* merging a change like this, of course.)

.travis.yml
blarz: Ok!

ci/clang-and-tidy.sh
blarz: Thanks!

ci/clang-and-tidy.sh
blarz: Thanks!

.travis.yml
dnnr: Now I had to remove it after all, because I didn't want to have the .clang-tidy lie around...

.travis.yml
blarz: Makes sense

.travis.yml
```diff
@@ -59,9 +59,10 @@ jobs:
               - docker run -u $UID --privileged -t --rm
                 --volume=/etc/passwd:/etc/passwd:ro
                 --volume=$PWD:/hlwm:rw
+                --volume=$HOME/.ccache-eoan-clang:/.ccache:rw
                 --volume=$HOME/.tox-cache:/hlwm/.tox:rw
                 focal sh -c '
-                    /hlwm/ci/build.py --build-dir=/hlwm/build-clang-10 --cxx=clang++-10 --cc=clang-10 --build-type=Debug --check-using-std --iwyu --clang-tidy --flake8
+                    /hlwm/ci/build.py --build-dir=/hlwm/build-clang-9 --cxx=/hlwm/ci/clang++-and-tidy.sh --cc=/hlwm/ci/clang-and-tidy.sh --build-type=Debug --ccache=/.ccache --compile --check-using-std --iwyu --flake8
```
blarz, line 65: Merge error? (build-clang-9)

dnnr: Yup! It doesn't really matter. I took this opportunity to change both build dirs to `/hlwm/build`, which is totally sufficient after this change.

ci/clang-and-tidy.sh
dnnr: Alright, this is fixed now, so the PR should be ready I think.

.travis.yml
blarz: Thanks!

blarz:
I like this very much and find the solution really really neat, not sure if anybody else has thought of combining ccache and clang-tidy that way.
I'd like to see if it works out well in day to day use.

Before approving I'd like to check in with @t-wissmann if he's ok with this.

t-wissmann:
> I like this very much and find the solution really really neat, not sure if anybody else has thought of combining ccache and clang-tidy that way.
> I'd like to see if it works out well in day to day use.
> 
> Before approving I'd like to check in with @t-wissmann if he's ok with this.

I'm fine with it and I'm looking forward to the  faster CI :-)

PR #863, opened 2020-04-25
Move frame window collection to x11 fixture

t-wissmann:
I thought I needed the list of frames in another testcase (but
fortunately it turned out that I don't). However, I think it's usefull
to have the get_hlwm_frames() in a separate method for further tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -707,9 +707,9 @@ def x11(x11_connection):
             frame decoration windows"""
             cmd = ['xdotool', 'search', '--class', '_HERBST_FRAME']
             frame_wins = subprocess.run(cmd,
-                                  stdout=subprocess.PIPE,
-                                  universal_newlines=True,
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -702,6 +702,19 @@ def get_absolute_top_left(self, window):
                 window = tree.parent
             return (x, y)
 
+        def get_hlwm_frames(self):
+            """get list of window handles of herbstluftwm
+            frame decoration windows"""
+            cmd = ['xdotool', 'search', '--class', '_HERBST_FRAME']
+            frame_wins = subprocess.run(cmd,
+                                  stdout=subprocess.PIPE,
```
blarz, line 710: Needs to be the same indentation level as the opening parentheses.

blarz:
Seems like the build failed because of some flaky client in a test?

Issue #20, opened 2014-10-09
Click-focus for empty frames


The-Compiler:
We should be able to focus empty frames with click, or with the mouse if focus_follows_mouse is on

jfchevrette:
+1
I would also like to be able to click empty frames to select them even though focus_follows_mouse is off

jmcblane:
Is there any hope for this?

t-wissmann:
Yes, definitely. And if someone finds time to fix this I'm happy to merge it.

PR #864, opened 2020-04-25
Focus frames with the mouse

t-wissmann:
This adds basic support for selecting frames with the mouse: either by
click or by mouse hover (if focus_follows_mouse is enabled). This change
only concerns the actual frame. If there is no frame shown (because
always_show_frame=off) or if there is a hole in the frame (because
frame_bg_transparent=on), then we only obtain events for the root window
that are currently ignored. (Interpreting event coordinates on the root
window is in the scope of another change)

On the implementation level, this introduces a static map that
transforms window ids into frame decorations (I think a 'framemanager'
is overkill here). The focusFrame() function is added to the root
window, because it affects the global state of hlwm (and I think
focus_client() should go to Root at some point, too).

This solves issue #20 (in principle) and adds a test case for it.

### DIFF
--- a/src/xmainloop.cpp
+++ b/src/xmainloop.cpp
@@ -10,6 +10,7 @@
 #include "clientmanager.h"
 #include "desktopwindow.h"
 #include "ewmh.h"
+#include "framedecoration.h"
 #include "frametree.h"
 #include "hlwmcommon.h"
 #include "ipc-server.h"
@@ -179,13 +180,20 @@ void XMainLoop::quit() {
 
 void XMainLoop::buttonpress(XButtonEvent* be) {
     MouseManager* mm = root_->mouse();
-    HSDebug("name is: ButtonPress on sub %lx, win %lx\n", be->subwindow, be->window);
+    HSDebug("name is: ButtonPress on sub 0x%lx, win 0x%lx\n", be->subwindow, be->window);
     if (!mm->mouse_handle_event(be->state, be->button, be->window)) {
         // if the event was not handled by the mouse manager, pass it to the client:
         Client* client = root_->clients->client(be->window);
         if (client) {
             bool raise = root_->settings->raise_on_click();
             focus_client(client, false, true, raise);
+        }
+    }
+    FrameDecoration* frameDec = FrameDecoration::withWindow(be->window);
+    if (frameDec) {
+        auto frame = frameDec->frame();
+        if (frame)  {
+            root_->focusFrame(frame);
         }
     }
     XAllowEvents(X_.display(), ReplayPointer, be->time);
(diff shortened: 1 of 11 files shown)


src/xmainloop.cpp
```diff
@@ -334,6 +342,18 @@ void XMainLoop::enternotify(XCrossingEvent* ce) {
...
+        if (!c) {
+            // if it's not a client window, it's maybe a frame
+            FrameDecoration* frameDec = FrameDecoration::withWindow(ce->window);
+            if (frameDec) {
+                auto frame = frameDec->frame();
+                HSWeakAssert(frame);
+                if (frame) {
+                    root_->focusFrame(frame);
+                }
+            } else {
```
blarz, line 354: Is the empty else condition really necessary?

tests/test_basic_events.py
```diff
@@ -50,3 +50,32 @@ def test_focus_follows_mouse(hlwm, mouse, focus_follows_mouse, single_floating):
...
+
+    # go to left frame
+    mouse.move_to(width * 0.25, height * 0.5)
+
+    # go to right frame
+    mouse.move_to(width * 0.75, height * 0.5)
+    if click:
+        mouse.click('1')
+
+    # we expect the focus to got to the second frame
```
blarz, line 77: s/got/go/

src/xmainloop.cpp
t-wissmann: I've removed it (I had a printf debugging statement there before).

Piping:
Hi,

Just found this awesome windows manager. Feels quite comfortable after an hour and trail and experiments.

Only thing I left un-configured is root windows access in X11. ( the mouse click behavior binding for root window)

Currently if I enabled this new build, I can only select the frame by clicking the frame border. ( if I enable hovering, and move mouse really fast, then the focus sometimes not working as well).

I assume it has to do with root windows access in the code right?  Is it possible to enable mouse binding to root windows now? or could transparent frame can catch mouse click like other windows that have the content?

t-wissmann:
> Currently if I enabled this new build, I can only select the frame by clicking the frame border. ( if I enable hovering, and move mouse really fast, then the focus sometimes does not work).

Yes. That's probably because the frame only consists of the border of the frame.
As of now, hlwm does not handle clicks on the root window.

> I assume it has to do with root windows access in the code right? Is it possible to enable mouse binding to root windows now? or could transparent frame can catch mouse click like other windows that have the content?

I don't get your last sentence, but the reason for the lacking features is that nobody got time yet to implement it. The "frame_bg_transparent" setting is implemented by cutting an actual hole in the frame, so the frame window does not receive any click events if someone clicks into the hole, but the window behind receives the event instead.

Piping:
> the reason for the lacking features is that nobody got time yet to implement it

@t-wissmann Cool. I would like to contribute code for the root windows access handles if it is not a philosophical design choice.

t-wissmann:
> > the reason for the lacking features is that nobody got time yet to implement it
> 
> @t-wissmann Cool. I would like to contribute code for the root windows access handles if it is not a philosophical design choice.

OK great!

PR #868, opened 2020-04-26
Handle mouse events on decoration window

t-wissmann:
Handle mouse events in the intuitive way:

  - Click on the decoration window focuses the client
  - If focus_follows_mouse is enabled, hovering is enough
  - Dragging the edge of the window decoration resizes the client
  - Dragging the more inner part of the window decoration moves the
    client.


t-wissmann:
I've already wasted too much time on the testcase, but I don't get focus_follows_mouse to work in Xvfb (the CI). At least, it should cover most of the new code in this PR.

blarz:
> Dragging the more inner part of the window decoration moves the
client.

This doesn't work for me, even with `herbstclient attr theme.border_width 50`. I always end up resizing, never moving. Where is the "more inner" part?

tiosgz:
> This doesn't work for me, even with `herbstclient attr theme.border_width 50`. I always end up resizing, never moving. Where is the "more inner" part?

Don't you need to also increase `theme.inner_width`?

t-wissmann:
> > Dragging the more inner part of the window decoration moves the
> > client.
> 
> This doesn't work for me, even with `herbstclient attr theme.border_width 50`. I always end up resizing, never moving. Where is the "more inner" part?

The `border_width` outer pixels of the decoration are used for resizing. This means if you set border_width to 50, everything is 'resize'-area. you need to increase `padding_top`. (Setting `inner_width` does not help because this does not increase the width of the decoration)

t-wissmann:
I agree that this might not make sense at the moment, because most people won't have `padding_top` activated, but I think it will make sense as soon as we have proper window titles.

blarz:
With `padding_top` I got it to move, thanks!

blarz:
The test currently only tests the focus change, can we add one that tests that resizing works maybe?

t-wissmann:
> The test currently only tests the focus change, can we add one that tests that resizing works maybe?

Originally, I didn't want to because the non-working focus_follows_mouse (which still does not work in CI) frutstrated me too much. Some hours later I was able to write tests for moving and resizing the clients via the decoration.

blarz:
> > The test currently only tests the focus change, can we add one that tests that resizing works maybe?
> 
> Originally, I didn't want to because the non-working focus_follows_mouse (which still does not work in CI) frutstrated me too much. Some hours later I was able to write tests for moving and resizing the clients via the decoration.

Perfect, thank you!

PR #872, opened 2020-04-26
Add test for floating resize

blarz:


tests/test_floating.py
```diff
@@ -107,3 +107,20 @@ def test_bring_floating_from_different_tag(hlwm, x11):
...
+    assert (geom_before.width, geom_before.height) == (300, 200)
+
+    hlwm.call(['resize', direction, '+0.5'])
+
+    height_offset = 300 if direction == 'down' else 0
+    width_offset = 450 if direction == 'right' else 0
+    expected_geom = (geom_before.width + width_offset, geom_before.height + height_offset)
+
+    geom_after = client.get_geometry()
+    assert (geom_after.width, geom_after.height) == expected_geom
```
blarz, line 126: This gets me
> assert (745, 200) == (750, 200)

Close, but still. Is this intended?

blarz: Ok I messed something up, let's see if this helps

t-wissmann: `resize` lets clients grow to the monitor edge. I've modified the testcase in your PR to check this.

blarz: Thanks! I forgot about that, I even approved the PR back then, that actually implemented that. The documentation does not really reflect that iMHO, I think I might create a follow up PR for this.

blarz:
Thanks for updating the PR @t-wissmann, the changes look good to me! (I cannot approve my own PR ;))

tests/test_floating.py
blarz: Created #878

PR #874, opened 2020-04-26
Test EWMH message _NET_ACTIVE_WINDOW

t-wissmann:


tests/test_ewmh.py
```diff
@@ -124,6 +124,38 @@ def test_wm_state_type(hlwm, x11):
     assert len(prop.value) == 2
 
 
+def test_ewmh_focus_client(hlwm, x11):
+    hlwm.call('set focus_stealing_prevention off')
+    # add another client that has the focus on the other tag
```
blarz, line 129: There is no other tag?

tests/test_ewmh.py
```diff
@@ -124,6 +124,38 @@ def test_wm_state_type(hlwm, x11):
...
+    hlwm.call('add tag2')
+    if on_another_monitor:  # if the tag shall be on another monitor
+        hlwm.call('add_monitor 800x600+600+0')
+    hlwm.call('rule tag=tag2 focus=off')
+    # add another client that has the focus on the other tag
+    x11.create_client()
+    handle, winid = x11.create_client()
+    assert 'focus' not in hlwm.list_children('clients')
+
+    x11.ewmh.setActiveWindow(handle)
```
blarz, line 152: The name "handle" makes it really hard to understand, which client is meant (same in the test case above).
I cannot think of a better name, but client_one and client_two would be an improvement.

dnnr: This comment just made me notice that we have an `x11.create_client()` that returns the window ID as a second value and a `hlwm.create_client()` that returns it as the first window ID. Maybe we should make that consistent.

t-wissmann: > This comment just made me notice that we have an `x11.create_client()` that returns the window ID as a second value and a `hlwm.create_client()` that returns it as the first window ID. Maybe we should make that consistent.

This confuses me as well a lot.

tests/test_ewmh.py
t-wissmann: I've fixed it, thanks! (And I've also renamed the variables)

blarz: Thanks! Can you maybe update the variables in the second testcase, too?

t-wissmann: I've renamed it in PR #882.

blarz: Alright, thanks!

PR #877, opened 2020-04-26
Try to fix flakyness in mouse tests

blarz:


blarz:
Hrm, that does not seem to fix it.

t-wissmann:
I think the "proper" way is what we are doing in `click` where we wait for the `ButtonPress` (resp. `ButtonRelease`) to be printed

t-wissmann:
> I think the "proper" way is what we are doing in `click` where we wait for the `ButtonPress` (resp. `ButtonRelease`) to be printed

I've added a attempt as #881.

blarz:
#881 should take care of that in a more elegant way

Issue #99, opened 2017-07-19
Balance frames? [feature request]


vyp:
Is there a way to reset the frame split fractions? (as in say to 0.5 for default, or to a given fraction)

t-wissmann:
No, there is none. It may change in future versions, but something like this is of quite a low priority.

You can simply adjust the split fraction using the existing (relative) keybinds.

t-wissmann:
In the current git version, you can write `0.5` to the `fraction` attribute of frame objects. For example, you could do the following:
```bash
for idx in .{,{0.,1.}{,{0.,1.}{,{0.,1.}}}} ; do
   hc attr tags.focus.tiling.root${idx}fraction 0.5
done
```
(This is a brute-force solution, because it tries all frame indices up to depth 3 and tries to set the fraction to 0.5)

t-wissmann:
As soon as #1394 is merged, you can recursively set the fraction to 0.5 for all frames on a tag:
```
$ hc foreach F tags.focus.tiling.root --recursive sprintf ATTR '%c.fraction' F try silent set_attr ATTR 0.5
```
This also tries to set the fraction attribute for leaf frames, which of course have no fraction attribute. Hence, I supress the error messages via `try silent`.

If there is anything else, feel free to (re-)open this or another issue.

PR #878, opened 2020-04-26
Update and clarify the documentation of 'resize'

blarz:


doc/herbstluftwm.txt
```diff
@@ -478,12 +478,14 @@ bring 'WINID'::
...
-    should not omit the sign '-' or '+', because in future versions, the
-    behaviour may change if the sign is omitted. Example:
+    Changes the size of the focused frame (or floating window) in specified
+    'DIRECTION' by 'FRACTIONDELTA' (which defaults to 0.02 if none is
+    supplied). 'DIRECTION' behaves as specified at the 'focus' command. If a
+    floating window is focused, it grows towards the monitor edge of the
+    specified 'DIRECTION' ('FRACTIONDELTA' is ignored in that case). In the
+    parameter 'FRACTIONDELTA', you should not omit the sign '-' or '+',
+    because in future versions, the behaviour may change if the sign is
+    omitted. Example:
```
dnnr, line 488: The first sentence is much clearer now. :+1: 

The last sentence confused me a bit because I missed the "not" when first reading. How about fool-proofing it a bit like this? `An explicit '-' or '+' is recommended for the parameter 'FRACTIONDELTA', because in future versions, the behavior without a sign may change.`

But maybe first of all: Do we know what you had in mind here, @t-wissmann? Maybe we can drop that entire statement. It's been there for 9 years now.

doc/herbstluftwm.txt
```diff
@@ -478,12 +478,14 @@ bring 'WINID'::
 
 
 resize 'DIRECTION' ['FRACTIONDELTA']::
-    Changes the next fraction in specified 'DIRECTION' by 'FRACTIONDELTA' (which
-    defaults to 0.02 if none is supplied). 'DIRECTION' behaves as specified at
-    the 'focus' command. If a floating window is focused, it is resized and
-    grows into the specified 'DIRECTION'. In the parameter 'FRACTIONDELTA', you
-    should not omit the sign '-' or '+', because in future versions, the
-    behaviour may change if the sign is omitted. Example:
+    Changes the size of the focused frame (or floating window) in specified
```
tiosgz, line 481: "(or floating window)" is a bit confusing here, because you first tell it's resized by `FRACTIONDELTA` and *then* change the behaviour expected for floating clients.

doc/herbstluftwm.txt
t-wissmann: Yes we can drop it, because I'd never change the behaviour now. I think I had something in mind like setting the fraction of a frame split to a fixed value (e.g. resetting it to 0.5). I think this is what is requested in #99.

More importantly: `resize` command for floating windows does not only grow to the monitor edge, but actually to the next visible edge on the monitor, that is, to the edge of another window or to the monitor edge if there is no other window in between.

blarz: I dropped the -+ sign text and updated the edge text

doc/herbstluftwm.txt
blarz: Thanks, rephrased it a bit

PR #881, opened 2020-04-27
Synchronize mouse press/release with hlwm output

t-wissmann:
In the mouse fixture, wait for herbstluftwm to process the ButtonPress
and ButtonRelease events as we were doing it already in the Mouse.click()
method. This should fix the current flakyness of
test_drag_resize_floating_client

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -860,11 +860,21 @@ def mouse(hlwm_process):
         def move_relative(self, delta_x, delta_y):
             self.call_cmd(f'xdotool mousemove_relative --sync {delta_x} {delta_y}', shell=True)
 
-        def mouse_press(self, button):
-            self.call_cmd(f'xdotool mousedown {button}', shell=True)
+        def mouse_press(self, button, wait=True):
+            cmd = ['xdotool', 'mousedown', button]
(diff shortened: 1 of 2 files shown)


t-wissmann:
An alternative to PR #877.

blarz:
Hm still seems flaky

t-wissmann:
> Hm still seems flaky

Hm maybe the python-xlib caches the geometry or didn't get the resize event ("configure notify") yet. Maybe a `x11.display.sync()` helps here. 

I've also added another parameter to run the test case multiple times (we did something similar in `test_drag_move`.)

t-wissmann:
> > Hm still seems flaky
> 
> Hm maybe the python-xlib caches the geometry or didn't get the resize event ("configure notify") yet. Maybe a `x11.display.sync()` helps here.
> 
> I've also added another parameter to run the test case multiple times (we did something similar in `test_drag_move`.)

It still fails. I added another assertion, maybe this helps us to understand what's going on.

t-wissmann:
Maybe we should still merge it, because I think the synchronization with the hlwm event queue should make things better anyway, even though it does not all race-conditions.

PR #888, opened 2020-04-28
Test 'focus_edge' command

t-wissmann:
This requires the return code fix of PR #885 to succeed.

### DIFF
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -697,8 +697,6 @@ def test_focus_edge(hlwm):
     assert hlwm.get_attr('monitors.focus.index') == '0'
     # but right-most frame means, if we go right once more, we're on
     # the other monitor:
-    hlwm.call('list_monitors')
-    hlwm.call('dump')
     hlwm.call('focus right')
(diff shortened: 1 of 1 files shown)


tests/test_layout.py
```diff
@@ -676,3 +676,29 @@ def test_focus_other_monitor(hlwm, other_mon_exists, setting):
...
+    assert layout_before == hlwm.call('dump').stdout
+    assert hlwm.get_attr('monitors.focus.index') == '0'
+
+    # focus_edge goes to the rightmost frame
+    hlwm.call('focus_edge right')
+    # we're still on the first monitor
+    assert hlwm.get_attr('monitors.focus.index') == '0'
+    # but right-most frame means, if we go right once more, we're on
+    # the other monitor:
+    hlwm.call('list_monitors')
```
blarz, line 700: Is this call needed? Same goes for the call one line below

t-wissmann: Ah, that was part of my printf-debugging and is not needed, thanks! I've removed these two lines.

PR #891, opened 2020-05-02
Add test for switchtag

### LINKED ISSUE #844: Add tests for the rule switchtag=
We don't have tests yet that the `switchtag` rule works.

blarz:
Fixes #844


blarz:
Let me know if I should expand the tests.

I thought it is a bit easier to read if I have two tests instead of one big test with three pytest parameters. But I am open to suggestions.

PR #897, opened 2020-05-03
Test fullscreen/pseudotile on invalid argument

t-wissmann:


blarz:
Failed with:
tests/test_herbstluftwm.py::test_reload 

[gw1] [ 11%] FAILED tests/test_herbstluftwm.py::test_reload 

=================================== FAILURES ===================================

_________________________________ test_reload __________________________________

[gw1] linux -- Python 3.8.2 /hlwm/.tox/py38/bin/python

hlwm_process = <conftest.HlwmProcess object at 0x7fcce903f3d0>

hlwm = <conftest.HlwmBridge object at 0x7fcce9022eb0>

    def test_reload(hlwm_process, hlwm):

        with hlwm_process.wait_stdout_match('hlwm started'):

>           hlwm.call('reload')

tests/test_herbstluftwm.py:15: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

/usr/lib/python3.8/contextlib.py:120: in __exit__

    next(self.gen)

tests/conftest.py:396: in wait_stdout_match

    self.read_and_echo_output(until_stdout=match)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <conftest.HlwmProcess object at 0x7fcce903f3d0>

until_stdout = 'hlwm started', until_stderr = None, until_eof = False

    def read_and_echo_output(self, until_stdout=None, until_stderr=None, until_eof=False):

        expect_sth = ((until_stdout or until_stderr) is not None)

        max_wait = 5

    

        # Track which file objects have EOFed:

        eof_fileobjs = set()

        fileobjs = set(k.fileobj for k in self.output_selector.get_map().values())

    

        stderr = ''

        stderr_bytes = bytes()

        stdout = ''

        stdout_bytes = bytes()

    

        def match_found():

            if until_stdout and (until_stdout in stdout):

                return True

            if until_stderr and (until_stderr in stderr):

                return True

            return False

    

        started = datetime.now()

        while (datetime.now() - started).total_seconds() < max_wait:

            select_timeout = 1

            # If we're not polling for a matching string (anymore), there is no

            # need for a dampening timeout:

            if not expect_sth or match_found():

                select_timeout = 0

            selected = self.output_selector.select(timeout=select_timeout)

            for key, events in selected:

                # Read only single byte, otherwise we might block:

                ch = key.fileobj.read(1)

    

                if ch == b'':

                    eof_fileobjs.add(key.fileobj)

    

                # Store in temporary buffer for string matching:

                if key.fileobj == self.proc.stderr:

                    stderr_bytes += ch

                    if ch == b'\n':

                        stderr += stderr_bytes.decode()

                        # Pass it through to the real stderr:

                        key.data.write(stderr_bytes.decode())

                        key.data.flush()

                        stderr_bytes = b''

    

                if key.fileobj == self.proc.stdout:

                    stdout_bytes += ch

                    if ch == b'\n':

                        stdout += stdout_bytes.decode()

                        # Pass it through to the real stdout:

                        key.data.write(stdout_bytes.decode())

                        key.data.flush()

                        stdout_bytes = b''

    

            if until_eof:

                # We are going to the very end, so carry on until all file

                # objects have returned EOF:

                if eof_fileobjs == fileobjs:

                    break

                else:

                    continue

    

            if selected != []:

                # There is still data available, so keep reading (no matter

                # what):

                continue

    

            # But stop reading if there is nothing to look for or we have

            # already found it:

            if not expect_sth or match_found():

                break

    

        duration = (datetime.now() - started).total_seconds()

        if expect_sth and not match_found():

>           assert False, f'Expected string not encountered within {duration:.1f} seconds'

E           AssertionError: Expected string not encountered within 5.0 seconds

E           assert False

tests/conftest.py:386: AssertionError

---------------------------- Captured stdout setup -----------------------------

hlwm started

---------------------------- Captured stderr setup -----------------------------

EWMH: 1 desktops: 'default'

0 managed clients: 

----------------------------- Captured stdout call -----------------------------

Client command ['reload'] succeeded (no output)

hlwm started

=========================== short test summary info ============================

FAILED tests/test_herbstluftwm.py::test_reload - AssertionError: Expected str...

!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!

!!!!!!!!!!!! xdist.dsession.Interrupted: stopping after 1 failures !!!!!!!!!!!!!

================== 1 failed, 128 passed in 199.23s (0:03:19) ===================

ERROR: InvocationError for command /hlwm/.tox/py38/bin/python -m pytest -n auto --cache-clear -v -x -m 'not exclude_from_coverage' (exited with code 2)

blarz:
Ugh, sorry, the mark up got seriously messed up

t-wissmann:
> Ugh, sorry, the mark up got seriously messed up

I've fixed it. (the raw travis log has double newlines for some reasons that I don't understand)

Regarding the failure: I have no idea what what's the issue. Maybe it's simply enough to increase the timeout. (Increasing the timeout is safe IMO: it does not increase the run time in the successful case)

blarz:
> > Ugh, sorry, the mark up got seriously messed up
> 
> I've fixed it. (the raw travis log has double newlines for some reasons that I don't understand)

Thanks!

> Regarding the failure: I have no idea what what's the issue. Maybe it's simply enough to increase the timeout. (Increasing the timeout is safe IMO: it does not increase the run time in the successful case)

Was the change maybe not based on top one of your previous fixes?

t-wissmann:
> > > Ugh, sorry, the mark up got seriously messed up
> > 
> > 
> > I've fixed it. (the raw travis log has double newlines for some reasons that I don't understand)
> 
> Thanks!
> 
> > Regarding the failure: I have no idea what what's the issue. Maybe it's simply enough to increase the timeout. (Increasing the timeout is safe IMO: it does not increase the run time in the successful case)
> 
> Was the change maybe not based on top one of your previous fixes?

I think it was recent enought (master branch of today). I guess we should just increase the timeout (#900) and carry on :-)

blarz:
> I think it was recent enought (master branch of today). I guess we should just increase the timeout (#900) and carry on :-)

True :)

Issue #590, opened 2019-11-15
Crash for unknown reason (herbstclient? mouse selection?)


t-wissmann:
I just quickly switched between my mail client (in urxvt) and the q3terminal.sh in order to copy some text out of the mail client to the urxvt in the q3terminal when herbstluftwm suddenly segfaultet. This here is the backtrace:
```
#0  0x00007ffff7877823 in __memmove_avx_unaligned_erms () from /usr/lib/libc.so.6
#1  0x00007ffff7b7dc71 in std::char_traits<char>::copy (__n=81, __s2=0x1ad122 <error: Cannot access memory at address 0x1ad122>, __s1=<optimized out>) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/char_traits.h:365
#2  std::basic_streambuf<char, std::char_traits<char> >::xsputn (this=0x7fffffffdfd8, __s=0x1ad122 <error: Cannot access memory at address 0x1ad122>, __n=81) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/streambuf.tcc:90
#3  0x00007ffff7b6c22b in std::basic_streambuf<char, std::char_traits<char> >::sputn (__n=81, __s=<optimized out>, this=<optimized out>) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/streambuf:457
#4  std::__ostream_write<char, std::char_traits<char> > (__n=81, __s=<optimized out>, __out=...) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/ostream_insert.h:50
#5  std::__ostream_insert<char, std::char_traits<char> > (__out=..., __s=0x1ad122 <error: Cannot access memory at address 0x1ad122>, __n=81) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/ostream_insert.h:101
#6  0x000055555565c526 in std::operator<< <char, std::char_traits<char>, std::allocator<char> > (__str=..., __os=...) at /usr/include/c++/8.3.0/bits/basic_string.h:6323
#7  MonitorManager::<lambda(Input, Output)>::operator() (output=..., input=..., __closure=0x555555814290) at /home/thorsten/git/winterbreeze-live-test/src/monitormanager.cpp:179
[...]

PR #898, opened 2020-05-03
Test feeding junk to commands

t-wissmann:
This adds a test case that tries to call commands with a junk argument.
Of course this should not affect the coverage, but I hope that with
this, we will find bugs in parser exception handling earlier (such as
pr #590).

### DIFF
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -245,7 +245,7 @@ def test_junk_args_dont_crash(hlwm, args_before, junk_arg):
             continue
         full_cmd = [cmd_name]
         for _ in range(0, args_before):
-            # find some arg apropriate for the current command
+            # find some arg appropriate for the current command
             completions = hlwm.unchecked_call(
                 ['complete', str(len(full_cmd))] + full_cmd).stdout.splitlines()
             completions.append('')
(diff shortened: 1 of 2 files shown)


t-wissmann:
This should fail until #899 is merged.

tests/test_completion.py
```diff
@@ -231,3 +231,24 @@ def test_posix_escape(hlwm):
...
+@pytest.mark.parametrize("args_before", [0, 1, 2])
+@pytest.mark.parametrize("junk_arg", ['junk', '234', ' '])
+def test_junk_args_dont_crash(hlwm, args_before, junk_arg):
+    commands = hlwm.call('list_commands').stdout.splitlines()
+    for cmd_name in commands:
+        if cmd_name in ['wmexec', 'quit']:
+            continue
+        full_cmd = [cmd_name]
+        for _ in range(0, args_before):
+            # find some arg apropriate for the current command
```
blarz, line 248: *appropriate

t-wissmann: Thanks! It's fixed.

blarz: Thanks!

PR #900, opened 2020-05-03
Increase timeout when testing stderr/stdout

t-wissmann:
In the tests, increase the timeout when waiting for certain lines on
stderr/stdout. A higher timeout should not affect the run time in the
successful case but simply be less picky regarding the scheduling.


blarz:
Still seems to fail :(
https://travis-ci.com/github/herbstluftwm/herbstluftwm/jobs/327099571

t-wissmann:
:-(
I've created an issue for it (#902)

PR #907, opened 2020-05-11
Ignore EnterNotify events by grabs and decoration

t-wissmann:
DWM ignores these events, too.

### DIFF
--- a/src/xmainloop.cpp
+++ b/src/xmainloop.cpp
@@ -338,7 +338,14 @@ void XMainLoop::destroynotify(XUnmapEvent* event) {
 }
 
 void XMainLoop::enternotify(XCrossingEvent* ce) {
-    HSDebug("name is: EnterNotify, focus = %d\n", ce->focus);
+    HSDebug("EnterNotify, focus = %d, window = 0x%lx\n", ce->focus, ce->window);
+    if (ce->mode != NotifyNormal || ce->detail == NotifyInferior) {
+        // ignore an event if it is caused by (un-)grabbing the mouse or
+        // if the pointer moves from a window to its decoration.
+        // for 'ce->detail' see:
+        // https://tronche.com/gui/x/xlib/events/window-entry-exit/normal.html
+        return;
+    }
     if (!root_->mouse->mouse_is_dragging()
         && root_->settings()->focus_follows_mouse()
         && ce->focus == false) {


blarz:
How can I verify this?

t-wissmann:
> How can I verify this?

I do not know how to reproduce this deterministically. I think that this avoids focus changes whenever a menu disappears and the cursor is now within another window.

I tried this with rofi, and it helped sometimes, but I'm not very sure. dwm is ignoring these enter notifies, so I think it's reasonable to do it.

However, I failed to write a test case for it, because if I grab/move/ungrab the cursor, I still get a focus change. For the record: I was using the following test case:
```
@pytest.mark.parametrize("grab", [False, True])
def test_no_focus_change_on_ungrab(hlwm, x11, grab):
    hlwm.call('set focus_follows_mouse on')
    handle1, winid1 = x11.create_client()
    handle2, winid2 = x11.create_client()
    mon_height = int(hlwm.call('monitor_rect').stdout.split(' ')[3])
    print('warping to y= {}'.format(mon_height // 4))
    x11.root.warp_pointer(50, mon_height // 4)
    x11.display.sync()
    hlwm.call('dump')
    print(x11.get_absolute_top_left(handle1))
    print(x11.get_absolute_top_left(handle2))
    assert hlwm.get_attr('clients.focus.winid') == winid1

    if grab:
        x11.root.grab_pointer(False, 0, X.GrabModeSync, X.GrabModeSync,
                              0, 0, X.CurrentTime)
    # move into the second window
    x11.root.warp_pointer(50, (3 * mon_height) // 4)
    x11.display.sync()

    if grab:
        assert hlwm.get_attr('clients.focus.winid') == winid1
        x11.display.ungrab_pointer(X.CurrentTime)
    x11.display.sync()
    hlwm.call('true')  # sync with hlwm

    # if we grabbed the cursor, no focus change is expected
    assert hlwm.get_attr('clients.focus.winid') == \
        (winid1 if grab else winid2)
```

blarz:
I played around with it, doesn't break anything for me. But doesn't fix anything for me either ;)
If it does for you, we can put it in for sure and I approve instantly.
I am just struggling right now, as I fail to see any results and maybe lack the tools to verify it.

t-wissmann:
I also don't have a high confidence, but I really think we should give it a try and see how it goes.

Issue #908, opened 2020-05-12
0.8.2: focus_follows_mouse has become unreliable


Nairou:
In 0.8.0 this worked perfectly, window focus would follow the mouse instantly. In the 2 days since I updated to 0.8.2, I've had numerous cases where moving the mouse over a window didn't move focus, and I had to move the mouse over another window and back again to trigger the focus change.

There was even one case where mouse focus failed entirely, and I had to actually click on windows to move focus until I restarted herbstluftwm.

Moving focus is normally something I do without thinking, so I tend to notice focus not moving when I try to type in the new window and nothing happens.

t-wissmann:
I have similar issues, but so far I failed reproducing it (It only happens from time to time here). Do you know a way of reproducing it?

I'm sure that this issue was introduced by one of the following pull requests: https://github.com/herbstluftwm/herbstluftwm/pull/868 or https://github.com/herbstluftwm/herbstluftwm/pull/864

t-wissmann:
I think I can reproduce it with two frames and with
```
hc set frame_border_width 20
```
so my first guess is that https://github.com/herbstluftwm/herbstluftwm/pull/864 is the bad guy.

Nairou:
Upping my frame_border_width from 3 to 20 didn't appear to make a difference for me.

It's still intermittent, though it seems more likely if a particular window has had focus for a while. When I try to test it, and move focus away from the browser and back again, I'll see it skip once on the way back (not giving focus back to the browser), but as I continue to move the mouse back and forth between windows quickly it doesn't skip any more.

Nairou:
I think I found a reliable way to reproduce it. It seems to happen if the mouse moves fast enough that it doesn't touch any part of the window frame or border.

If you flick the mouse over fast enough that you can see the trail of cursor movement where it goes directly from the contents of one window to the contents of the next window, without touching anything in between, then focus doesn't follow.

t-wissmann:
Does it make a difference whether the two windows are in the same frame or in different frames?

Nairou:
Just tested with two windows in the same frame, and it does the same thing there as well. Focus only switches if the border gets touched. All other tests were against windows in separate frames.

t-wissmann:
And can you reproduce it if you switch the tag to floating mode?

t-wissmann:
[...]

PR #909, opened 2020-05-14
Don't drop later EnterNotify during enternotify()

t-wissmann:
This fixes a race condition, reported in #908.

In general, Monitor::applyLayout() possibly drops future EnterNotify
events in the X11 event queue that might have been artifically triggered
because the new tiling on a monitor might have moved a window to the
cursor.

During the handling of an EnterNotify event, no such artificial enter
notify events can be created, because only the focus is changed, so
there is not necessity to drop those. Moreover, on quick mouse
movements, multiple enter notify events can be created if the cursor
traverses over multiple windows. In this case, we must not drop these
later EnterNotify events.

(These events can be triggered while hlwm is already processing the
first EnterNotify event, so it is not enough to skip to the newest
EnterNotify event in the enternotify() event handler).

To accomplish this restriction, the dropEnterNotifyEvents() function is
moved to XMainLoop where we keep track of whether we are in
enternotify(). On a positive side, we get rid of calling an X function
in monitor.cpp.

There are two new test cases:

  - The first test case verifies that after resizing, non-welcome
    EnterNotify events are dropped.
  - The second one checks that for a number of EnterNotify events, the
    last one survives. Unfortunately, I did not manage to establish a
    test case that fails before the present commit. (But I think this is
    better than nothing because it still tests something)

In order to test this manually, do the following:

  - herbstclient set focus_follows_mouse on
  - herbstclient set frame_border_width 20
  - put two clients in separate frames side by side
  - move the mouse cursor quickly between the two clients.
  - before the present commit, the focus isn't updated reliably, but
    with this commit, it is. (Instead of switching between git commits
    one can also simply deactivate the 'return' in
    XMainLoop::dropEnterNotifyEvents())

### DIFF
--- a/src/monitor.cpp
+++ b/src/monitor.cpp
@@ -49,7 +49,7 @@ Monitor::Monitor(Settings* settings_, MonitorManager* monman_, Rectangle rect_, 
     , mouse { 0, 0 }
     , rect(rect_)
     , settings(settings_)
-    //, monman(monman_)
+    , monman(monman_)
 {
     for (auto i : {&pad_up, &pad_left, &pad_right, &pad_down}) {
         i->setWriteable();
@@ -261,7 +261,7 @@ void Monitor::applyLayout() {
 
     // remove all enternotify-events from the event queue that were
     // generated while arranging the clients on this monitor
-    drop_enternotify_events();
+    monman->dropEnterNotifyEvents.emit();
 }
 
 Monitor* find_monitor_by_name(const char* name) {
@@ -472,8 +472,6 @@ int monitor_set_tag(Monitor* monitor, HSTag* tag) {
             monitor->restack();
             other->applyLayout();
             monitor->applyLayout();
-            // discard enternotify-events
-            drop_enternotify_events();
             monitor_update_focus_objects();
             Ewmh::get().updateCurrentDesktop();
             emit_tag_changed(other->tag, other->index());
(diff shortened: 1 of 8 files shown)


src/xmainloop.cpp
```diff
@@ -346,6 +364,8 @@ void XMainLoop::enternotify(XCrossingEvent* ce) {
         // https://tronche.com/gui/x/xlib/events/window-entry-exit/normal.html
         return;
     }
+    // Warning: we have to set this to false again!
+    duringEnterNotify_ = true;
```
t-wissmann, line 368: Here, I wasn't sure whats cleaner:
* a warning stating that we read if we add additional `return` instructions
* creating a kind of `OnScopeLeave` callback object that sets `duringEnterNotify_=false` in its destructor. This is more complicated, but works if we decide to add `return` statements somewhere in the following `if`-`else`-blocks.

src/monitor.cpp
```diff
@@ -49,7 +49,7 @@ Monitor::Monitor(Settings* settings_, MonitorManager* monman_, Rectangle rect_,
     , mouse { 0, 0 }
     , rect(rect_)
     , settings(settings_)
-    //, monman(monman_)
+    , monman(monman_)
```
t-wissmann, line 52: I know that the names are the wrong way round... actually the member variables should have the `_`-suffix, not the constructor arguments.

t-wissmann:
I'm using this already, and I think it improves the focus_follows_mouse thing a lot..

PR #910, opened 2020-05-17
Add minimal test for tag_status

blarz:


blarz:
Not sure if test_rootcommands.py is the best place to put it.

PR #911, opened 2020-05-17
Add test for 'drag zoom'

blarz:


tests/test_mousebind.py
```diff
@@ -206,6 +206,27 @@ def test_drag_resize_floating_client(hlwm, x11, mouse, live_update):
...
 
 
+def test_drag_zoom_floating_client(hlwm, x11, mouse):
+    client, winid = x11.create_client(geometry=(50, 50, 300, 200))
+    hlwm.call(f'set_attr clients.{winid}.floating true')
+    geom_before = client.get_geometry()
+    x_before, y_before = x11.get_absolute_top_left(client)
+    assert (geom_before.width, geom_before.height) == (300, 200)
+    # move cursor to the top left corner, so we change the
+    # window position and the size (and the bottom right corner is fixed)
```
t-wissmann, line 216: That comment isn't true for zoom because for zoom all corners change
and instead the center of the window is fixed.

t-wissmann: Apart from that, it looks good. I leave it to you whether you also want to add an assertion that the center of the window is preserved.

snke: Good idea, will add that tonight, thanks!

blarz: Thanks I will look into this tonight!

blarz: Like this?

PR #912, opened 2020-05-18
Mention asciidoc filetype in vim modelines

t-wissmann:
It seems that github now respects them for rendering the files online.


t-wissmann:
For comparison, see https://github.com/herbstluftwm/herbstluftwm/blob/master/INSTALL in master vs https://github.com/herbstluftwm/herbstluftwm/blob/asciidoc-vim-modeline/INSTALL in this PR's branch.

PR #916, opened 2020-05-31
Minor changes to release.sh

blarz:
Fix typos and remove an unused statement

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -13,9 +13,6 @@ if git status --porcelain | grep '^ M' ; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -13,9 +13,6 @@ if git status --porcelain | grep '^ M' ; then
     exit 1
 fi
 
-IFS=. read -ra versionargs <<< "$version"
```
blarz, line 16: This doesn't seem to be used anymore?

t-wissmann: Right, `$version` is used as-is.

PR #919, opened 2020-06-06
Add invalid arg tests for raise command

blarz:
Also add check for the argument count, because this method needs at
least the window ID.


blarz:
That test failure is new:
> ../src/rulemanager.cpp: 253: Invalid argument "[b-a]" for rule consequence "keys_inactive": Invalid range in bracket expression.

in `tests/test_keybind.py::test_invalid_keys_inactive_via_rule `

blarz:
> That test failure is new:
> 
> > ../src/rulemanager.cpp: 253: Invalid argument "[b-a]" for rule consequence "keys_inactive": Invalid range in bracket expression.
> 
> in `tests/test_keybind.py::test_invalid_keys_inactive_via_rule `

Ah no, that is not the error.
> assert hlwm.get_attr('tags.0.client_count') == '0'

failed. Do we have race condition in `keyboard.press('x')`?

blarz:
Tried to fix it in #921

PR #921, opened 2020-06-06
tests: Wait for keyboard events to be processed

blarz:
Similar to the whole mouse event handling in the fixtures, add a "wait"
command to keyboard handling, too.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -858,16 +858,25 @@ class MultiscreenDisplay:
 
 
 @pytest.fixture()
-def keyboard():
+def keyboard(hlwm):
     class KeyBoard:
-        def press(self, key_spec):
+        def press(self, key_spec, wait=True):
             subprocess.check_call(['xdotool', 'key', key_spec])
+            if wait:
+                # wait until the keyboard event is processed:
+                hlwm.call('true')
 
-        def down(self, key_spec):
+        def down(self, key_spec, wait=True):
             subprocess.check_call(['xdotool', 'keydown', key_spec])
+            if wait:
+                # wait until the keyboard event is processed:
+                hlwm.call('true')
 
-        def up(self, key_spec):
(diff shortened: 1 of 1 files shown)


t-wissmann:
I think this does not fully solve the issue in https://travis-ci.com/github/herbstluftwm/herbstluftwm/jobs/345436612 because there the code failed with the back trace:
```
hlwm = <conftest.HlwmBridge object at 0x7fc1b31881f0>
keyboard = <conftest.keyboard.<locals>.KeyBoard object at 0x7fc1b3188940>
    def test_invalid_keys_inactive_via_rule(hlwm, keyboard):
        hlwm.call('keybind x close')
        # Note: In future work, we could make this fail right away. But
        # currently, that is not the case.
        hlwm.call('rule once keys_inactive=[b-a]')
        hlwm.create_client()
    
        keyboard.press('x')
    
>       assert hlwm.get_attr('tags.0.client_count') == '0'
E       AssertionError: assert '1' == '0'
E         - 0
E         + 1
tests/test_keybind.py:126: AssertionError
```
Inserting a `hlwm.call('true')` just changes the timing / probabilities a bit, but won't solve any racy condition, because, the next operation here is a hlwm command anyway. The only proper way to make `press()` synchronous is to activate the debugging output in the `keypress`-handler in `xmainloop.cpp` and then wait for it's output and then run the hlwm "ping".


In the above back trace, the actual race codition is the following: running the `close` command only asks the window to close, and does not close it synchronously! That is, it takes some time until the client processes the close request and actually closes the window. I think the easiest we can do here is to replace the `close` command by a command that is fully synchronous, e.g. `add someothertag` and then assert that the command `get_attr tags.by-name.someothertag.name` succeeds.

t-wissmann:
I've replaced the command in #926.

blarz:
> I've replaced the command in #926.

Awesome, thank you. Closing this PR.

PR #923, opened 2020-06-07
codecov: Remove -y flag

blarz:
The flag is deprecated:

DeprecationWarning: The -y flag is no longer supported by Codecov.
  codecov.yml must be located underneath the root, dev/, or .github/
  directorie

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ jobs:
                 focal sh -c '
                     /hlwm/ci/build.py --build-dir=/hlwm/build --cxx=g++-9 --cc=gcc-9 --build-type=Debug --ccache=/.ccache --compile --run-tests
                 '
-              - bash <(curl -s https://codecov.io/bash) -y .codecov.yml -f coverage.info
+              - bash <(curl -s https://codecov.io/bash) -f coverage.info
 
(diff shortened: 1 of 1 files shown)


dnnr:
Doesn't this mean that we need to rename the file as well? There reason for using `-y` was that our file wasn't named exactly the way CodeCov expects it to be (which is: `codecov.yml`)

blarz:
> Doesn't this mean that we need to rename the file as well? There reason for using `-y` was that our file wasn't named exactly the way CodeCov expects it to be (which is: `codecov.yml`)

It does seem to find it though:
```
project root: .
Yaml found at: ./.codecov.yml
```

blarz:
```git ls-files "*codecov.yml" "*codecov.yaml" 2>/dev/null \``` is how it checks for the file.

blarz:
> Doesn't this mean that we need to rename the file as well? There reason for using `-y` was that our file wasn't named exactly the way CodeCov expects it to be (which is: `codecov.yml`)

https://docs.codecov.io/docs/codecov-yaml#can-i-name-the-file-codecovyml

dnnr:
I see, thanks :+1:

PR #926, opened 2020-06-08
Use synchronous commands in keys_inactive tests

t-wissmann:
The 'close' command is asynchronous, because it only sends the client a
request to close the window, and it takes some time until the client
reacts. Instead we use the 'add' command which synchronously creates a
new tag.

### DIFF
--- a/tests/test_keybind.py
+++ b/tests/test_keybind.py
@@ -102,7 +102,9 @@ def test_keys_inactive(hlwm, keyboard, maskmethod, whenbind, refocus):
 
     keyboard.press('x')
 
-    # Expect client to have quit because of received keypress:
+    # we expect that the keybind command is not executed:
+    assert hlwm.list_children('tags.by-name.') == ['default']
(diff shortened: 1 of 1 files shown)


tests/test_keybind.py
```diff
@@ -115,15 +117,17 @@ def test_keys_inactive(hlwm, keyboard, maskmethod, whenbind, refocus):
 
 
 def test_invalid_keys_inactive_via_rule(hlwm, keyboard):
-    hlwm.call('keybind x close')
+    hlwm.call('keybind x add anothertag')
     # Note: In future work, we could make this fail right away. But
     # currently, that is not the case.
     hlwm.call('rule once keys_inactive=[b-a]')
     hlwm.create_client()
```
blarz, line 124: This can be removed now, right?

tiosgz: And where would the client with `keys_inactive` be then?

blarz: You're right, that was a brainfart.

PR #927, opened 2020-06-08
Update flake8 to latest version and fix warnings

blarz:


blarz:
l is an ambiguous variable name (I agree), so it had to be renamed

PR #929, opened 2020-06-09
Unpin pytest-xvfb / PyVirtualDisplay

The-Compiler:
PyVirtualDisplay v2.0.0 is out now, which is compatible with the newest PyVirtualDisplay again.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,8 +20,8 @@ skipsdist = true
 commands = {envpython} -m pytest {posargs}
 deps =
     ewmh
-    pytest-xvfb == 1.2.0
-    PyVirtualDisplay < 1.0 # pytest-xvfb <= 1.2.0 breaks with versions >= 1.0
+    pytest-xvfb
+    PyVirtualDisplay
     pytest-xdist
     python-xlib
 


blarz:
Awesome, thank you @The-Compiler!
Just looking into the build error, cannot reproduce it locally.

t-wissmann:
Same for me locally, works again, thanks! (strangely, the tests are quite slow, but I guess that's a different issue or some mistake in my setup)

t-wissmann:
The tests now fail with:
```python
    @pytest.fixture()
    def hlwm_spawner(tmpdir):
        """yield a function to spawn hlwm"""
>       assert os.environ['DISPLAY'] != ':0', 'Refusing to run tests on display that might be your actual X server (not Xvfb)'
E       AssertionError: Refusing to run tests on display that might be your actual X server (not Xvfb)
E       assert ':0' != ':0'
```
Is there a way to define the display number? Or does xvfb pick it automatically? (If so, we should make this `!= ':0'`-assertion configurable with a `--allow-display-0` flag or so)

blarz:
There are a lot of changes in PyVirtualDisplay, did some defaults change maybe? Cannot see the changelog..
https://github.com/ponty/PyVirtualDisplay#usage

blarz:
> Is there a way to define the display number? Or does xvfb pick it automatically? (If so, we should make this `!= ':0'`-assertion configurable with a `--allow-display-0` flag or so)

I'd expect that Xvfb/PyVirtualDisplay (or whoever opens the display) will pick up a display number that is >0. I guess something isn't working as expected, allowing display to be 0, is probably not the fix.

blarz:
> I think you could just use the `xvfb` fixture and do `assert xvfb is not None` instead?

I did that in #931, thanks!

The-Compiler:
Since #931 is in, I rebased on the latest master now.

t-wissmann:
> Since #931 is in, I rebased on the latest master now.

We're doing squash-merges anyway, so clicking on 'Update Branch' would have been sufficient

The-Compiler:
Good to know! Looks like one Travis build is failing with:

```
[gw0] [ 81%] PASSED tests/test_x11.py::test_client_moveresizes_itself xterm: fatal IO error 11 (Resource temporarily unavailable) or KillClient on X server ":0"
```

Not sure if that's related or not to the update?

t-wissmann:
```
[gw0] [ 81%] PASSED tests/test_x11.py::test_client_moveresizes_itself xterm: fatal IO error 11 (Resource temporarily unavailable) or KillClient on X server ":0"

No output has been received in the last 10m0s, this potentially indicates
a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on:
https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
The build has been terminated
```
I think this is unrelated. For some unknown reason, the build got stuck (this happens from time to time, and I'm not sure whether this is our fault or that of pytest (note that this did not happen *during* a test case)

blarz:
I cannot even retrigger the build, or am I overlooking something?

t-wissmann:
> I cannot even retrigger the build, or am I overlooking something?

One can click 'Restart Job' on the top right of the build log page. I had indeed restarted the job ≥half an hour ago and it errored again. I've just gave it another try...

Issue #914, opened 2020-05-29
Mark target window urgent if focus_stealing_prevention is active


t-wissmann:
If focus_stealing_prevention is active and prevents a window being activated, we should at list mark it urgent (instead of doing nothing).

blarz:
Fixed with #934

PR #934, opened 2020-06-21
Mark client urgent if focus stealing is not active

blarz:


blarz:
Tried to address #914.

I think my assumptions on how this should work are wrong,
at least the test I wrote doesn't do what I expect it to do.

"activating" (NetActiveWindow) means also "starting" an application, right?
I thought creating a client on another tag should at least make it urgent.

Or did I understand the issue wrong?

blarz:
Ok seems like starting the application does not trigger the activate thing, but for example opening a URL with from a terminal, which opens a new tab in firefox does.

So the change as such seems to work, I just don't know how to test it.

t-wissmann:
> Ok seems like starting the application does not trigger the activate thing, but for example opening a URL from a terminal, which opens a new tab in firefox, does.
> 
> So the change as such seems to work, I just don't know how to test it.

I've changed it in your PR with the commit 9d10c70

t-wissmann:
NetActiveWindow is the request to focus a particular window (e.g. task bars do this), but in fact any window can send such a request. The source-field usually tells the window manager whether the sender of the request considers itself as a 'taskbar/pager/...' or as an application window.

PR #935, opened 2020-06-22
X Properties with format 32 are arrays of long

t-wissmann:
The documentation for XGetWindowProperty[1] specifies that if the
format parameter is 32, then the data array must be an array of long.

The present commit fixes this both in the herbstluftwm and in the
herbstclient code.

[1] https://tronche.com/gui/x/xlib/window-information/XGetWindowProperty.html


siccegge:
Hi

I can confirm this fixes that problem (herbstclient does not detect hlwm running) at least on s390x

Thanks!

t-wissmann:
> I can confirm this fixes that problem (herbstclient does not detect hlwm running) at least on s390x

Great, thanks for the hints, @siccegge!

PR #940, opened 2020-06-25
Do not set focused clients urgent

t-wissmann:

PR #941, opened 2020-06-25
Test #940: the focused client is not made urgent

t-wissmann:


t-wissmann:
This should fail until #940 is merged.

blarz:
> This should fail until #940 is merged.

Merged, updated your branch. Let's wait for the build result and this can go in, too.

Issue #936, opened 2020-06-22
Testsuite fails due to double imprecision


siccegge:
Test failure on 32bit x86 due to float imprecision:

```
=================================== FAILURES ===================================
____________________________ test_dump_frame_index _____________________________

hlwm = <conftest.HlwmBridge object at 0xf68d78f8>

    def test_dump_frame_index(hlwm):
        layout = {}
        layout['00'] = "(clients vertical:0)"
        layout['01'] = "(clients grid:0)"
        layout['0'] = '(split vertical:0.7:0 {} {})'.format(
            layout['00'], layout['01'])
        layout['1'] = '(clients horizontal:0)'
        layout[''] = '(split horizontal:0.65:0 {} {})'.format(
            layout['0'], layout['1'])
        hlwm.call(['load', layout['']])
    
        tag = hlwm.get_attr('tags.focus.name')
        for index in ["", "0", "1", "00", "01"]:
>           assert hlwm.call(['dump', tag, index]).stdout == layout[index]
E           AssertionError: assert '(split horiz...orizontal:0))' == '(split horizo...orizontal:0))'
E             Skipping 33 identical leading characters in diff, use -v to show
E             - ertical:0.6999:0 (clients vertical:0) (clients grid:0)) (clients horizontal:0))
E             ?           ^^^^
E             + ertical:0.7:0 (clients vertical:0) (clients grid:0)) (clients horizontal:0))
E             ?           ^

../tests/test_layout.py:139: AssertionError
---------------------------- Captured stdout setup -----------------------------
hlwm started
---------------------------- Captured stderr setup -----------------------------
EWMH: 1 desktops: 'default'
0 managed clients: 
----------------------------- Captured stdout call -----------------------------
Client command ['load', '(split horizontal:0.65:0 (split vertical:0.7:0 (clients vertical:0) (clients grid:0)) (clients horizontal:0))'] succeeded (no output)
Client command ['get_attr', 'tags.focus.name'] succeeded with output:
default
Client command ['dump', 'default', ''] succeeded with output:
(split horizontal:0.65:0 (split vertical:0.6999:0 (clients vertical:0) (clients grid:0)) (clients horizontal:0))
--------------------------- Captured stdout teardown ---------------------------
Client command ['version'] succeeded with output:
herbstluftwm 0.8.3
Copyright (c) 2011-2014 Thorsten Wißmann
Released under the Simplified BSD License
xrandr support: on
xinerama support: on

[...]

PR #942, opened 2020-06-27
Use fixed precision decimals for frame fractions

t-wissmann:
We introduce a new class FixPrecDec that encapsulates in- and output of
fixed precision decimal numbers, and we use them for reading and
printing the fractions in the frame layout tree.

This solves #936.

### DIFF
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -825,7 +825,13 @@ int FrameTree::splitCommand(Input input, Output output)
         return HERBST_NEED_MORE_ARGS;
     }
     bool userDefinedFraction = input >> strFraction;
-    FixPrecDec fraction = Converter<FixPrecDec>::parse(strFraction);
+    FixPrecDec fraction = FixPrecDec::fromInteger(0);
+    try {
+        fraction = Converter<FixPrecDec>::parse(strFraction);
+    }  catch (const std::exception& e) {
+        output << "invalid argument: " << e.what() << endl;
+        return HERBST_INVALID_ARGUMENT;
+    }
     fraction = FrameSplit::clampFraction(fraction);
     auto frame = focusedFrame();
     int lh = frame->lastRect().height;
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -7,6 +7,7 @@
 
 #include "client.h"
 #include "completion.h"
+#include "fixprecdec.h"
 #include "framedata.h"
 #include "frameparser.h"
 #include "ipc-protocol.h"
@@ -56,7 +57,7 @@ void FrameTree::dump(shared_ptr<Frame> frame, Output output)
             << "(split "
             << Converter<SplitAlign>::str(s->align_)
             << ":"
-            << ((double)s->fraction_) / (double)FRACTION_UNIT
+            << Converter<FixPrecDec>::str(s->fraction_)
             << ":"
             << s->selection_
             << " ";
@@ -210,7 +211,7 @@ int FrameTree::rotateCommand() {
                     s->align_ = SplitAlign::vertical;
                     s->selection_ = s->selection_ ? 0 : 1;
                     swap(s->a_, s->b_);
-                    s->fraction_ = FRACTION_UNIT - s->fraction_;
+                    s->fraction_ = FixPrecDec::fromInteger(1) - s->fraction_;
                     break;
             }
         };
@@ -261,7 +262,8 @@ shared_ptr<TreeInterface> FrameTree::treeInterface(
         size_t childCount() override { return 2; };
         void appendCaption(Output output) override {
             output << " " << Converter<SplitAlign>::str(s_->align_)
-                   << " " << (s_->fraction_ * 100 / FRACTION_UNIT) << "%"
+                   << " " << (s_->fraction_.value_ * 100 / s_->fraction_.unit_)
+                   << "%"
                    << " selection=" << s_->selection_;
         }
     private:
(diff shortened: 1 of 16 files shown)


t-wissmann:
There is still some exception handling missing, because the former parser `atof()` didn't complain about anything.

t-wissmann:
Now it's ready :-)

src/frametree.cpp
```diff
@@ -819,13 +820,19 @@ vector<SplitMode> SplitMode::modes(SplitAlign align_explode, SplitAlign align_au
...
         return HERBST_NEED_MORE_ARGS;
     }
     bool userDefinedFraction = input >> strFraction;
-    double fractionFloat = userDefinedFraction ? atof(strFraction.c_str()) : 0.5;
-    int fraction = FrameSplit::clampFraction(FRACTION_UNIT * fractionFloat);
+    FixPrecDec fraction = FixPrecDec::fromInteger(0);
+    try {
+        fraction = Converter<FixPrecDec>::parse(strFraction);
+    }  catch (const std::exception& e) {
+        output << "invalid argument: " << e.what() << endl;
```
dnnr, line 832: I think most (all?) of our error messages start with a capital letter. I'd suggest sticking to that style here.

dnnr: Oh and maybe even be more explicit about the problem. I recall being frustrated whenever a tool tells me that it doesn't like my input, but does not tell me what the problem is. For example, it could say "Cannot parse as decimal number: " instead.

tests/test_layout.py
```diff
@@ -735,3 +738,69 @@ def test_tree_style_utf8(hlwm):
...
     """)
+
+
+def test_split_invalid_argument(hlwm):
+    wrongDecimal = [
+        '0.0f', '.3', '.', 'b',
+    ]
+    for d in wrongDecimal:
+        hlwm.call_xfail(['split', 'top', d]) \
+            .expect_stderr('invalid argument')
```
dnnr, line 749: I vaguely recall that we once missed a bug because we did not check if an expected error occurred for the right reason. Maybe we should verify the exception string here as well?

src/frametree.cpp
t-wissmann: > I think most (all?) of our error messages start with a capital letter. I'd suggest sticking to that style here.

It's not entirely consistent. On `master`, we have:
```bash
$ grep -ni 'invalid argument' *.cpp
frametree.cpp:565:        output << "invalid argument: " << e.what() << endl;
monitor.cpp:370:                ": Invalid argument \"" << argv[1] << "\"\n";
rulemanager.cpp:253:                    HSWarning("Invalid argument \"%s\" for rule consequence \"%s\": %s\n",
tag.cpp:283:        output << "invalid argument: " << e.what() << endl;
```

tests/test_layout.py
t-wissmann: > I vaguely recall that we once missed a bug because we did not check if an expected error occurred for the right reason. Maybe we should verify the exception string here as well?

First I was too lazy because the messages were different, but now I have extended it (and I added some more cases to test all exceptions possibly thrown by the `parse()`)

src/frametree.cpp
t-wissmann: And I'd like to add that I also want to simplify the argument parsing anyway, because I'm re-writing the same code over and over again. You're partly right with your second comment: the error message says what's wrong with the argument (e.g. "only digits after '.'"), but it doesn't say *which* argument could not be parsed. But I'd address both these two issues in a separate PR.

tests/test_layout.py
```diff
@@ -735,3 +738,78 @@ def test_tree_style_utf8(hlwm):
...
+        ('0.+8', "After '.' only digits"),
+        ('0.-8', "After '.' only digits"),
+        ('0..0', "A decimal must have at most one '.'"),
+        ('.3', "There must be at least one digit"),
+        ('.', "There must be at least one digit"),
+        ('b', "stoi"),
+        ('-.3', "stoi"),
+        ('+.8', "stoi"),
+    ]
+    for d, msg in wrongDecimal:
```
The-Compiler, line 756: Why those for loops rather than `@pytest.mark.parametrize` (which will run the tests isolated from each other and don't e.g. abort the remaining cases if one fails)?

t-wissmann:
> Why those for loops rather than `@pytest.mark.parametrize` (which will run the tests isolated from each other and don't e.g. abort the remaining cases if one fails)?

For performance reasons. For these tests, there's no need to run them
isolated, so we save time & energy by not booting up hlwm for each of
these cases (I've already done this in other test cases). None of them
should fail anyway for most PRs in the future.

tests/test_layout.py
dnnr: Let me continue the discussion here. :-)

> For performance reasons. 

That's an understandable reason. @The-Compiler I think we should take this as input to explore if we can achieve this runtime behavior without having it bleed into the test definitions like this.

The-Compiler: Hmm, I see! I guess we could have a `hlwm_shared` fixture (or whatever) with `scope='module'`, but that'd then be the same hlwm process for the entire test module. Also see: https://github.com/pytest-dev/pytest/issues/1552

t-wissmann: Isn't this makeing the testsuite more complicated than it needs to be? Or asking the other way round: where is the issue with the test case as it is at the moment?

dnnr: IMHO it's not *much* of an issue, but I don't think it is desirable that you have to take these things into account when writing tests. This should be abstracted away.

Just to be clear: I'm quite fine with how this looks right now, but I'd be happier if the test framework removed the need for you to make that choice (and get the same performance advantage).

An ideal outcome would be that there is much less restarting of hlwm happening in a full test run, because I think a lot of tests do not strictly need a fresh instance.

blarz:
This PR looks good to me. Anything left to discuss?

t-wissmann:
I'm not aware of any open issue. I'd also say that it's ready

PR #945, opened 2020-07-02
ArgParse utility for processing the Input in commands

t-wissmann:
The new ArgParse class allows to define positional command line arguments in a
type-safe way. The benefit of it is not only shorter code, but also more
verbose and more consistent error message in the case of parser errors.


t-wissmann:
I'd like to reduce redundancy in the code, and this is the best I came up with. I've already used this new `InputConverter` class to some of the commands. There are still some things that are questionable:

  - My usage of `operator int()` for the return code may be inconsistent. I'm not aware of the C++ language specification rules, but note that for `InputConverter ip`, the terms `(bool)(int)ip` and `(bool)ip` always evaluate to *different* values.
  - InputConverter is yet another class. I couldn't find a clean way to incorporate this into the `Input` class, because I didn't want `Input` to require on `Output`

t-wissmann:
Regarding the use of `operator>>()` here:

The advantage is that it matches nicely with `Input::operator>>()`, so one can even mix them (note that in `splitCommand`, I check `input.empty()` after calling `InputConvert::operator>>()`; I don't claim that this is perfect programming style, but it was the best compromise I could find for the moment)

The disadvantage is, that the `Optional()` marker is a bit ugly. We could also replace `>>` by function calls which would change something like
```C++
inpconv >> firstarg >> secondarg >> InputConvert::Optional() >> optionalarg1 >> optionalarg2;
```
into something roughly as follows:
```C++
inpconv.mandatory(firstarg).mandatory(secondarg).optional(optionalarg1).optional(optionalarg2);
```

t-wissmann:
> * My usage of `operator int()` for the return code may be inconsistent. I'm not aware of the C++ language specification rules, but note that for `InputConverter ip`, the terms `(bool)(int)ip` and `(bool)ip` always evaluate to _different_ values.

The linter agrees that this is very confusing and complains about
```C++
    LayoutAlgorithm layout;
    if (!(inpconv >> layout)) {
        return inpconv;
    }

    auto curFrame = focusedFrame();
    curFrame->setLayout(layout);
```
with the following:
```
/hlwm/src/frametree.cpp:764:5: error: 1st function call argument is an uninitialized value [clang-analyzer-core.CallAndMessage,-warnings-as-errors]
    curFrame->setLayout(layout);
    ^
/hlwm/src/frametree.cpp:758:5: note: 'layout' declared without an initial value
    LayoutAlgorithm layout;
    ^
/hlwm/src/frametree.cpp:759:11: note: Calling 'InputConvert::operator>>'
    if (!(inpconv >> layout)) {
          ^
/hlwm/src/inputconvert.h:30:13: note: Assuming field 'errorCode_' is equal to 0
        if (errorCode_ != 0) {
            ^
/hlwm/src/inputconvert.h:30:9: note: Taking false branch
        if (errorCode_ != 0) {
        ^
/hlwm/src/inputconvert.h:36:13: note: Assuming the condition is true
        if (!(input_ >> valueString)) {
            ^
/hlwm/src/inputconvert.h:36:9: note: Taking true branch
        if (!(input_ >> valueString)) {
        ^
/hlwm/src/inputconvert.h:41:17: note: Assuming field 'onlyOptionalArgumentsRemain_' is true
            if (!onlyOptionalArgumentsRemain_) {
                ^
/hlwm/src/inputconvert.h:41:13: note: Taking false branch
            if (!onlyOptionalArgumentsRemain_) {
            ^
/hlwm/src/inputconvert.h:44:13: note: Returning without writing to 'value'
            return *this;
            ^
/hlwm/src/frametree.cpp:759:11: note: Returning from 'InputConvert::operator>>'
    if (!(inpconv >> layout)) {
          ^
/hlwm/src/frametree.cpp:759:10: note: Calling 'InputConvert::operator bool'
    if (!(inpconv >> layout)) {
         ^
/hlwm/src/inputconvert.h:64:9: note: Returning the value 1, which participates in a condition later
        return errorCode_ == 0;
        ^
/hlwm/src/frametree.cpp:759:10: note: Returning from 'InputConvert::operator bool'
    if (!(inpconv >> layout)) {
         ^
/hlwm/src/frametree.cpp:759:5: note: Taking false branch
    if (!(inpconv >> layout)) {
    ^
/hlwm/src/frametree.cpp:764:5: note: 1st function call argument is an uninitialized value
    curFrame->setLayout(layout);
    ^
```

t-wissmann:
@dnnr I've now transformed this into a proper `ArgParse` object, where one first defines the format (order, optionality) in an ArgParse object and only in the final step, the `Input` is fed into the `ArgParse`.

There are two patterns used here, and I don't know which one I prefer. The first one is:
```C++
 Direction direction = Direction::Left;
 FixPrecDec delta = FixPrecDec::approxFrac(1, 50); // 0.02
 auto ap = ArgParse().mandatory(direction).optional(delta);
 if (ap.parseOrExit(input, output)) {
    return ap.exitCode();
 }
```
Here, the `=` might be a bit tricky, because I think there are actually two `ArgParse` objects involved, and I don't know the precise scope of the anonymous object created by `ArgParse()`. The safer but slightly more verbose version of that code would be the following:
```C++
 Direction direction = Direction::Left;
 FixPrecDec delta = FixPrecDec::approxFrac(1, 50); // 0.02
 ArgParse ap;
 ap.mandatory(direction).optional(delta);
 if (ap.parseOrExit(input, output)) {
    return ap.exitCode();
 }
```
It even needs two characters less but a line more... (Note that I need to give the object a name `ap`, because otherwise I don't see how to apply this `if (...) { return ..; }` pattern.)

src/argparse.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+using std::string;
+
+/**
+ * @brief try to parse the arguments
+ * @param input
+ * @param output
+ * @return return whether there has been an error (true = error, false = no error)
+ */
+bool ArgParse::parseOrExit(Input& input, Output& output)
```
blarz, line 13: The name is a bit confusing IMHO, especially if you see it in use.
if(parseOrExit) is not obvious just by looking at the method name, only after consulting the doc string.
Could be a bit more obvious, e.g. inputParsed, or another name that reflects some kind of success/failure?
What do you think?

src/argparse.h
```diff
@@ -0,0 +1,78 @@
+#ifndef INPUTCONVERT_H
```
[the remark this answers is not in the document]
t-wissmann: Right! I've replaced it. (I thought we once agreed on `#pragma once`, but as I just saw, it's not part of the `C++11` standard, so I'd rather avoid it)

src/argparse.cpp
t-wissmann: I've renamed it to `parsingFails()`, because then something like this reads well:
```C++
if (ap.parsingFails(input, output)) {
    return ap.exitCode();
}
```
(For `inputParsed()` I think its ambigious whether `true` means "error" or "success")

src/argparse.h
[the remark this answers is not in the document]
blarz: Thanks!

src/argparse.cpp
blarz: That is definitely better!

PR #946, opened 2020-07-04
Wait for completion of move_into in mouse-based test

siccegge:


siccegge:
On closer inspection iff I do it correctly also always fails for me.
Might be something interesting going on with enternotify (as it has some disable logic all over the place)

t-wissmann:
Below is the failing test case, there's a FIXME describing the issue, saying that "hlwm doesn't get an EnterNotify event in xvfb but it works in Xephyr". So I think we should pass `wait=False` to `move_into` here.
```
@pytest.mark.parametrize("click,focus_follows_mouse", [
        (False, False),
        (True, False),
        (True, True)
        #  FIXME: here, hlwm doesn't get an EnterNotify event in xvfb
        # but it works in Xephyr.
        # (False, True)
    ])
    def test_focus_client_by_decoration(hlwm, mouse, x11, click, focus_follows_mouse):
        hlwm.call('attr theme.border_width 50')
        hlwm.call('attr theme.active.color red')
        hlwm.call('attr theme.normal.color blue')
        hlwm.call('set_layout vertical')
        hlwm.call(['set', 'focus_follows_mouse', hlwm.bool(focus_follows_mouse)])
        winids = hlwm.create_clients(2)
        hlwm.call(['jumpto', winids[0]])
        assert hlwm.get_attr('clients.focus.winid') == winids[0]
    
        # move mouse to the decoration:
        decwin = x11.get_decoration_window(x11.window(winids[1]))
        # here, xdotool hangs on the second invokation of this testcase
        mouse.move_into(x11.winid_str(decwin), 10, 10)
```

siccegge:
> Below is the failing test case, there's a FIXME describing the issue, saying that "hlwm doesn't get an EnterNotify event in xvfb but it works in Xephyr". So I think we should pass `wait=False` to `move_into` here.

There's anjother one that explicitely tells it's important to not sync. With that "fixed" as well it looks good now

```python
@pytest.mark.parametrize("client_count", [7, 8, 9, 10])
def test_enternotify_do_not_drop_events(hlwm, mouse, client_count):
    """test that when triggering multiple enter notify events
    that the last enter notify event survives
    """
    hlwm.call('set focus_follows_mouse on')
    winid = hlwm.create_clients(client_count)

    # place many enter notify events in the event queue
    for i in range(0, client_count):
        # here, it's important that move_into does not sync with hlwm
        # such that the event queue in hlwm builds up
        mouse.move_into(winid[i], 10, 10, wait=False)

    # finally, all enter notify events must survive
    assert hlwm.get_attr('clients.focus.winid') == winid[client_count - 1]
```

t-wissmann:
If I see it correctly, then there is only one `move_into()` call left which does not specify `wait=False`, namely in `test_focus_follows_mouse`. I'm wondering why the waiting works there but not for the other test cases...

t-wissmann:
The fix I was thinking of is done in PR #952.

PR #951, opened 2020-07-09
Add pgid (process group id) as rule condition

wilriker:
I use a slightly modified version of `exec_on_tag.sh` to start applications. This works mostly fine with the exception of a starter application forking into the actual application (e.g. eclipse, IntelliJ Idea, Microsoft Teams). For these cases `pgid` is a lot more usefull as it's the same for all applications. With an adjusted timeout and the `once` modifier removed from the rule this leads to the desired behavior of not only having the splash screen but also the actual application spawn on the tag where it was started.

I recon that getting the `pgid` is mostly not trivial but e.g. my small tool [getpgid](https://github.com/wilriker/getpgid) can be used for this purpose.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1174,6 +1174,10 @@ Valid 'properties' are:
     the client's process id (Warning: the pid is not available for every client.
     This only matches if the client sets _NET_WM_PID to the pid itself).
 
++pgid+::
+    this client's process group id. Since the pgid of a window is  is derived from
+    its +pid+ the same restrictions apply as above.
+
 +maxage+::
     matches if the age of the rule measured in seconds does not exceed 'value'.
     This condition only can be used with the +=+ operator. If maxage already is
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1175,7 +1175,7 @@ Valid 'properties' are:
     This only matches if the client sets _NET_WM_PID to the pid itself).
 
 +pgid+::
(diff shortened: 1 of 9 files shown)


t-wissmann:
Thanks! The diff looks good to me!

There are some minor CI-issues:

1. C headers should be included before C++ headers:
```
+#include <unistd.h>
 #include <climits>
 #include <iostream>
-#include <unistd.h>
```

2. The `numeric_props` array in `tests/test_rules.py` needs to be extended by `pgid`.

3. Can you please also add documentation on it in `doc/herbstluftwm.txt` next to the documentation on `pid`? (Please also mention that `pgid` only works if the assumptions in `pid` are met).

4. While you are at it, you can also extend the `NEWS` file by  a line `new rule condition 'pgid'` :-)

wilriker:
All requested changes applied. :-)

doc/herbstluftwm.txt
```diff
@@ -1174,6 +1174,10 @@ Valid 'properties' are:
     the client's process id (Warning: the pid is not available for every client.
     This only matches if the client sets _NET_WM_PID to the pid itself).
 
++pgid+::
+    this client's process group id. Since the pgid of a window is  is derived from
```
blarz, line 1178: typo "is is"

wilriker: Fixed in 218cbb0.

blarz: Thanks!

wilriker:
Thanks! :-)

PR #962, opened 2020-07-21
Create Xvfb manually in the test cases

t-wissmann:
This introduces a new fixture 'xvfb' that creates an Xvfb instance
whose scope is only the test case and not the entire test session. The
python code for this has already been written in the MultiscreenDisplay
class.

Having the Xvfb scope only within the testcase gets rid of the necessity
to kill existing/remaining windows at the beginning and the end of the
testcase run.

Moreover, I am passing -noreset to Xvfb, and so the X properties are not
reset when the last client window closes, hence we don't need this
x11_connection fixture hack anymore.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -509,35 +509,6 @@ def hc_idle(hlwm):
     hc.shutdown()
 
 
-def kill_all_existing_windows(show_warnings=True):
-    xlsclients = subprocess.run(['xlsclients', '-l'],
-                                stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE,
-                                check=False)
-    print(xlsclients.stderr.decode(), file=sys.stderr, end='')
-    if re.search('unable to open display', xlsclients.stderr.decode()):
-        # if the display is already closed since there are no clients left,
-        # then that's fine for us
-        return
-    else:
-        # otherwise, assert successfull termination
-        assert xlsclients.returncode == 0
-    clients = []
-    for line in xlsclients.stdout.decode().splitlines():
-        m = re.match(r'Window (0x[0-9a-fA-F]*):', line)
-        if m:
-            clients.append(m.group(1))
-    if clients and show_warnings:
-        warnings.warn(UserWarning("There are still some clients "
-                                  "from previous tests."))
-    for c in clients:
-        if show_warnings:
-            warnings.warn(UserWarning("Killing " + c))
-        # send close and kill ungently
-        subprocess.run(['xdotool', 'windowclose', c])
-        subprocess.run(['xdotool', 'windowkill', c])
-
-
 @pytest.fixture()
 def hlwm_spawner(tmpdir):
     """yield a function to spawn hlwm"""
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -577,12 +577,10 @@ def xvfb():
 def hlwm_process(hlwm_spawner, xvfb):
     """Set up hlwm and also check that it shuts down gently afterwards"""
     hlwm_proc = hlwm_spawner(['--no-tag-import'], display=xvfb.display)
-    # kill_all_existing_windows(show_warnings=True)
 
     yield hlwm_proc
 
     hlwm_proc.shutdown()
-    # kill_all_existing_windows(show_warnings=False)
 
 
 @pytest.fixture(params=[0])
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,7 +16,6 @@ import time
 import types
 
 import pytest
-import warnings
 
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -564,15 +563,26 @@ def spawn(args=[], display=None):
...
+        yield xserver
+
+
+@pytest.fixture()
+def hlwm_process(hlwm_spawner, xvfb):
     """Set up hlwm and also check that it shuts down gently afterwards"""
-    hlwm_proc = hlwm_spawner(['--no-tag-import'])
-    kill_all_existing_windows(show_warnings=True)
+    hlwm_proc = hlwm_spawner(['--no-tag-import'], display=xvfb.display)
+    # kill_all_existing_windows(show_warnings=True)
```
blarz, line 580: This can be removed, right?

tests/conftest.py
```diff
@@ -564,15 +563,26 @@ def spawn(args=[], display=None):
...
 
     yield hlwm_proc
 
     hlwm_proc.shutdown()
-    kill_all_existing_windows(show_warnings=False)
+    # kill_all_existing_windows(show_warnings=False)
```
blarz, line 585: Same here

blarz:
Also the commit message says

>  that creates and Xvfb instance

should that be "creates an Xvfb instance"?

blarz:
Tested it locally, something seems off in regards to the open file descriptors.
> ~/git/herbstluftwm/build ((9f5920a...))$ tox -c ..  -e py38 -- -v

gives me 

```
E       OSError: [Errno 24] Too many open files

/usr/lib/python3.8/subprocess.py:1603: OSError
----------------------------------------------------------------------------------------------------------- Captured stdout setup -----------------------------------------------------------------------------------------------------------
Running: Xvfb -nolisten tcp -screen 0 800x600x8 -noreset -displayfd 999
Xvfb is using the display ":1"
hlwm started
----------------------------------------------------------------------------------------------------------- Captured stderr setup -----------------------------------------------------------------------------------------------------------
_XSERVTransSocketUNIXCreateListener: ...SocketCreateListener() failed
_XSERVTransMakeAllCOTSServerListeners: server already running
EWMH: 0 desktops:
0 managed clients: 
--------------------------------------------------------------------------------------------------------- Captured stdout teardown ----------------------------------------------------------------------------------------------------------
Client command ['version'] succeeded with output:
herbstluftwm 0.8.3 (9f5920a built on 2020-07-21)
Copyright (c) 2011-2014 Thorsten Wißmann
Released under the Simplified BSD License
xrandr support: on
xinerama support: on

--------------------------------------------------------------------------------------------------------- Captured stderr teardown ----------------------------------------------------------------------------------------------------------
interrupted by signal 15
/home/snke/git/herbstluftwm/src/main.cpp: 465: Interrupted by signal 15
/home/snke/git/herbstluftwm/src/xconnection.cpp: 36: Closing display
========================================================================================================== short test summary info ==========================================================================================================
FAILED tests/test_root_commands.py::test_new_attr_without_removal[-my_test-string] - OSError: [Errno 24] Too many open files
```
The tox process hits my limit of 1024 open fds.

tests/conftest.py
t-wissmann: Right, I've removed both! Should I also remove the `kill_all_existing_windows()` function? It is not used anymore, but I can't exclude that it might be of use in the future.

t-wissmann:
I've fixed the message body.

Thanks for testing it locally :) I'm now closing both pipe file descriptors in `MultiscreenDisplay`, maybe this was the root of the issue. Is there a `ulimit`-oneliner such that I can also test it here with the fd-limit?

tests/conftest.py
blarz: I'd remove it, yes.
If we will need it again, we have git to restore it :)

blarz:
> I've fixed the message body.

Thanks!

> Thanks for testing it locally :) I'm now closing both pipe file descriptors in `MultiscreenDisplay`, maybe this was the root of the issue. Is there a `ulimit`-oneliner such that I can also test it here with the fd-limit?

Yes, closing the fd fixed it for me, thanks!
According to ulimit -a, the correct flag is ulimit -n. My limit for open files is 1024.

tests/conftest.py
t-wissmann: Aaaand it's gone :)

blarz: Thanks :D

blarz:
Seems like one flake8 error is left

t-wissmann:
Oh right, I've fixed that!

t-wissmann:
Oh, I had accidentally pushed it to master... But it should be OK now, hopefully..

PR #964, opened 2020-07-23
Quickfix race-condition in hc --wait testcases

t-wissmann:
There is a race-condition in both test cases for herbstclient --wait
that comes from the fact that we don't know how fast herbstclient
connects to hlwm and starts listening for hooks. In this commit, we
simply hope that `hc --wait` boots up within the time needed for two `hc
true` calls.

This issue only exists for --wait and not for --idle, because for
--wait, we need to test that herbstclient indeed exits after a fixed
number of hooks (--count).

For hc --idle, this issue does not appear in the HcIdle class in
conftest.py, because there, we can wait for hc --idle to boot up by
calling 'emit_hook' multiple times (hence, this is not applicable for
synchronizing the bootup of --wait).

I assume that this fixed the issue #954 of the erroring testsuite (at
least it tackles the backtraces gathered there so far).

### DIFF
--- a/tests/test_herbstclient.py
+++ b/tests/test_herbstclient.py
@@ -71,16 +71,23 @@ def test_herbstclient_help_on_stdout():
     assert re.search(r'--wait', result.stdout)
 
 
-@pytest.mark.parametrize('num_hooks_sent', [4])
+@pytest.mark.parametrize('repeat', range(0, 2))  # use more repetitions to test the race-condtion
+@pytest.mark.parametrize('num_hooks_sent', [6])  # see below comment on the race-condition
 @pytest.mark.parametrize('num_hooks_recv', range(1, 4 + 1))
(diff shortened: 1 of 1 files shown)


blarz:
Could we fix this another way? Maybe instead of hoping that hlwm(true) fixes this race condition,
we could add a -V flag (the counterpart to -q) and explicitly print (to stderr?) that we successfully connected to hlwm.

We could then wait for that output in the tests.

t-wissmann:
> Could we fix this another way? Maybe instead of hoping that hlwm(true) fixes this race condition,
> we could add a -V flag (the counterpart to -q) and explicitly print (to stderr?) that we successfully connected to hlwm.
> 
> We could then wait for that output in the tests.

Yes, we can do this eventually, but I'd merge this PR anyway because it alrady improves the situation a lot.

PR #965, opened 2020-07-23
Fix cycle_layout for omitted layout names

t-wissmann:
If a custom list of layouts is specified for cycle_layout and the
current layout is not part of this list, then cycle_layout has to pick
the first layout in the list according to the docs. The present commit
re-establishes this behaviour and adds test cases for it.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,7 +14,8 @@ Current git version
     ('floatplacement=none')
   * New rule condition 'pgid'
   * Bug fixes:
-    - Fix wrong behaviour in 'cycle_layout'
+    - Fix wrong behaviour in 'cycle_layout' in the case where the current layout
+      is not contained in the layout list passed to 'cycle_layout'.
 
 Release 0.8.3 on 2020-06-06
 ---------------------------
--- a/NEWS
+++ b/NEWS
@@ -13,6 +13,8 @@ Current git version
     clients ('floatplacement=center') or leaves the floating position as is
     ('floatplacement=none')
   * New rule condition 'pgid'
+  * Bug fixes:
+    - Fix wrong behaviour in 'cycle_layout'
 
 Release 0.8.3 on 2020-06-06
 ---------------------------
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -726,7 +726,12 @@ int FrameTree::cycleLayoutCommand(Input input, Output output) {
         string curname = Converter<LayoutAlgorithm>::str(cur_frame->getLayout());
         size_t count = input.end() - input.begin();
         auto curposition = std::find(input.begin(), input.end(), curname);
-        size_t idx = MOD((curposition - input.begin()) + delta, count);
+        size_t idx = 0;  // take the from the list per default
+        if (curposition != input.end()) {
+            // if the current layout name is in the list, take the next one
+            // (respective to the delta)
+            idx = MOD((curposition - input.begin()) + delta, count);
+        }
         try {
             layout_index = (int)Converter<LayoutAlgorithm>::parse(*(input.begin() + idx));
         } catch (const std::exception& e) {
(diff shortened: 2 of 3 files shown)


src/frametree.cpp
```diff
@@ -726,7 +726,12 @@ int FrameTree::cycleLayoutCommand(Input input, Output output) {
         string curname = Converter<LayoutAlgorithm>::str(cur_frame->getLayout());
         size_t count = input.end() - input.begin();
         auto curposition = std::find(input.begin(), input.end(), curname);
-        size_t idx = MOD((curposition - input.begin()) + delta, count);
+        size_t idx = 0;  // take the from the list per default
```
bew, line 729: Missing word in the comment?

t-wissmann: Right, thanks! I've fixed it.

tests/test_layout.py
```diff
@@ -813,3 +813,40 @@ def test_resize_clamp_argument_bigger(hlwm):
...
+
+
+@pytest.mark.parametrize("delta", ['-1', '+1'])
+def test_cycle_layout_name_not_in_list(hlwm, delta):
+    hlwm.call('set_layout vertical')
+
+    # if the current name is not contained in the custom list
+    # of layouts, then the first entry in the custom list
+    # has to be picked
+    hlwm.call('cycle_layout {} max grid'.format(delta))
```
blarz, line 831: Use f-string here?

tests/test_layout.py
```diff
@@ -813,3 +813,40 @@ def test_resize_clamp_argument_bigger(hlwm):
...
+    assert current_layout_name(hlwm) == 'max'
+
+
+@pytest.mark.parametrize("delta", ['-1', '+1'])
+def test_cycle_layout_name_in_the_list(hlwm, delta):
+    hlwm.call('set_layout vertical')
+    layouts = ['vertical', 'max', 'grid']
+    # here, we omit 'horizontal' to verify that we never reach
+    # the 'horizontal' layout
+    command = ['cycle_layout', delta] + layouts
```
blarz, line 842: Call it cycle_command maybe for more clarity

tests/test_layout.py
```diff
@@ -813,3 +813,40 @@ def test_resize_clamp_argument_bigger(hlwm):
...
+    hlwm.call('cycle_layout {} max grid'.format(delta))
+
+    assert current_layout_name(hlwm) == 'max'
+
+
+@pytest.mark.parametrize("delta", ['-1', '+1'])
+def test_cycle_layout_name_in_the_list(hlwm, delta):
+    hlwm.call('set_layout vertical')
+    layouts = ['vertical', 'max', 'grid']
+    # here, we omit 'horizontal' to verify that we never reach
```
blarz, line 840: This comment confused me. How do verify this in this test?

NEWS
```diff
@@ -13,6 +13,8 @@ Current git version
     clients ('floatplacement=center') or leaves the floating position as is
     ('floatplacement=none')
   * New rule condition 'pgid'
+  * Bug fixes:
+    - Fix wrong behaviour in 'cycle_layout'
```
blarz, line 17: Could me a bit more verbose (without reading the corresponding commit message I wouldn't know what happened here).

tests/test_layout.py
t-wissmann: If you prefer.

tests/test_layout.py
t-wissmann: I called it `cycle_layout_cmd`.

tests/test_layout.py
t-wissmann: I clarified the comment. 'horizontal' is cleary not contained in the 'expected_layouts' list, so we implicitly verify that we never reach the 'horizontal' layout.

NEWS
t-wissmann: I don't know how it helps, since the bug is fixed now. But I clarified it a bit.

tests/test_layout.py

tests/test_layout.py

tests/test_layout.py
blarz: Ah I see, thanks for clarifying!

NEWS
blarz: Thank you!

PR #969, opened 2020-07-26
herbstclient: Quiet more errors in case of --quiet

blarz:
Some error messages regarding display or herstluftwm availability would still be printed if -q was given.

### DIFF
--- a/ipc-client/main.c
+++ b/ipc-client/main.c
@@ -112,7 +112,9 @@ int main_hook(int argc, char* argv[]) {
     }
     HCConnection* con = hc_connect_to_display(display);
     if (!hc_check_running(con)) {
-        fprintf(stderr, "Error: herbstluftwm is not running\n");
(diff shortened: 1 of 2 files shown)


blarz:
Just to get this straight, the man page says -q disables messages regarding hlwm availability, right?
So other error messages, like "Could not send command" should still be printed, regardless of -q, right?

t-wissmann:
> Just to get this straight, the man page says -q disables messages regarding hlwm availability, right?

The "Could not send command" error can only arise if hlwm suddenly becomes inavailable, so `-q` can hide it, too.

blarz:
> > Just to get this straight, the man page says -q disables messages regarding hlwm availability, right?
> 
> The "Could not send command" error can only arise if hlwm suddenly becomes inavailable, so `-q` can hide it, too.

Added that, thanks! Testing those paths could get a bit hard..

PR #971, opened 2020-07-26
Handle changes of the tag attr 'floating_focused'

### LINKED ISSUE #947: tags.*.floating_focused shouldn’t be writeable
…because setting it is ignored (just messes up the attribute).

Or, what was probably the original idea, respect setting it.

t-wissmann:
When changing the attribute 'floating_focused' of a tag the tag needs a
relayout. This commit connects the signals accordingly. We don't want the
unnecessary relayout when the change of 'floating_focused' comes from
hlwm itself, e.g. within commands like 'jumpto' or 'cycle_all'. Hence,
this commit introduces a new signal on attributes that is only fired
when the user triggers a change, more specifically, when
Attribute::change() is called. The new testcase verifies this commit
(i.e. the testcase failed before this commit).

This fixes #947.

While implementing this, I noticed a bug in the Signal-types and the
overloading of Signal::connect(). The non-working call of connect() is
added as a comment, but fixing this issue is beyond the scope of this
commit.


blarz:
Two typos in the commit message:

> unecessary relayout

unnecessary

> when the use triggers a change, more specifically, when

the user?

t-wissmann:
> Two typos in the commit message:

Fixed.Thanks!

tests/test_tags.py
```diff
@@ -206,6 +206,22 @@ def test_close_or_remove_client(hlwm):
...
+def test_floating_focused_change(hlwm):
+    # create a floating client which is not focused
+    # because the tiling layer is focused
+    hlwm.call('rule floating=on focus=off')
+    floating_winid, _ = hlwm.create_client()
+    assert hlwm.get_attr('tags.focus.floating_focused') == hlwm.bool(False)
+    assert 'focus' not in hlwm.list_children('clients')
+
+    # switch to floating layer
+    hlwm.call('attr tags.focus.floating_focused on')
```
blarz, line 218: Extend the test to see what happenes when we set it to "off" again maybe? Would feel more complete.
Also: What happens if we have two floating clients? How gets focus?

t-wissmann: I'll add the test cases later.

t-wissmann: I've added a testcase and I'm  also disallowing the floating layer to be focused if there are no floating windows at all.

blarz: Awesome thank you for extending the tests!
Denying the floating layer focusing broke another test case though

t-wissmann: I've fixed it (the test case was the automatic generation and execution of commands via the completions)

blarz: Thanks!

PR #979, opened 2020-07-30
Add initial argument to new_attr command

t-wissmann:
Equip the new_attr command with an optional argument that initializes
the value of the fresh attribute. While at it, I removed the square
brackets around the type argument, because it looks as if the type
parameter was optional (the type parameter is mandatory). Since it makes
the code easier, I'm also switching to the new ArgParse syntax.

In the man page, I'm not using the term 'initial value', because it is
not the initial value but the second value of the attribute (this can be
relevant in future hlwm versions as soon as it is possible for the user
to hook into the change events of attributes).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,7 @@ Current git version
     clients ('floatplacement=center') or leaves the floating position as is
     ('floatplacement=none')
   * New rule condition 'pgid'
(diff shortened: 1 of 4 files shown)


t-wissmann:
Thanks!

Issue #906, opened 2020-05-08
Splitting frames that are not leaves of the binary tree 


elvisjauld:
Hello!
I'm not sure if this is the right place to ask, but I've read the man and I cannot figure out a way to easily change the layout of frames so that a non-leaf frame is split.
Basically I would like to change the layout in the following way:
```                     
 A | B     ->       _A_|_B_
                       C
```
by splitting the root frame (or any other frame that is not a leaf).
Is there a way to reorganize the tree like this in herbstluftwm without relying on scripts ?

I've been following this project for a little while. I really like herbstluftwm, but never got to stick with it. The recent and exciting developments,  as well as all the friendly and open discussions I see in the issue section, have encouraged me to switch from bspwm. Thank you for this!

t-wissmann:
Hi,

> Basically I would like to change the layout in the following way:
> 
> ```
>  A | B     ->       _A_|_B_
>                        C
> ```

So you'd like to what the following shell script does?

```
herbstclient load "(split vertical:0.5:1 $(herbstclient dump) (clients horizontal:0))"
```

You're right, that this is not possible directly in hlwm. The main reason for this is that I didn't come up with a good way of specifying a user interface this split. One can think of having a command like: `split_at_root horizontal|vertical` but I have no idea how to specify a clean/easy interface for splitting frames somewhere within the tree.

The only way I can think of it at the moment is to add another parameter to the `split` command that contains the frame index of the frame that should be split (and the defualt argument is `@`, refering to the focused frame(-leaf)).

What do you think?


> I've been following this project for a little while. I really like herbstluftwm, but never got to stick with it. The recent and exciting developments, as well as all the friendly and open discussions I see in the issue section, have encouraged me to switch from bspwm. Thank you for this!

Thanks and welcome on board!

bew:
Maybe it would help to be able to focus a non-leaf frame?
The UI would then be pretty straight forward, maybe with a different focus color so we can differentiate leaf vs non-leaf frames.

> add another parameter to the `split` command that contains the frame index of the frame that should be split (with default to `@`)

This looks good too IMO!

elvisjauld:
Hi, thanks for the response !
> So you'd like to what the following shell script does?

Yes, I wrote  a little script to split the root frame in a given direction, but was wondering if there was a native way to do this. 

[...]

PR #980, opened 2020-07-30
Allow to split non-leaf frames

t-wissmann:
This extends the 'split' command by an optional argument containing a
frame index. The FrameTree::splitCommand() was quite specific to
FrameLeafs, so I had to change the type of 'frame' and tried to keep the
meaning of the other variables as good as possible for the case where
'frame' is a FrameSplit.

The completion of the 'split' command is still defined in command.cpp.
Migrating this is in the scope of another command.

This implements #906.


t-wissmann:
This PR will have file conflits with #974 because both mark the `FrameTree::replaceNode` function as public.

NEWS
```diff
@@ -9,6 +9,7 @@ Current git version
   * Fix handling of delta -1 in 'focus_monitor' and 'cycle_monitor'
   * Fixed precision decimals in the layout tree (more reliable in- and output
     of fractions in frame splits)
+  * New optional argument 'split' for splitting non-leaf frames
```
blarz, line 12: "New optional argument for the 'split' command" maybe?

doc/herbstluftwm.txt
```diff
@@ -398,6 +399,7 @@ split 'ALIGN' ['FRACTION']::
         * split horiz 0.3
         * split vertical 0.5
         * split h
+        * split top 0.2 ''   (splits the root frame)
```
blarz, line 402: So an empty frame index means root frame? Is this "known"? ;)

NEWS
t-wissmann: Yes; fixed!

doc/herbstluftwm.txt
t-wissmann: Yes, it's documented in the section "frame index" (I've added a reference)

blarz: Awesome, thank you!

NEWS
blarz: Thanks!

Issue #944, opened 2020-07-02
When calling stoi() we should always check that all characters are read


t-wissmann:
`stoi(s)` does not throw an exception if the string `s` ends with some non-parseable junk. I think we should always verify that `s` is parsed in it's entirity.
```
$ herbstclient new_attr int my_int
$ herbstclient attr my_int 1337suffix
$ herbstclient attr my_int           
1337
```
Concretely, we should pass a `size_t* pos` parameter to `stoi()` and verify `s.size() == pos` afterwards.

PR #982, opened 2020-08-02
Fix crash while parsing empty decimal

blarz:
An empty decimal would result in a crash
while trying to dereference parts[0] in fixdecprec.cpp.


blarz:
May be related to #944?

blarz:
dmesg says:

```herbstluftwm[67772]: segfault at 8 ip 00007f37dc0b0224 sp 00007ffc47a71228 error 4 in libstdc++.so.6.0.28[7f37dc003000+f0000]```

blarz:
Tests are failing, but not related to this commit as far as I can tell.
The re-run button does not seem to retrigger the check for me..

PR #983, opened 2020-08-02
Reorder some NEWS entries

blarz:
Some entries were actually bug fixes and one fix is rather a feature

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,9 +6,6 @@ Current git version
 
   * When an application fails to focus itself (because
     focus_stealing_prevention is active), then the window is marked as urgent.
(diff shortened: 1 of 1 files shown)


t-wissmann:
For the fixed-precission and the reaction to the attribute change I'd say that it's a highly philosophical discussion whether one should regard these as 'new features' or 'bug fixes'...

blarz:
> For the fixed-precission and the reaction to the attribute change I'd say that it's a highly philosophical discussion whether one should regard these as 'new features' or 'bug fixes'...

I was thinking about the "it's a bug not a feature" thing as well.
But the actual wording of the entries gave me an urge to move them into the other section,
probably because they started/didn't start with "Fix" ;)

Thanks for approving!

PR #985, opened 2020-08-07
Add 'index' attribute to frame objects

t-wissmann:
This equips the frame objects with a dynamic attribute 'index'
containing the frame's frame index. Since there is no documentation for
frame objects yet, there is no documentation for the 'index' attribute
here.

This feature has been requested by @bew during the discussion of #906.

### DIFF
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -970,5 +970,5 @@ def test_frame_index_attribute(hlwm):
         hlwm.call(['split', 'auto', '0.5', index])
 
         # after each splitting operation, check that
-        # the frame_index attribute is correct:
+        # the frame's index attribute is correct:
         verify_frame_tree('tags.focus.tiling.root', '')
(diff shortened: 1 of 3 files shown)


tests/test_layout.py
```diff
@@ -952,3 +952,23 @@ def test_split_something_in_between(hlwm, splitmode, context):
...
+        assert hlwm.get_attr(f'{object_path}.index') == index_prefix
+        if '0' in hlwm.list_children(object_path):
+            verify_frame_tree(f'{object_path}.0', f'{index_prefix}0')
+            verify_frame_tree(f'{object_path}.1', f'{index_prefix}1')
+
+    for index in split_index:
+        hlwm.call(['split', 'auto', '0.5', index])
+
+        # after each splitting operation, check that
+        # the frame_index attribute is correct:
```
blarz, line 973: It is just "index" not frame_index, right?

blarz: Is frame_index maybe a better name for the actual attribute? Not sure, seems redundant

t-wissmann: I've changed the comment to "the frame's index" attribute. I think to avoid redundancy, we should keep the name `index`.

blarz: Cool, thanks!

PR #987, opened 2020-08-08
update

Piping:


t-wissmann:
Hm? What is this PR about? It seems like a combination of multiple things and I don't see what the purpose of the changes are. Did you submit this PR by accident?

Piping:
@t-wissmann oops. It is by accident. It was meant to be merged into my own fork right now.

Piping:
@t-wissmann One question I have is how `vector<string> cmds` looks like ? I tried to use `mm->mouse_call_command` to invoke my command, but it does not work. For example, i tried `spwan xterm`, `xterm`, neither of them works.

src/xmainloop.cpp
```diff
@@ -216,6 +216,10 @@ void XMainLoop::buttonpress(XButtonEvent* be) {
                     mm->mouse_initiate_move(client, {});
                 }
             }
+        } else {
+            // root windows handling
+            HSDebug("handle default RootWindow ButtonPress on win 0x%lx\n", be->window);
+            mm->mouse_call_command_root_window({"call", "spwan","xmenu.sh"});
```
t-wissmann, line 222: There's a typo `spawn` vs `spwan`. Also you're making a unencessary round-trip through the `mm` module. you can do the `callCommand` directly here:
```
std::ostringstream discardedOutput;
Input input("spawn", {"xmenu.sh"});
Command::call(input, discardedOutput);
```

t-wissmann:
What does your xmenu.sh look like? Does it launch an Openbox/Fluxbox style menu on the root window?

Piping:
image

Glad it works now on the build ( it might look ugly due to missing Xresource file in xephyr). Thanks for the correction. for the xmenu details, you can check out my script at 

`https://github.com/instantOS/instantUBUNTU/blob/dev/core/xmenu.sh`

PR #988, opened 2020-08-11
Fix crash in smart placement on invisible tags

t-wissmann:
If multiple floating clients are placed smartly on a tag, then the
'last_scheme' member of Decoration is used to compute the correct
geometry of a client. If this involves clients that have never been
shown, then the 'last_scheme' is not yet initialized.

This commit checks the 'last_scheme' pointer explicitly in
Decoration::inner_to_outer() and adds some test cases which reproduced
the crash earlier. Moreover, I've also added the check in
Decoration::redrawPixmap(), just to be on the safe side.

There is no entry in the NEWS section for this, because this bug was not
present in the last released version.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -139,6 +139,10 @@ void Decoration::resize_inner(Rectangle inner, const DecorationScheme& scheme) {
 }
 
 Rectangle Decoration::inner_to_outer(Rectangle rect) {
+    if (!last_scheme) {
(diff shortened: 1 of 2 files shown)


t-wissmann:
I've merged this directly because it's fixing a crash but doesn't do anything else.

blarz:
> I've merged this directly because it's fixing a crash but doesn't do anything else.

Sure thing! Was looking at it this morning but got side tracked before approving. Good find!

PR #990, opened 2020-08-11
Add test list_rules printing 'not'

t-wissmann:
This adds a test case where list_rules prints a rule involving negation.


blarz:
Fixed a typo in the PR title

Issue #26, opened 2014-10-09
objecttree: foreach


The-Compiler:
Some people wanted a `hc foreach tags ...` to do some action for each child of an object in the tree. This pushes one step closer to a turing-complete hlwm but still should not cause any possible hangs.

t-wissmann:
Implemented by #993.

PR #993, opened 2020-08-17
Add the 'foreach' command for object iteration

t-wissmann:
The foreach command iterates over all children of a given object. With
this, it becomes easy to iterate over all clients, tags, monitors. It's
probably only a first version of the 'foreach' because often one needs
to filter the children in a certain way (e.g. 'tags' has more children
than tags because there is something like tags.focus and tags.by-name).

This implements #26.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -196,6 +196,8 @@ unique_ptr<CommandTable> commands(shared_ptr<Root> root) {
                                             &RootCommands::print_object_tree_complete} },
         {"substitute",     { root_commands, &RootCommands::substitute_cmd,
                                             &RootCommands::substitute_complete} },
+        {"foreach",        { root_commands, &RootCommands::foreachCmd,
+                                            &RootCommands::foreachComplete} },
         {"sprintf",        { root_commands, &RootCommands::sprintf_cmd,
                                             &RootCommands::sprintf_complete} },
         {"new_attr",       { root_commands, &RootCommands::new_attr_cmd,
--- a/src/rootcommands.cpp
+++ b/src/rootcommands.cpp
@@ -180,6 +180,56 @@ void RootCommands::substitute_complete(Completion& complete)
     }
 }
 
+int RootCommands::foreachCmd(Input input, Output output)
+{
+    string ident, pathString;
+    if (!(input >> ident >> pathString)) {
+        return HERBST_NEED_MORE_ARGS;
+    }
+    // remove trailing dots to avoid parsing issues:
+    while (!pathString.empty() && pathString.back() == OBJECT_PATH_SEPARATOR) {
+        // while the last character is a dot, erase it:
+        pathString.erase(pathString.size() - 1);
+    }
+    Path path { pathString, OBJECT_PATH_SEPARATOR };
+    Object* object = root.child(path, output);
+    if (!object) {
+        return HERBST_INVALID_ARGUMENT;
+    }
+
+    // collect the paths of all children of this object
+    vector<string> childPaths;
+    // collect the children's name first to ensure that
+    // object->childrens() is not changed by the commands we are
+    // calling.
+    if (!pathString.empty()) {
+        pathString += OBJECT_PATH_SEPARATOR;
+    }
+    for (const auto& entry : object->children()) {
+        childPaths.push_back(pathString + entry.first);
+    }
+    int  lastStatusCode = 0;
+    for (const auto& child : childPaths) {
+        Input carryover = input.fromHere();
+        carryover.replace(ident, child);
+        lastStatusCode = Commands::call(carryover, output);
+    }
+    return lastStatusCode;
+}
+
+void RootCommands::foreachComplete(Completion& complete)
+{
+    if (complete == 0) {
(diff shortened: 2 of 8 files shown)


src/main.cpp
```diff
@@ -196,6 +196,8 @@ unique_ptr<CommandTable> commands(shared_ptr<Root> root) {
                                             &RootCommands::print_object_tree_complete} },
         {"substitute",     { root_commands, &RootCommands::substitute_cmd,
                                             &RootCommands::substitute_complete} },
+        {"foreach",        { root_commands, &RootCommands::foreachCmd,
```
blarz, line 199: Not related to this PR, but where does the mixture of camel and snake case in the RootCommands come from?

src/rootcommands.cpp
```diff
@@ -180,6 +180,56 @@ void RootCommands::substitute_complete(Completion& complete)
...
+    Path path { pathString, OBJECT_PATH_SEPARATOR };
+    Object* object = root.child(path, output);
+    if (!object) {
+        return HERBST_INVALID_ARGUMENT;
+    }
+
+    // collect the paths of all children of this object
+    vector<string> childPaths;
+    // collect the children's name first to ensure that
+    // object->childrens() is not changed by the commands we are
```
blarz, line 203: children()?

src/types.cpp
```diff
@@ -13,11 +13,11 @@ Input &Input::operator>>(string &val)
 Input Input::fromHere()
 {
     string cmd;
-    if (!(*this >> cmd)) {
+    if (empty()) {
         return {{}, {}};
     }
 
-    return Input(cmd, toVector());
+    return Input(*(begin()), Container(begin() + 1, end()));
```
blarz, line 20: It is not really clear to me how does this change corresponds to this PR?

src/rootcommands.cpp
```diff
@@ -180,6 +180,56 @@ void RootCommands::substitute_complete(Completion& complete)
...
+    }
+    Path path { pathString, OBJECT_PATH_SEPARATOR };
+    Object* object = root.child(path, output);
+    if (!object) {
+        return HERBST_INVALID_ARGUMENT;
+    }
+
+    // collect the paths of all children of this object
+    vector<string> childPaths;
+    // collect the children's name first to ensure that
```
blarz, line 202: child's?

blarz (changes requested):
Very nice PR, great test coverage, couldn't really think of another case to break it.
Just a minor non-related question and two typos (probably)

src/rootcommands.cpp
```diff
@@ -180,6 +180,56 @@ void RootCommands::substitute_complete(Completion& complete)
...
+    int  lastStatusCode = 0;
+    for (const auto& child : childPaths) {
+        Input carryover = input.fromHere();
+        carryover.replace(ident, child);
+        lastStatusCode = Commands::call(carryover, output);
+    }
+    return lastStatusCode;
+}
+
+void RootCommands::foreachComplete(Completion& complete)
```
blarz, line 220: Coverage diff shows, that we only test the complete == 0 case. Can we add tests for the other if branches?

src/main.cpp
t-wissmann: I think we agreed on "strict" camelcase after some of the root commands already had been implemented. ("strict" meaning that camelcase also applies to these "Cmd" and "Complete" suffixes)

src/rootcommands.cpp
t-wissmann: Yes, thanks!

src/types.cpp
t-wissmann: This just fixes the behaviour of `Input::fromHere()` such that it can be called multiple times. Before, calling `fromHere()` consumed the command name but not the args, now it does not consume anything. We didn't notice this issue before, because before this PR, we never used the Input after calling `fromHere()`.

src/rootcommands.cpp
t-wissmann: It should be "children's names".

src/rootcommands.cpp
t-wissmann: I've added a few basic test cases for the other branches. There's not much to test here because it only calls the existing completion routines.

src/types.cpp
blarz: I see, thanks!

src/main.cpp
blarz: Ok, just wondered if there was some kind of system behind this

src/rootcommands.cpp
blarz: Ok, normally I am not the "we need a test for coverage" person, but seeing that you fixed a bug in the sprintf %c PR regarding completion I thought it couldn't hurt

blarz (approved):
Very nice!

src/rootcommands.cpp
t-wissmann: Right, fair enough. Tough in the sprintf case, the test case even tested for the wrong behaviour (and I actually noticed the wrong test case).

PR #994, opened 2020-08-17
New constant string format specifier for sprintf

t-wissmann:
This introduces a new format specifier '%c' in the sprintf format
strings. Of course it never makes sense to manually write something like

    hc sprintf S '%c.name' 'tags.0'

but it makes sense in combination with the other commands that replace
certain identifiers in commands, especially in combination with the foreach
command.

This change required an update in the completion code for the 'sprintf'
command, and while at it, I fixed a minor bug in the completion. The bug
was that completing the command

   hc sprintf S '%s'

erroneously yielded S.

### DIFF
--- a/src/rootcommands.cpp
+++ b/src/rootcommands.cpp
@@ -239,7 +239,7 @@ int RootCommands::sprintf_cmd(Input input, Output output)
     for (const auto& blob : format) {
         if (blob.literal_) {
             replacedString += blob.data_;
-        } else if (blob.data_ == "c"){
+        } else if (blob.data_ == "c") {
             // a constant string argument
             string constString;
             if (!(input >> constString)) {
@@ -279,7 +279,7 @@ void RootCommands::sprintf_complete(Completion& complete)
             complete.invalidArguments();
             return;
         }
-        int	indexOfNextArgument = 2;
+        int indexOfNextArgument = 2;
         for (const auto& b : fs) {
             if (b.literal_ == true) {
                 continue;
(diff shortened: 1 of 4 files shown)


blarz:
Fixed some typos in the PR description.

src/rootcommands.cpp
```diff
@@ -237,6 +239,14 @@ int RootCommands::sprintf_cmd(Input input, Output output)
     for (const auto& blob : format) {
         if (blob.literal_) {
             replacedString += blob.data_;
+        } else if (blob.data_ == "c"){
```
blarz, line 242: Style: Add a space between ) and {

src/rootcommands.cpp
```diff
@@ -262,22 +272,30 @@ void RootCommands::sprintf_complete(Completion& complete)
...
+            fs = parseFormatString(complete[1]);
         }  catch (const std::invalid_argument&) {
             complete.invalidArguments();
+            return;
         }
-        if (complete < 2 + placeholderCount) {
-            completeAttributePath(complete);
-        } else {
-            complete.completeCommands(2 + placeholderCount);
+        int	indexOfNextArgument = 2;
```
blarz, line 282: Double whitespace

blarz (changes requested):

t-wissmann:
I've fixed both whitespace issues (the 'double' whitespace was a tab; I don't know how qtcreator was able to sneak that in)

blarz (dismissed):
Nice! Looking forward to see what the users will do with this :D

t-wissmann:
I've pushed a new commit resolving the merge conflicts in `doc/herbstluftwm.txt` that arised from the merge of the foreach-PR.

As soon as this is merged, I can answer a few open issues/feature requests :-) (closing all clients is one of them)

blarz (approved):

PR #996, opened 2020-08-19
Fix manpage for foreach command

The-Compiler:
Follow-up to #993


blarz (approved):
Thank you! Totally overlooked that in the review

t-wissmann:
Thanks!

PR #997, opened 2020-08-23
Remove one DynAttribute_ constructor

t-wissmann:
Using the 'Owner'-aware constructors of DynAttribute_ improve the
readability. The particular constructor removed by this commit was used
in only one place anyway.

This commit does not require new test cases, because there are a couple
of test cases for the attribute 'tags.count'.

### DIFF
--- a/src/attribute_.h
+++ b/src/attribute_.h
@@ -210,12 +210,6 @@ class DynAttribute_ : public Attribute {
 public:
     // each time a dynamic attribute is read, the getter_ is called in order to
     // get the actual value
-    DynAttribute_(const std::string &name, std::function<T()> getter)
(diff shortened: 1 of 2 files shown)


t-wissmann:
This fails on 32 bit:
```
/hlwm/src/indexingobject.h:18:52: error: no matching function for call to 'DynAttribute_<long unsigned int>::DynAttribute_(IndexingObject<HSTag>* const, const char [6], size_t (IndexingObject<HSTag>::*)())'
     : count(this, "count", &IndexingObject<T>::size)
```
So I'll add another `sizeUnsignedLong()` member function to `IndexingObject` to avoid the implicit cast.

PR #998, opened 2020-08-25
Generate json documentation on the object tree

t-wissmann:
This adds a new script doc/gendoc.py which extracts as much information about
herbstluftwm's object tree as possible. So far, this includes the objects,
their attributes, and possible children. The output format of doc/gendoc.py is
json, so we can pass this information around easily and in fact can verify the
json-formatted documentation by the test suite (more specifically by
tests/test_doc.py).

Motivation
----------
There are three reasons for me starting this:

  - Much of the documentation plainly copies information that is already
    in the source code
  - In the long run, I'd like to have live documentation (something like
    a 'help' command as in bash), so in the end the documentation
    (including explanations) should be easily accessible for such a
    'help' command.
  - I'd like to move the object tree documentation out of
    doc/herbstluftwm.txt, because this file has already 1700 lines.

Caveats / Alternatives
----------------------
The doc/gendoc.py script will easily become (or already is?)
unmaintainable. Writing a tokenizer and half a parser from scratch feels
like reinventing the wheel, but I couldn't find a better way to do it.
I've failed with the following alternative approaches:

   - doxygen seems to be the canonical solution for documentation,
     however, it does not seem to parse anything in actual code blocks
     (e.g. the attribute names are only present in the arguments passed
     to the constructor of Attribute_).
   - clang/llvm even has a python api and I have a kind of skeleton to
     parse the hlwm source code, but it seems that sometimes it gets
     confused by template arguments.

I understand that the doc/gendoc.py script itself is way too long to
review, but fortunately, we have the much shorter test suite and one
easily gets a feeling what is going on by simply looking at the
doc/hlwm-doc.json file that is generated by 'make'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,6 @@ doc: $(BUILDDIR)
 tar: doc
 	git archive --prefix=herbstluftwm-$(VERSION)/ -o $(TARFILE) HEAD
 	tar --transform="flags=r;s,$(BUILDDIR),herbstluftwm-$(VERSION),"  --owner=0 --group=0 \
-		-uvf $(TARFILE) $(BUILDDIR)/doc/*.html $(BUILDDIR)/doc/*.[1-9]
+		-uvf $(TARFILE) $(BUILDDIR)/doc/*.{html,json} $(BUILDDIR)/doc/*.[1-9]
 	gzip $(TARFILE)
 	gpg --detach-sign $(TARFILE).gz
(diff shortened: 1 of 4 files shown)


blarz:
Fixed some typos in the comments and in the commit message.

One sentence confuses me in the commit message:
> should be in some parsable format and/or
directly by present in the source code.

You might want to rephrase that.

Other than that, I guess we can put it in, if it does what you need.
You'll be the one maintaining it :D

t-wissmann:
Thanks, I've rephrased it :)

blarz (approved):
Let's do this!

PR #1002, opened 2020-09-11
Add -h and --help to herbstluftwm

t-wissmann:
This is long overdue, but herbstluftwm has been lacking the -h and
--help command line flags. On -h and --help, herbstluftwm now prints a
small explanation what it is, prints all command line options and
mentions 'herbstclient' and the man page. I think this is enough for the
time being, because one never interacts manually with the herbstluftwm
binary (only when calling it mistakenly instead of 'herbstclient'
maybe).

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -451,7 +451,7 @@ static void parse_arguments(int argc, char** argv, Globals& g) {
                 break;
             case 'h':
                 std::cout << "This starts the herbstluftwm window manager. In order to" << endl;
-                std::cout << "interact with a running herbstluftwm instance use the" << endl;
+                std::cout << "interact with a running herbstluftwm instance, use the" << endl;
                 std::cout << "\'herbstclient\' command." << endl;
                 std::cout << endl;
                 std::cout << "The herbstluftwm command accepts the following options:" << endl;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -432,7 +432,7 @@ static void parse_arguments(int argc, char** argv, Globals& g) {
     // parse options
     while (true) {
         int option_index = 0;
-        int c = getopt_long(argc, argv, "+c:vh", long_options, &option_index);
+        int c = getopt_long(argc, argv, "+c:vlh", long_options, &option_index);
         if (c == -1) {
             break;
         }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -463,7 +463,7 @@ static void parse_arguments(int argc, char** argv, Globals& g) {
                         if (long_options[i].has_arg) {
                             std::cout << " ARG";
                         }
-                        std::cout << "  ";
+                        std::cout << ", ";
                     }
                     std::cout << "--" << long_options[i].name;
                     if (long_options[i].has_arg) {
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -448,6 +449,31 @@ static void parse_arguments(int argc, char** argv, Globals& g) {
             case 'l':
                 g.initial_monitors_locked = 1;
                 break;
+            case 'h':
+                std::cout << "This starts the herbstluftwm window manager. In order to" << endl;
+                std::cout << "interact with a running herbstluftwm instance use the" << endl;
```
dnnr, line 454: I think there should be a comma after `instance`

src/main.cpp
```diff
@@ -431,7 +432,7 @@ static void parse_arguments(int argc, char** argv, Globals& g) {
     // parse options
     while (true) {
         int option_index = 0;
-        int c = getopt_long(argc, argv, "+c:vl", long_options, &option_index);
+        int c = getopt_long(argc, argv, "+c:vh", long_options, &option_index);
```
dnnr, line 435: I'm not totally sure about this syntax, but shouldn't the `l` have stayed here?

t-wissmann: Oh right. I've accidentally deleted it...

src/main.cpp
```diff
@@ -448,6 +449,31 @@ static void parse_arguments(int argc, char** argv, Globals& g) {
...
+                std::cout << "The herbstluftwm command accepts the following options:" << endl;
+                std::cout << endl;
+                for (size_t i = 0; long_options[i].name; i++) {
+                    std::cout << "    ";
+                    if (long_options[i].val != 1) {
+                        std::cout << "-" << static_cast<char>(long_options[i].val);
+                        if (long_options[i].has_arg) {
+                            std::cout << " ARG";
+                        }
+                        std::cout << "  ";
```
dnnr, line 466: I think `", "` would be a more idiomatic way for listing the short and long form of arguments

dnnr: Besides that, I even tried the PR locally since there's no test (not appropriate, I guess...), and I looks good to me. :+1:

t-wissmann: I didn't know what to test (beside that running it does not crash). If you have an idea, I can also add a testcase.
I've replaced the separator by `", "`. (I didn't know which command to check for comparison; but `cp` indeed uses `", "` and it looks good):

```
$ ./herbstluftwm -h
This starts the herbstluftwm window manager. In order to
interact with a running herbstluftwm instance, use the
'herbstclient' command.

The herbstluftwm command accepts the following options:

    -v, --version
    -h, --help
    -c ARG, --autostart ARG
    -l, --locked
    --exit-on-xerror
    --no-tag-import
    --verbose

See the herbstluftwm(1) man page for their meaning.
```

dnnr: Imho, a proper test would just run `herbstluftwm -h` (and `--help`) and compare its output. But as I said, I don't think that's appropriate because it's not really a kind of test that fits very well into the test framework, and it's rather low value. That value is mainly that it would have allowed me to take a look at the expected output without actually running it manually.

dnnr (approved):

PR #1005, opened 2020-09-16
Add per-client keyboard layout script

nkrkv:
Hello and thank you for the HLWM!

What I was missing when switched from Gnome is the feature to keep the keyboard layout bound to windows individually. Thanks to Herbstluftwm flexibility, this can be recovered. I’ve been testing the script for a week or so on a daily basis, works fine, found no problems.

### DIFF
--- a/scripts/perclient_kb_layout.sh
+++ b/scripts/perclient_kb_layout.sh
@@ -37,8 +37,8 @@ FOCUS_WINID=$(hc attr clients.focus.winid)
 hc --idle focus_changed | while read hook winid name
 do
     # Save current keyboard layout for window loosing focus
-    hc silent new_attr int clients.${FOCUS_WINID}.my_kb_layout || true;
-    hc silent attr clients.${FOCUS_WINID}.my_kb_layout $(xkblayout-state print "%c");
+    hc try silent new_attr int clients.${FOCUS_WINID}.my_kb_layout;
+    hc silent attr clients.${FOCUS_WINID}.my_kb_layout "$(xkblayout-state print '%c')";
 
     # Save the currently focused win id to be able referring it as the one loosing focus
     FOCUS_WINID=$winid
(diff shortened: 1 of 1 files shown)


scripts/perclient_kb_layout.sh
```diff
@@ -0,0 +1,52 @@
...
+hc() {
+    herbstclient "$@"
+}
+
+FOCUS_WINID=$(hc attr clients.focus.winid)
+
+hc --idle focus_changed | while read hook winid name
+do
+    # Save current keyboard layout for window loosing focus
+    hc silent new_attr int clients.${FOCUS_WINID}.my_kb_layout || true;
```
t-wissmann, line 40: You can also do `hc try …` instead of `hc … || true`.

scripts/perclient_kb_layout.sh
```diff
@@ -0,0 +1,52 @@
...
+    hc silent attr clients.${FOCUS_WINID}.my_kb_layout $(xkblayout-state print "%c");
```
t-wissmann, line 41: Maybe it's safer to use quotes around the `$(...)` (implying the necessity of single quotes for the inner `'%c'`)

t-wissmann:
Thanks for the well-written and cleanly documented script! Independent from the functionality, I think it's good to ship examples of these `hc --idle` scripts together with hlwm :)

t-wissmann:
I've updated your PR. if it's also fine for you, I can merge this PR.

On a related note: in ibus (layout setter for more advanced language inputs), you can also activate a per-applicate layout (but I never tried it because it's already confusing enough to switch between layouts globally)

nkrkv:
Sure! I’ve, actually, never know about `hc try` (overlooked the docs), thank you for teaching :+1:

t-wissmann (approved):
looks good to me, thanks!

PR #1007, opened 2020-09-17
'compare': fall back to string-based comparison

t-wissmann:
If compare encounters an attribute of a certain type without an explicit
comparison function, then it should simply use the string
representation for the comparison.

An example is the type of layout algorithms and a test case is added for
that.


t-wissmann:
There is no formal dependency on #1007, but I think it makes sense to merge those new test cases before merging the present PR.

blarz:
Fixed a typo in the commit message

blarz (approved):
Looks good to me!

Not directly related to this PR: The "only allowed for numeric types" check could use a test, too, as far as I can tell from the coverage.

PR #1009, opened 2020-09-18
Improve try/silent handling in maximize.sh

The-Compiler:
- Remove 'try' for remove_attr
    
The underlying issue was fixed, most likely in 5664c64b67f2f44e2e828f6ba231eccab60c5324.

- Make new_attr silent
    
If we aborted above because the user switched frames/layouts, the my_unmaximized_layout attribute could still exist from the previous run.


t-wissmann (approved):
Thanks :-) You're absolutely right about the fixed exit code issue!

PR #1013, opened 2020-09-27
www: Abandon frame and poisonous colors

t-wissmann:
This updates the visual layout of the web page in order to make it look
more modern:

  - change poisonous colors
  - remove the unnecessary frame around the text content such that the
    text margin on the left and right more look like the margin around
    the text on a piece of paper.

In recent weeks, many spam mails reached the (moderation queue of the)
mailinglist critizing that our web page looks much older and less modern
than those of our competitors; and actually that's true. The present
change tries to fix this.

### DIFF
--- a/www/compose.py
+++ b/www/compose.py
@@ -55,7 +55,6 @@ print("""\
   <title>{title}</title>
  </head>
  <body>
-  <div id="frame">
    <div id="header">
     <div id="logoname">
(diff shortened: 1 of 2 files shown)


t-wissmann:
I've skipped the entire CI-build because the asciidoc isn't built by the CI.

PR #1015, opened 2020-10-25
Emit the "fullscreen" hook again

### LINKED ISSUE #845: fullscreen hook
According to the old code (and the man page), the `fullscreen` hook is fired whenever the fullscreen state of a client changes. There are two options:
* Implement it again, by attaching to the `changed()` signal of the fullscreen attribute of every client
* Keep it dropped because I'm not aware of any scenario where this is useful.

blarz:
Fixes  #845


t-wissmann (approved):
Nice patch and test :-)

Issue #999, opened 2020-09-04
tag_renamed hook does not show old name


The-Compiler:
The `tag_renamed` hook only has the new name as argument, not the old one. Thus, it's only really useful as as a notification to call `tag_status` again, but it'd be nice if it carried that kind of information.

Conveniently, it's also currently undocumented in the manpage, so I suppose we could just change it? From a quick GitHub search, nobody really seems to use its argument:

- [`activities.pl` by @ypnos](https://github.com/ypnos/hlwm/blob/61b6fccd6259257229fd891bbf67b949d1b8ce21/activities.pl#L64-L73) (which mentions the same issue)
- [Someone's `panel.sh`](https://github.com/razzfox/dotfiles/blob/5bf5649fa80281f0e1190ec2270eca15dc28c084/config/herbstluftwm/panel.sh) (the default uses `tag*` instead)
- [ggbar](https://github.com/lutobler/ggbar/blob/c51f2c828763a39c8d3f6b007475f551fbb5c330/src/main.rs#L290)

blarz:
Thanks for the report! Should be fixed with #1016

The-Compiler:
@blarz Thanks for the fix! :+1:

PR #1016, opened 2020-10-25
Add the old tag name to tag_renamed hook

blarz:
As stated correctly in #999, this hook would be more useful
if you'd know which tag was renamed.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1667,6 +1667,9 @@ tag_added 'TAG'::
 tag_removed 'TAG'::
     The tag named 'TAG' was removed.
 
+tag_renamed 'OLD' 'NEW::
+    The tag name changed from 'OLD' to 'NEW'.
+
 urgent [on|off] 'WINID'::
     The urgent state of client with given 'WINID' has been changed to [on|off].
 
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1667,7 +1667,7 @@ tag_added 'TAG'::
 tag_removed 'TAG'::
     The tag named 'TAG' was removed.
 
-tag_renamed 'OLD' 'NEW::
+tag_renamed 'OLD' 'NEW'::
     The tag name changed from 'OLD' to 'NEW'.
 
 urgent [on|off] 'WINID'::
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -263,3 +263,17 @@ def test_urgent_count(hlwm, x11):
 
     # since one of them gets focused, 4 urgent clients remain
     assert int(hlwm.get_attr('tags.focus.urgent_count')) == 4
+
+
+def test_rename_different_tags(hlwm, hc_idle):
+    hlwm.call('add foo_old')
+    hlwm.call('add bar_old')
+    hc_idle.hooks()  # clear hooks
+
+    hlwm.call('rename foo_old foo_new')
+    hlwm.call('rename bar_old bar_new')
+
+    assert hc_idle.hooks() == [
+        ['tag_renamed', 'foo_old', 'foo_new'],
+        ['tag_renamed', 'bar_old', 'bar_new']
+    ]
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -263,7 +263,7 @@ def test_urgent_count(hlwm, x11):
     assert int(hlwm.get_attr('tags.focus.urgent_count')) == 4
 
 
-def test_rename_different_tags(hlwm, hc_idle):
+def test_rename_multiple_tags(hlwm, hc_idle):
     hlwm.call('add foo_old')
     hlwm.call('add bar_old')
     hc_idle.hooks()  # clear hooks
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -92,7 +92,14 @@ def test_rename_tag(hlwm, hc_idle, rename_command):
     hlwm.call(rename_command + ['foobar'])
 
     assert hlwm.get_attr('tags.0.name') == 'foobar'
-    assert hc_idle.hooks() == [['tag_renamed', 'foobar']]
+    assert hc_idle.hooks() == [['tag_renamed', 'default', 'foobar']]
+
+    rename_once_more = rename_command.copy()
+    rename_once_more[1] = rename_once_more[1].replace('default', 'foobar')
+    hlwm.call(rename_once_more + ['baz'])
+
+    assert hlwm.get_attr('tags.0.name') == 'baz'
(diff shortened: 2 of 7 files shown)


blarz:
Adding a static variable to the `tag->name.changed()` lambda was the only way I could see, to carry the old name with me somehow. Maybe somebody else has a better idea?

The `Attribute_::changed` signal doesn't give us the old value, so I had to figure out another way to keep track of the old value.

blarz:
Also the `replace()` dance I do in the test feels a bit weird, maybe there is a more elegant solution there, too.

t-wissmann:
Thanks for fixing these open bugs :)

> Adding a static variable to the `tag->name.changed()` lambda was the only way I could see, to carry the old name with me somehow. Maybe somebody else has a better idea?

The `static` creates only one variable for this lambda (and not for one closure!) and so the `static` leads to wrong results when there is more than one tag. I've pushed a test case where I do the renaming:
```
foo_old -> foo_new
bar_old -> bar_new
```
but the second call creates the hook `tag_renamed foo_new bar_new`, because the static variable is set to `foo_new`.

I think the best way is to have a private member in the `Tag` class `oldName_` that is used here. By the way, in the "Validator"-callbacks, the value of the attribute is still the old one, but I think it's cleaner to run this hook emision in the `changed()` callback.

Another thing regarding compatibility: This PR changes the semantics from `tag_renamed new` to `tag_renamed old new`, so if someone has code that listens for it and uses the first parameter, then the code now gets the old instead of the new tag name. While I think it is fine to introduce this incompatibility since the order "old new" makes much more sense, we should mention this in the `MIGRATION` file.

t-wissmann:
> Also the `replace()` dance I do in the test feels a bit weird, maybe there is a more elegant solution there, too.

I've made the entries in the `RENAME_COMMANDS` array lambda functions, so one can call
```python
hc.call(rename_command('old', 'new'))
```
If you don't like it feel free to revert my commit 44c6306 (and it thus will disappear in the squashed commit for this PR)

doc/herbstluftwm.txt
```diff
@@ -1667,6 +1667,9 @@ tag_added 'TAG'::
 tag_removed 'TAG'::
     The tag named 'TAG' was removed.
 
+tag_renamed 'OLD' 'NEW::
```
tiosgz, line 1670: Missing `'`?

tests/test_tags.py
```diff
@@ -256,3 +261,17 @@ def test_urgent_count(hlwm, x11):
 
     # since one of them gets focused, 4 urgent clients remain
     assert int(hlwm.get_attr('tags.focus.urgent_count')) == 4
+
+
+def test_rename_different_tags(hlwm, hc_idle):
```
tiosgz, line 266: I think `test_rename_multiple_tags` is a better name.

doc/herbstluftwm.txt
blarz: Thanks!

tests/test_tags.py
blarz: True, changed that. Thanks!

blarz:
> If you don't like it feel free to revert my commit [44c6306](https://github.com/herbstluftwm/herbstluftwm/commit/44c6306569b3dfeaeca38e11313e5dbca59c73ef) (and it thus will disappear in the squashed commit for this PR)

Perfect, thank you!

blarz:
> Thanks for fixing these open bugs :)

Sure thing! The autumn air outside inspired me to do some hlwm hacking :smile: 

> > Adding a static variable to the `tag->name.changed()` lambda was the only way I could see, to carry the old name with me somehow. Maybe somebody else has a better idea?
> 
> The `static` creates only one variable for this lambda (and not for one closure!) and so the `static` leads to wrong results when there is more than one tag. I've pushed a test case where I do the renaming:
> 
> ```
> foo_old -> foo_new
> bar_old -> bar_new
> ```
> 
> but the second call creates the hook `tag_renamed foo_new bar_new`, because the static variable is set to `foo_new`.

I knew there was something I've overlooked, thank you for the new test!

> I think the best way is to have a private member in the `Tag` class `oldName_` that is used here. By the way, in the "Validator"-callbacks, the value of the attribute is still the old one, but I think it's cleaner to run this hook emision in the `changed()` callback.

I'll try that, thanks!

> Another thing regarding compatibility: This PR changes the semantics from `tag_renamed new` to `tag_renamed old new`, so if someone has code that listens for it and uses the first parameter, then the code now gets the old instead of the new tag name. While I think it is fine to introduce this incompatibility since the order "old new" makes much more sense, we should mention this in the `MIGRATION` file.

I'll add an entry in the next version of this PR. Although as pointed out in #999 tag_renamed isn't even officially documented?

t-wissmann:
> > I think the best way is to have a private member in the `Tag` class `oldName_` that is used here. By the way, in the "Validator"-callbacks, the value of the attribute is still the old one, but I think it's cleaner to run this hook emision in the `changed()` callback.
> 
> I'll try that, thanks!

As you noticed already, `private` doesn't work here; but I think `public` or `friend` are both OK.


> > Another thing regarding compatibility: This PR changes the semantics from `tag_renamed new` to `tag_renamed old new`, so if someone has code that listens for it and uses the first parameter, then the code now gets the old instead of the new tag name. While I think it is fine to introduce this incompatibility since the order "old new" makes much more sense, we should mention this in the `MIGRATION` file.
> 
> I'll add an entry in the next version of this PR. Although as pointed out in #999 tag_renamed isn't even officially documented?

I'd say we should put a warning there, since `tag_renamed` exists for quite a while already

blarz:
> > > I think the best way is to have a private member in the `Tag` class `oldName_` that is used here. By the way, in the "Validator"-callbacks, the value of the attribute is still the old one, but I think it's cleaner to run this hook emision in the `changed()` callback.
> > 
> > 
> > I'll try that, thanks!
> 
> As you noticed already, `private` doesn't work here; but I think `public` or `friend` are both OK.

Yes, I struggled a bit but then added a public member :)

> > > Another thing regarding compatibility: This PR changes the semantics from `tag_renamed new` to `tag_renamed old new`, so if someone has code that listens for it and uses the first parameter, then the code now gets the old instead of the new tag name. While I think it is fine to introduce this incompatibility since the order "old new" makes much more sense, we should mention this in the `MIGRATION` file.
> > 
> > 
> > I'll add an entry in the next version of this PR. Although as pointed out in #999 tag_renamed isn't even officially documented?
> 
> I'd say we should put a warning there, since `tag_renamed` exists for quite a while already

Makes sense. I added an entry, although I put it under "current git version" because I don't know what the next version is (and also I don't know since when tag_renamed has been around with one parameter).

t-wissmann (approved):
Looks good, thanks!

Issue #937, opened 2020-06-22
Urgent window uses “tiling” theme instead of “minimal”


tiosgz:
~~I couldn’t reproduce with anything else than Vivaldi, hence the title.~~ Seems to happen also with PCManFM-Qt when navigating folders.

Every time I open a new tab, the window uses `theme.tiling.active` instead of `theme.minimal.active` until I click something or change focus to another window (then it uses `minimal` again). As per what `hc -i` says, it seems the window marks & then unmarks itself (or is marked & unmarked) urgent, if that helps. I seem to have `focus_stealing_prevention` on (though I thought it’s off).

I guess it has something to do with #934 but I’m not sure as I have updated from 0.8.3 or a few commits before.

Do you have any idea what happens or do you want me to try to find out more?

(Also while testing this I’ve had some issues with border width and Firefox but I’ll keep these for later.)

tiosgz:
OK so I finally found a way how to get more info (it was “really difficult” :-P). Urgent client is always decorated as if it was in a non-`max` leaf with more clients even if it’s alone and both `smart_window_surroundings` and `smart_frame_surroundings` are on.

t-wissmann:
> OK so I finally found a way how to get more info (it was “really difficult” :-P). Urgent client is always decorated as if it was in a non-`max` leaf with more clients even if it’s alone and both `smart_window_surroundings` and `smart_frame_surroundings` are on. (Another edit:) Oh, and it keeps this till client switching.

I can't reproduce the issue with pcmanfm and new tubs, but I think I know where it comes from: `Client::setup_border()` is called whenever the urgency state changes, and it calls `Client::getDecTriple()` to find out which child of `theme.` to use, and this only takes the tiling and  fullscreen state of the client into account, but ignores these `smart_window_surroundings` aspects.

I think the fix (and anyway cleaner method) is that the window remembers the `DecTriple` (i.e. the child of the `theme.` object) which has been used last, and uses that when the urgency state changes.

> Edit: also, as a side note, I think a client should be marked urgent only if it’s not active.

I'm not sure about this one, but you're right in the sense that I don't know a single application that sets itself urgent while it's active anyway. The EWMH-Spec is quite vague regarding this:



[...]

PR #1017, opened 2020-10-27
Fix theme of urgent clients with smart_window

blarz:
As #937 reported, if smart_window_surroundings is enabled, the only
client on a tag correctly uses the minimal theme for surroundings. If
however the client gets an urgency hint, it suddenly uses the normal
theme for displaying the urgent border.

Add a test for this and fix it by caching the most recently used theme
type in order to choose the correct one when drawing the urgent border
(as suggested by Thorsten).


t-wissmann (approved):
Looks good & clean :)

PR #1019, opened 2020-10-31
Fix the value of WM_STATE for hidden windows

t-wissmann:
A window that is managed but hidden must have WM_STATE=IconicState
whereas hlwm set it to WithdrawnState before. This is described in
ICCCM[1] but also in the implementation note on virtual desktops in
EWMH[2].

Also, when a window is managed but moved to a hidden tag by the rules,
hlwm didn't set WM_STATE at all.

To make it clear to taskbars and pagers that this IconicState does not
mean 'minimized to taskbar', I've added the _NET_WM_STATE_HIDDEN to
_NET_SUPPORTED to make it clear that hlwm would set _NET_WM_STATE_HIDDEN
flag if a window was minimized.

Both these bugs regarding WM_STATE are fixed by the present commit and
checked via test cases.

[1] https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.3.1
[2] https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46035376813792

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,8 +9,8 @@ Current git version
 
   * Bug fixes:
     - When hiding windows, correctly set their WM_STATE to IconicState (we set
-      it to WithDrawn state before, which means "unmanaged" and thus is wrong).
-      This may require restarting pagers when uprading hlwm live.
+      it to Withdrawn state before, which means "unmanaged" and thus is wrong).
+      This may require restarting pagers when upgrading hlwm live.
 
 Release 0.9.0 on 2020-10-31
 ---------------------------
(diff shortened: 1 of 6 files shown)


t-wissmann:
I've double checked with openbox, and openbox also sets `WM_STATE` to `IconicState`.

t-wissmann:
Thanks, I've fixe them!

blarz (approved):
Looks good to me!

t-wissmann:
Codecov seems to be broken at the momen:
```
3.19s$ bash <(curl -s https://codecov.io/bash) -f coverage.info
/dev/fd/63: line 2: syntax error near unexpected token `<'
/dev/fd/63: line 2: `<html><head>'
The command "bash <(curl -s https://codecov.io/bash) -f coverage.info" exited with 2.
cache.2
store build cache
```

PR #1020, opened 2020-10-31
Remove old debug functions ls() and print()

t-wissmann:
This removes unused duplicate printing methods that provide a subset of
the features of the `attr` command.


blarz (approved):
I love approving this kind of PR!

Issue #71, opened 2016-05-24
Can not get current frame layout using attributes


gayprogrammer:
I would like to use an attribute like `tags.focus.curframe_layout` to find the current frame layout (tiling algorithm), but it does not exist.

My goal is to display the current frame's layout (tiling algorithm) in my panel. The easiest way would be if a hook was emitted when the layout changes. This is what I tried to do:
`herbstclient substitute LAYOUT tags.focus.curframe_layout emit_hook layout_changed LAYOUT`

Will it be possible to add something like `curframe_layout` or `tiling_algorithm` to the attributes?

bklaase:
While probably not exactly what you are looking for, the layout of the current frame can best be obtained from running:
herbstclient layout

```
% herbstclient layout 
╾─┐ horizontal 50% selection=0
  ├─╼ vertical: 0x2a000d2 0x2a00235 [FOCUS]
  └─╼ vertical: 0x2a00004
```

The "[FOCUS]" gives you easy access to finding the right frame. Now your remaining problem is finding the best way to know when to ask for this info. What I would suggest is plugging it in a subset of hooks and events that together would constitute every possible event of being in a different layout:
**Hooks**
- focus_changed (also fires when switching to an empty frame (and therefore different tag or monitor))

**user actions / events**
- frame deletion (only required when the focus would not change to another frame/client)
- frame layout change

You would need to add your script to the shortcuts or automatic triggers that can cause the above events to happen. I *think* that these three would give you the complete set to have the behaviour that you require.

The alternative is of course easier to maintain, but less elegant: simply keep polling/parsing `herbstclient layout `.

The-Compiler:
Some ~~months~~ years ago, I worked on [frame objects](http://git.the-compiler.org/herbstluftwm/log/?h=frame_objects) which would have that attribute - however it was reverted after merging due to a crash, and then we decided to wait until [winterbreeze](https://github.com/herbstluftwm/herbstluftwm/tree/winterbreeze) (the "proper" C++ rewrite) was merged and I'll pick that up then again.

bklaase:
Is there news on this issue? :) 
I'm using the method I desribe above when I see a focus_changed, or framelayout_changed, which works fine. Just curious

t-wissmann:
No, there aren't news, sorry. But my plan is to implement the frame objects!

t-wissmann:
@bklaase , @gayprogrammer: you can now do the following in the git-version:
```
[...]

PR #1021, opened 2020-10-31
Add dynamic children DynChild_<> for objects

t-wissmann:
This allows that a child entry of an object is not an explicit Object*
but a lambda instead that dynamically returns the Object* in this child
entry. This is usefull for entries that are indirectly determined by
other values. For those entries it would be much more coding effort to
keep the links up to date, so instead we just look the information up
when the information is queried, just as in the case of DynAttribute_.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,7 +7,7 @@ older versions.
 Current git version
 -------------------
 
-  * A child object 'focused_client' for each tag object.
+  * New child object 'focused_client' for each tag object.
   * Bug fixes:
     - When hiding windows, correctly set their WM_STATE to IconicState (we set
       it to Withdrawn state before, which means "unmanaged" and thus is wrong).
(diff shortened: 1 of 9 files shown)


t-wissmann:
- I didn't update the usage of `children_` in the dead code in object.cpp, since the code is removed anyway by #1020.
  - I mainly added this to have a link `tags.0.tiling.focused_frame` to the focused frame on tag `0` to implement the request of #71.

NEWS
```diff
@@ -7,6 +7,7 @@ older versions.
 Current git version
 -------------------
 
+  * A child object 'focused_client' for each tag object.
```
blarz, line 10: Maybe add a verb somewhere in that sentence?

blarz (commented):
Just a tiny comment on the NEWS entry, otherwise looks great!

NEWS
t-wissmann: I've replaced 'A' by 'New', but added no verb because this is essentially how the items in the NEWS file were written so far: https://herbstluftwm.org/news.html (I think it makes sense, because otherwise the verbs will become very repetitive)

blarz: Thanks! Makes sense and fits better into the pattern of the other entries

blarz (approved):

PR #1022, opened 2020-11-01
Make codecov interaction optional

t-wissmann:
Also switch from curl to wget, which does not print the response body to
stdout on failure. Before, curl printed the error message in html
format, which was piped directly into bash.


t-wissmann:
I don't know how long codecov will have problems, so this PR makes the codecov report optional.

PR #1024, opened 2020-11-01
Parse doc string from sourcecode

t-wissmann:
This contributes the following:

  - A new C++ base class HasDocumentation is added. It has a member
    function setDoc() which allows to set user documentation within the
    hlwm source code. So far, there is no functionality to pass the doc
    string further to the user
  - doc/gendoc.py is extended such that it reads these invocation of
    setDoc() and puts this documentation text into the hlwm-doc.json

### DIFF
--- a/doc/gendoc.py
+++ b/doc/gendoc.py
@@ -682,7 +682,6 @@ class TokTreeInfoExtrator:
             if stream.try_match(arg1, '.', 'setDoc', parameters, ';'):
                 doc_tokens = parameters.value.enclosed_tokens
                 doc_string = ast.literal_eval(' '.join(doc_tokens))
-                # print(f"doc: {classname}::{arg1.value} => {doc_string}")
                 self.objInfo.member_doc(classname, arg1.value, doc_string)
             else:
                 stream.pop()
--- a/doc/gendoc.py
+++ b/doc/gendoc.py
@@ -317,6 +317,7 @@ class ObjectInformation:
             self.default_value = None
             self.attribute_class = None  # whether this is an Attribute_ or sth else
             self.constructor_args = None  # the arguments to the constructor
+            self.doc = None  # a longer doc string
 
         def add_constructor_args(self, args):
             # split 'args' by the ',' tokens
@@ -389,6 +390,7 @@ class ObjectInformation:
             self.child_class = None  # whether this is a Link_ or a Child_
             self.type = None  # the template argument to Link_ or Child_
             self.constructor_args = None
+            self.doc = None  # a longer doc string
 
         def add_constructor_args(self, args):
             if len(args) >= 4:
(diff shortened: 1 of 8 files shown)


doc/gendoc.py
```diff
@@ -656,6 +672,20 @@ def stream_consume_member_initializers(self, classname, stream):
                 self.objInfo.member_init(classname, arg1.value, init_list)
             else:
                 stream.pop()
+        # we're now on a codeblock
+        self.stream_consume_setDoc(classname, TokenStream(codeblock.value.enclosed_tokens))
+
+    def stream_consume_setDoc(self, classname, stream):
```
blarz, line 678: Should test_gendoc.py be extended for the whole doc parsing thing?

src/tag.cpp
```diff
@@ -57,6 +57,15 @@ HSTag::HSTag(string name_, TagManager* tags, Settings* settings)
     floating_focused.setValidator([this](bool v) {
         return this->floatingLayerCanBeFocused(v);
     });
+
+    floating.setDoc("if the entire tag is set to floating mode");
```
blarz, line 61: For me it feels weird to extend the code by a new class in order to extend documentation/generated JSON.

blarz (changes requested):

doc/gendoc.py
t-wissmann: There is no way to obtain the doc string from a running hlwm instance, so we can't test it. Once there is a `help` command, I planned to test it.

src/tag.cpp
t-wissmann: In the end, I'd like to have a command `help` with the following behaviour:
```
$ hc help tags.0.floating
type: boolean setting
current value: false
meaning: if the entire tag is set to floating mode
```
https://github.com/herbstluftwm/herbstluftwm/pull/998

doc/gendoc.py
blarz: I see, thanks!

src/tag.cpp
blarz: My initial dislike was directed to the latter, but seeing the example of the help command, it makes much more sense to me now. I didn't realize the potential of that plan in the commit message of other PR.

I was just confused because normally "documentation" of this sort was kind of generated by doxygen or doxygen comments, but I know that this is not something we can use here.

So I'm gonna approve this (after a few minor typos are fixed) and see where this goes :)

blarz (changes requested):
Two minor things. Also the commented out setDoc call in frametree.cpp can be enabled, right?

t-wissmann:
I've removed the comment, added the setDoc and fixed the typo.

Unfortunatelly, I didn't manage to accomplish this documentation generation with doxygen. Doxygen seems to work well for api documentation but less well for user documentation.

blarz:
> I've removed the comment, added the setDoc and fixed the typo.

Thanks!

> Unfortunatelly, I didn't manage to accomplish this documentation generation with doxygen. Doxygen seems to work well for api documentation but less well for user documentation.

I know, it's just my goto source documentation software I guess.
But your approach is interesting, let's see how it works out :)

blarz (approved):

PR #1025, opened 2020-11-01
Make index attribute of tags writable

t-wissmann:
This requires the respective array operations in the IndexingObject and
also it requires that the EWMH properties are updated accordingly.
While writing test cases for that, I noticed that the EWMH properties
are not updated properly when a tag is removed (this changes the index
of the clients on later tags).

The wiring of the signals now solves both issues.

### DIFF
--- a/src/indexingobject.h
+++ b/src/indexingobject.h
@@ -4,7 +4,6 @@
 #include <cstddef>
 #include <string>
 #include <vector>
-#include <algorithm>
 
 #include "attribute_.h"
 #include "object.h"
--- a/src/indexingobject.h
+++ b/src/indexingobject.h
@@ -103,7 +103,6 @@ public:
         if (newIndex == oldIndex) {
             return;
         }
-        printf("fixing indices\n");
         // go from newIndex to oldIndex
         int delta = (newIndex < oldIndex) ? 1 : -1;
         T* lastValue = object;
--- a/src/indexingobject.h
+++ b/src/indexingobject.h
@@ -4,6 +4,7 @@
 #include <cstddef>
 #include <string>
 #include <vector>
+#include <algorithm>
 
 #include "attribute_.h"
 #include "object.h"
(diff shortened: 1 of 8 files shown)


src/indexingobject.h
```diff
@@ -83,7 +84,48 @@ class IndexingObject : public Object {
...
+        }
+        int oldIndexSigned = index_of(object);
+        if (oldIndexSigned < 0) {
+            return;
+        }
+        size_t oldIndex = static_cast<size_t>(oldIndexSigned);
+        if (newIndex == oldIndex) {
+            return;
+        }
+        printf("fixing indices\n");
```
blarz, line 106: Debug leftover?

tests/test_ewmh.py
```diff
@@ -28,6 +28,44 @@ def test_net_wm_desktop_after_bring(hlwm, x11):
...
+def test_net_wm_desktop_after_tag_index_increment(hlwm, x11):
+    hlwm.call('add tag1')
+    hlwm.call('add tag2')
+    hlwm.call('add tag3')
+    # put a new window on the tag with index 1
+    hlwm.call('rule tag=tag1')
+    win, winid = x11.create_client()
+    assert x11.get_property('_NET_WM_DESKTOP', win)[0] == 1
+
+    # move the tag3 to the begin of the list
```
blarz, line 40: beginning

tests/test_ewmh.py
```diff
@@ -28,6 +28,44 @@ def test_net_wm_desktop_after_bring(hlwm, x11):
...
+    assert x11.get_property('_NET_WM_DESKTOP', win)[0] == 3
+
+    if method == 'index_change':
+        # move the tag1 to the end of the list
+        hlwm.call('attr tags.by-name.tag1.index 3')
+    elif method == 'tag_removal':
+        # remove 'tag1' so all later tags experience a index shift
+        hlwm.call('merge_tag tag1 default')
+
+    # so the index of the tag of the client increased by one
```
blarz, line 64: decreased?

blarz (changes requested):

src/indexingobject.h
t-wissmann: Oh right, thanks!

t-wissmann:
Thanks, I've also fixed the two comments :)

blarz:
While playing around with this change I noticed that default panel.sh plays nicely with this, but I am using xfce4-panel and it seems to be dragging behind somehow:

I am on a tag with just one terminal, in this terminal I change the index of that tag. In the "workspace view" of xfce4-panel and in the list of windows for that workspace I still see that terminal listed under the old index.

Is there maybe something else that needs to get updated or set to "dirty"?

t-wissmann:
Hm there are still multiple bugs, that I can reproduce. I'm on it now.

t-wissmann:
There were two issues:
  - I forgot to update `_NET_CURRENT_DESKTOP` (index of the focused tag)
  - Regarding your issue: Originally, I tried to automatically correct it if the user supplies an index that is too large. However, this implied that I had to do thiese double `index != new_index` checks and in the end I know require that the index supplied by the user is in the interval [0, tag_count)

I've added test cases for both issues

blarz (approved):
Looks good to me!

PR #1027, opened 2020-11-04
Expose visibility information via the object tree

t-wissmann:
This extends the tag and client objects by a new read-only attribute
'visible'. It indicates whether a tag (resp. client) is shown on the
screen at the moment.

The information is there anyway, and I found it helpful when debugging
the first drafts of the window minimization code.

### DIFF
--- a/src/tag.cpp
+++ b/src/tag.cpp
@@ -31,6 +31,7 @@ static bool    g_tag_flags_dirty = true;
 HSTag::HSTag(string name_, TagManager* tags, Settings* settings)
     : frame(*this, "tiling")
     , index(this, "index", 0, &HSTag::isValidTagIndex)
+    , visible(this, "visible", false)
     , floating(this, "floating", false, [](bool){return "";})
     , floating_focused(this, "floating_focused", false, [](bool){return "";})
     , name(this, "name", name_,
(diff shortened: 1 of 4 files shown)


src/tag.cpp
```diff
@@ -142,8 +144,9 @@ void HSTag::applyFloatingState(Client* client)
     needsRelayout_.emit();
 }
 
-void HSTag::setVisible(bool visible)
+void HSTag::setVisible(bool newVisible)
 {
+    visible = newVisible;
```
blarz, line 149: Do we also need to check if visible already has the value of newVisible here?

blarz (commented):
Looks good, just one question (if "no" is the answer, this can go in ;))

src/tag.cpp
t-wissmann: I'd say 'no' because we didn't check it before either and all the functions we call here have their own visibility flag anyway. Also, if we don't do any shortcuts here, we know that after `HSTag::setVisible()` the visibility-flags of all windows on this tag are in sync.

blarz: Ok, makes sense, thanks!

blarz (approved):

PR #1029, opened 2020-11-04

bmwiedemann:

See https://reproducible-builds.org/ for why this is good

CMake already uses `SOURCE_DATE_EPOCH` for `TIMESTAMP`
The SOURCE_DATE_EPOCH variable is documented in

Also use UTC to be independent of timezone.

This is mostly a rebase of #553 that I finally managed to test on 0.9.0 now.

This PR was done while working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -16,6 +16,7 @@ function(gen_manpage sourcefile man_nr)
     set(src_orig "${CMAKE_CURRENT_SOURCE_DIR}/${sourcefile}.txt")
     set(src "${CMAKE_CURRENT_BINARY_DIR}/${sourcefile}.txt")
     set(dst "${CMAKE_CURRENT_BINARY_DIR}/${sourcefile}.${man_nr}")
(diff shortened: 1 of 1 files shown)


t-wissmann (approved):
Thanks for fixing it! Sorry that I forgot about your original PR after winterbreeze had been merged.

Am I correct, that now, package maintainers can set `BUILD_DATE` manually to achieve a reproducable build? By the way, for the man pages in released versions, the man page date can also be taken from the first line matching `Release.* on .*` and only for git versions we need to improvise (e.g. take a `TIMESTAMP` or look up the `git log`)

bmwiedemann:
maintainers can set `SOURCE_DATE_EPOCH` and cmake will use that to fill BUILD_DATE.
rpmbuild and Debian's and ArchLinux' package build automatically set `SOURCE_DATE_EPOCH` to the latest package changelog entry date, so it can be reproducible by default there.

For reproducible git builds, you could set it to `git log -1 --format=%ct`

PR #1030, opened 2020-11-04
Sort input file list

bmwiedemann:
Sort input file list
so that `hlwm-doc.json` builds in a reproducible way
in spite of indeterministic filesystem readdir order

See https://reproducible-builds.org/ for why this is good.

This PR was done while working on reproducible builds for openSUSE.

### DIFF
--- a/doc/gendoc.py
+++ b/doc/gendoc.py
@@ -13,7 +13,8 @@ def findfiles(sourcedir, regex_object):
     filename matches 'regex_object'
     """
     for root, dirs, files in os.walk(sourcedir):
-        for file in files:
+        dirs.sort()
+        for file in sorted(files):
             if regex_object.match(file):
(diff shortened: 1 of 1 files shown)


t-wissmann (approved):
Thanks! I'm surprised that the input file order has an effect on the output because I thought the output is entireliy determined by the 'dict' datatype. It seems I was wrong, so it's good to have this. Thanks for the patch and thanks for packaging hlwm :)

(Note to self: I should also remove the unused loop variables here)

bmwiedemann:
ah, might be because we `export PYTHONHASHSEED=0` in our build env. In that case, sorting after the dict would be a good addition for those that dont set it.

t-wissmann:
I've also reproduced the change of the order locally (the diff posted above has been generated locally).

PR #1032, opened 2020-11-07
Implement client minimization

t-wissmann:
This adds a new attribute 'minimized' that hides a client until it is
unminimized. Only floating clients can be minimized, so we move a tiled
client from the 'tiling' to the 'floating' structures in a tag. Such a
client still has 'floating' set to false and will be moved back to the
tiling layer if it gets un-minimized.

It is important that hlwm ensures that minimized clients can not be
focused. If a minimized client shall be focused, it has to be
unminimized first. This is not yet respected by the present PR in the
cycle_all command but must be implemented next.

There is no way to manipulate the minimization state through
EWMH-Requests, but I found the following keybinding helpful for testing.

    herbstclient keybind Alt-m attr clients.focus.minimized toggle

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,7 @@ older versions.
 Current git version
 -------------------
 
+  * Client minimization (controlled by the attribute 'minimized' of every client)
   * The 'index' attribute of tags is now writable. This allows adjusting the
     order of existing tags.
   * New child object 'focused_client' for each tag object.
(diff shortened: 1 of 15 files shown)


t-wissmann:
Making window minimization work gave me more trouble than I expected. First I tried to implement that minimized windows can be focused and are made visible temporarily until they loose the focus again. However, this caused a lot of interference with multiple monitors so now I'm following a new invariance: a minimized client can not be focused. If you want to focus one, unminimize it first (this is done automatically by the `jumpto` command resp. the `focus_client()`-function).

The skipping of minimized clients is still missing in cycle_all, but I think it's ok to do this in a separate PR, since the present PR #1032 is big enough.

blarz (approved):
Looks good to me!

The keybind of the commit message could probably be added to the default autostart.sh soon?
And maybe we the documentation could be extended too?

But all this can be done in a separate PR, let's get the base in first :)

PR #1034, opened 2020-11-14
Fix moving a minimized client between tags

t-wissmann:
This sets the visibility of minimized clients correctly when they are
moved between tags. The test cases for it use the 'apply_rules' command
to move a client between tags since there is no direct way to do so via
the object system but it is possible via EWMH (see the test case
test_ewmh_move_client_to_tag), but I wanted the test cases in the
present change to be independent of X11.

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -393,7 +393,7 @@ def test_minimized_window_stays_floating_on_tag_change(hlwm):
     assert hlwm.get_attr(f'clients.{winid}.tag') == this_tag
 
     # the minimized client is not in the tiling layer
-    assert hlwm.get_attr(f'tags.focus.tiling.root.client_count') == '0'
+    assert hlwm.get_attr('tags.focus.tiling.root.client_count') == '0'
     assert hlwm.get_attr(f'clients.{winid}.floating') == 'false'
     # but after un-minimizing it, it is
     hlwm.call(f'set_attr clients.{winid}.minimized false')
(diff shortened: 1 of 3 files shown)


tests/test_tags.py
```diff
@@ -351,3 +351,50 @@ def test_focused_client_multiple_tags(hlwm):
...
+    # after bringing the client to a visible tag, it must stay
+    # invisible
+    assert hlwm.get_attr(f'clients.{winid}.minimized') == 'true'
+    assert hlwm.get_attr(f'clients.{winid}.visible') == 'false'
+    # and the minimized client is not in the tiling layer!
+    assert hlwm.get_attr('tags.focus.tiling.root.client_count') == '0'
+    assert hlwm.get_attr(f'clients.{winid}.floating') == 'false'
+
+
+def test_minimized_window_stays_floating_on_tag_change(hlwm):
```
blarz, line 381: The test case name is a bit confusing, The client is neither floating nor tiling after the tag change?

blarz:
Updated the commit message as it said "there is no direct via to do so via". Changed the first via to way.

tests/test_tags.py
t-wissmann: Hm right. Maybe it should be called `test_minimized_window_stays_minimized_on_tag_change`

t-wissmann:
> Updated the commit message as it said "there is no direct via to do so via". Changed the first via to way.

Thanks!

tests/test_tags.py
blarz: Thanks!

blarz (approved):

PR #1035, opened 2020-11-14
Correctly 'load' minimized clients to a frame tree

t-wissmann:
This makes the 'load' command handle minimized clients correctly and
adds test cases for it. I'm also setting 'floating' to 'false', since
the client is sucked into the frame tree by the 'load' command and thus
definitely not in the floating state anymore.

### DIFF
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -676,6 +676,10 @@ void FrameTree::applyFrameTree(shared_ptr<Frame> target,
         // this might even involve the above targetLeaf / targetSplit
         // so we need to do this before everything else
         for (const auto& client : sourceLeaf->clients) {
+            // first un-minimize and un-float the client
+            // such that we now that it is in the frame-tree
+            client->floating_ = false;
+            client->minimized_ = false;
             client->tag()->frame->root_->removeClient(client);
             if (client->tag() != tag_) {
                 client->tag()->stack->removeSlice(client->slice);
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -677,7 +677,7 @@ void FrameTree::applyFrameTree(shared_ptr<Frame> target,
         // so we need to do this before everything else
         for (const auto& client : sourceLeaf->clients) {
             // first un-minimize and un-float the client
-            // such that we now that it is in the frame-tree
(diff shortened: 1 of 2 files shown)


src/frametree.cpp
```diff
@@ -676,6 +676,10 @@ void FrameTree::applyFrameTree(shared_ptr<Frame> target,
         // this might even involve the above targetLeaf / targetSplit
         // so we need to do this before everything else
         for (const auto& client : sourceLeaf->clients) {
+            // first un-minimize and un-float the client
+            // such that we now that it is in the frame-tree
```
blarz, line 680: s/now/know/?

tests/test_load.py
```diff
@@ -168,3 +168,46 @@ def test_fraction_precision_outside_range(hlwm):
...
             .expect_stderr('but actually is ' + v)
+
+
+def test_load_floating_client(hlwm):
+    winid, _ = hlwm.create_client()
+    hlwm.call(f'set_attr clients.{winid}.floating true')
+    hlwm.call('set_layout max')
+    assert hlwm.call('dump').stdout.rstrip() == '(clients max:0)'
+
+    # soak the client into the frame tree
```
blarz, line 179: Isn't it "suck" more than "soak"?

tests/test_load.py
```diff
@@ -168,3 +168,46 @@ def test_fraction_precision_outside_range(hlwm):
...
+    # and such that it is invisible initially
+    (True, True), (True, False), (False, True)
+])
+@pytest.mark.parametrize("floating", [True, False])
+def test_load_minimized_client(hlwm, othertag, minimized, floating):
+    if othertag:
+        hlwm.call('add othertag')
+        hlwm.call('rule tag=othertag')
+    winid, _ = hlwm.create_client()
+    if minimized:
```
blarz, line 199: Could be removed

t-wissmann:
Thanks, I've fixed all three issues.

tests/test_load.py
blarz: I also updated it in the commit message

blarz (approved):

blarz:
> Thanks, I've fixed all three issues.

Thanks! Approved :)

PR #1036, opened 2020-11-15
Let cycle_all skip minimized clients

t-wissmann:
Following the principle of #1032 that minimized clients shall never be
focused or visible, the cycle_all command must skip minimized clients.
This implements the skipping of minimized clients and adds a number of
test cases for it.

While at it, I've removed a unnecessary call to the raise() function of
the newly focused client. It was unnecessary because the raise()
function is called anyway at the end of cycleAllCommand() for floating
clients.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -341,7 +341,7 @@ cycle_all [*--skip-invisible*] ['DIRECTION']::
     acts similar to the 'cycle' command. If *--skip-invisible* is given,
     then this only cycles through all visible windows and skips invisible
     windows in the max layout (the flag only affects invisible windows in the
-    max layout; minimized windows are always skipped). After after each focus
+    max layout; minimized windows are always skipped). After each focus
     change, the focused window is raised.
 
 cycle_frame ['DIRECTION']::
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -335,14 +335,14 @@ cycle ['DELTA']::
     means: cycle in the opposite direction by 1.
 
 cycle_all [*--skip-invisible*] ['DIRECTION']::
-    Cycles through all windows and frames on the current tag. 'DIRECTION' = 1
-    means forward, 'DIRECTION' = -1 means backward, 'DIRECTION' = 0 has no
-    effect. 'DIRECTION' defaults to 1. If there are multiple windows within one
-    frame, then it acts similar to the 'cycle' command. (The 'cycle_all' command
-    focuses the next/previous leaf in the 'layout' tree.). If
-    *--skip-invisible* is given, then this only cycles through all visible
-    windows and skips invisible windows in the max layout. The focused window
-    is raised.
+    Cycles through all non-minimized windows and frames on the current tag.
+    'DIRECTION' = 1 means forward (default value), 'DIRECTION' = -1 means backward, 'DIRECTION'
+    = 0 has no effect. If there are multiple windows within one frame, then it
+    acts similar to the 'cycle' command. If *--skip-invisible* is given,
+    then this only cycles through all visible windows and skips invisible
+    windows in the max layout (the flag only affects invisible windows in the
+    max layout; minimized windows are always skipped). After after each focus
+    change, the focused window is raised.
 
 cycle_frame ['DIRECTION']::
     Cycles through all frames on the current tag. 'DIRECTION' = 1 means forward,
--- a/src/tag.cpp
+++ b/src/tag.cpp
@@ -361,6 +361,13 @@ int HSTag::cycleAllCommand(Input input, Output output)
     }
     if (floating_focused()) {
         int newIndex = static_cast<int>(floating_clients_focus_) + delta;
+        // skip minimized clients
+        while (newIndex >= 0
+               && static_cast<size_t>(newIndex) < floating_clients_.size()
+               && floating_clients_[newIndex]->minimized_())
+        {
+            newIndex += delta;
+        }
         if (newIndex < 0) {
             floating_focused = false;
             frame->cycleAll(FrameTree::CycleDelta::End, skip_invisible);
@@ -369,7 +376,6 @@ int HSTag::cycleAllCommand(Input input, Output output)
             frame->cycleAll(FrameTree::CycleDelta::Begin, skip_invisible);
         } else {
             floating_clients_focus_ = static_cast<size_t>(newIndex);
-            floating_clients_[floating_clients_focus_]->raise();
         }
     } else {
         FrameTree::CycleDelta cdelta = (delta == 1)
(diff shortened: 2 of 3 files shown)


bew:
`cycle_all` already has a `--skip-invisible` option, I think it would make sense to skip minimized windows as well when this is passed, but not in all cases.
I'm worried about not having an option of reliably cycle all windows, without respect to their state. For example on Windows, Alt-tab will cycle all, even if they're minimized.

Another option would be to add an option similar to the invisible one, but for minimized windows: `--skip-minimized`.
This has the benefit that we can more precisely target windows I think.

What do you think?

blarz:
Fixed some typos in the commit message

tests/test_layout.py
```diff
@@ -395,6 +395,53 @@ def test_cycle_all_with_floating_clients(hlwm, delta, floating_clients):
...
+        winid, _ = hlwm.create_client()
+        hlwm.call(f'set_attr clients.{winid}.minimized {hlwm.bool(minimized)}')
+        if not minimized:
+            non_minimized_clients.append(winid)
+    if delta == -1:
+        non_minimized_clients = list(reversed(non_minimized_clients))
+    expected_winids = non_minimized_clients + [tiled_client]
+
+    assert hlwm.get_attr('clients.focus.winid') == tiled_client
+    print("expecting the clients in the order {}".format(' '.join(expected_winids)))
```
blarz, line 432: Leftover debug print?

blarz (changes requested):

tests/test_layout.py
t-wissmann: I wanted to keep it because it makes sense to have this if the test case fails.

t-wissmann:
> `cycle_all` already has a `--skip-invisible` option, I think it would make sense to skip minimized windows as well when this is passed, but not in all cases.
> I'm worried about not having an option of reliably cycle all windows, without respect to their state. For example on Windows, Alt-tab will cycle all, even if they're minimized.

Yes, I understand that that one wants to cycle through all windows including the minimized ones. The present change is only ensures that hlwm never goes into an invalid state (e.g. because a minimized window is visible).

For the cycle_all command it is not so easy to replicate the Alt-Tab behaviour of Windows and other window managers, because for this usual Alt-Tab behaviour, it is crucial whether <kbd>Alt</kbd> is kept pressed or released between two <kbd>Tab</kbd> presses:

  - if you alt-tab to a minimized window, keep alt pressed and press tab again, that window will be minimized.
  - if you alt-tab to a minimized window, release alt, then it is un-minimized and if you press alt-tab again, then the window stays unminimized.

But with the keybinding-system of hlwm, it is not possible to distinguish between these two cases. Concretely, if you `cycle_all` to a minimized window "M", then it will be unminimized. If you then run `cycle_all` again to some other window, should the window "M" be minimized again or should it stay visible?

> Another option would be to add an option similar to the invisible one, but for minimized windows: `--skip-minimized`.

Yes, I also thought about something like that (assuming the issue above is solved).

tests/test_layout.py
blarz: I see! Maybe add the message as an argument to the assert statement then, e.g. assert foo == bar, "expected foo to be bar"

And also use an f-string instead of .format()?

t-wissmann: I want the message in the log if one of the asserts in the loop fails, so I had to add it as a message to all of the asserts in the loop. I can replace `format` by an f-string if you strongly prefer it, but I don't see the advantage of it. I think the line is perfectly readable as it is. Writing it with an f-string implies that we need to nest single and double quotes:
```python
print(f"expecting the clients in the order {' '.join(expected_winids)}")
```

blarz: You are right, the f-string doesn't add more readability.

Should this assertion not only apply to the client.focus.winid assertion?
Adding the print statement to that assert (and maybe moving it up, before the minimized and visible asserts) should do the trick, or am I missing something?

t-wissmann: If you prefer it I can do it. I don't care I just want a reasonable test case and found the print statement helpful and don't really see what's so bad about.

blarz: The PR as such is very nice and the tests are great and look good to me! Sorry for focusing so much on that tiny detail. There is nothing bad about the print statement, I just stumbled over it while reviewing, because it felt "unusual" to have this in a unit test (at least that is my feeling).

In case the tests go green, this statement would not add any value, but on the other hand will be very helpful in case of the assertion being triggered and that is why I thought combining it as an argument for the assert statement would be a good solution.

src/tag.cpp
```diff
@@ -369,7 +376,6 @@ int HSTag::cycleAllCommand(Input input, Output output)
             frame->cycleAll(FrameTree::CycleDelta::Begin, skip_invisible);
         } else {
             floating_clients_focus_ = static_cast<size_t>(newIndex);
-            floating_clients_[floating_clients_focus_]->raise();
```
blarz, line 372: Just trying to make sure: This will be taken care of in the caller of cycleAllCommand, right?

t-wissmann: There is a `newFocus->raise()` in line 402 at the end of the `cycleAllCommand()` function. I should have mentioned that in the commit description already. (the commit/pr description mentions this now)

tests/test_layout.py
t-wissmann: Sorry for my frustated answer yesterday. I have no experience whether it is usual to have print statements in unit tests, if there are any reasonable arguements against such print statements I'd be interested to hear them.
For non-reproducible race conditions it is good to have as much information as possible in the log file when the race condition happens to happen. And this is why I didn't want to infer which assert statement is the one that fails.

src/tag.cpp
blarz: Awesome, thanks for adding it to the commit message!

tests/test_layout.py
blarz: There is no real reason against it, just my gut feeling that it is "unusual" and "I never do it this way" ;)

I am sorry, too. This may have read like I have stronger feelings about this than I actually have.

So let's close this discussion.

blarz (approved):
Looks good and tested locally

doc/herbstluftwm.txt
```diff
@@ -335,14 +335,14 @@ cycle ['DELTA']::
...
-    *--skip-invisible* is given, then this only cycles through all visible
-    windows and skips invisible windows in the max layout. The focused window
-    is raised.
+    Cycles through all non-minimized windows and frames on the current tag.
+    'DIRECTION' = 1 means forward (default value), 'DIRECTION' = -1 means backward, 'DIRECTION'
+    = 0 has no effect. If there are multiple windows within one frame, then it
+    acts similar to the 'cycle' command. If *--skip-invisible* is given,
+    then this only cycles through all visible windows and skips invisible
+    windows in the max layout (the flag only affects invisible windows in the
+    max layout; minimized windows are always skipped). After after each focus
```
blarz, line 344: After twice

blarz (changes requested):

blarz (approved):

PR #1039, opened 2020-11-18
Disallow 'drag' for minimized clients

t-wissmann:
It makes no sense to drag a minimized client, so lets be safe and forbid
it.


blarz (approved):
Makes sense :D

PR #1042, opened 2020-11-21
Test apply_rules for minimized clients

### LINKED ISSUE #1040: Add test cases for passing minimized clients to apply_rules
There should be test cases that verify that the `apply_rules` command correctly handles minimized clients, especially when the rules move the client

  * from an invisible to a visible tag
  * vice versa
  * between visible tags

t-wissmann:
Add tests that ensure that minimized clients stay invisible, even if
apply_rules brings them from an invisible to a visible tag.

This fixes #1040.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -480,6 +480,42 @@ def test_apply_rules_all_focus_retained(hlwm):
         assert hlwm.get_attr('clients.focus.winid') == client
 
 
+@pytest.mark.parametrize('floating', [True, False])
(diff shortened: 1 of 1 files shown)


t-wissmann:
Looking at the implementation of `apply_rules` it is unsurprising that everything works fine, because the implementation does not call `set_visible` on the client directly, but only calls `TagManager::moveClient()` which has the minimization-handling already implemented.

blarz (approved):
More tests, more good!

PR #1043, opened 2020-11-21
Make rule evaluation a member function of Rule

t-wissmann:
This moves the rule evaluation code into the class Rule. This has the
advantage that (a) the RuleManager does less member access to members of
the Rule class and (b) we can implement a command for applying a
standalone rule to a given window without adding the rule to the full
rule list.


blarz (approved):
Makes sense!

PR #1045, opened 2020-11-22
Extend ArgParse with boolean command line flags

t-wissmann:
This adds support for command line flags to the ArgParse class. A flag
can appear at any position in the command line arguments between the
(mandatory or optional) positional arguments. A flag is usually prefixed
with one or two dashes, but there is no formal requirement for this in
the ArgParse implementation.

A unknown flag will simply be ignored and parsed as a positional
argument instead. This is because I don't want negative numbers on the
command line to be misinterpreted as unknown command line flags.

For illustration and testing, this migrates the focus command to ArgParse with
flags.

### DIFF
--- a/src/argparse.cpp
+++ b/src/argparse.cpp
@@ -51,7 +51,7 @@ bool ArgParse::parsingFails(Input& input, Output& output)
         string valueString;
         input >> valueString;
         // try to parse this token as a flag
-        if (optionalArgumentsRemaining && tryParseFlag(valueString)) {
+        while (optionalArgumentsRemaining && tryParseFlag(valueString)) {
             optionalArgumentsRemaining--;
             // if this token was the last optional argument
             if (optionalArgumentsRemaining == 0 && arg.optional_) {
--- a/src/argparse.cpp
+++ b/src/argparse.cpp
@@ -100,7 +100,7 @@ bool ArgParse::parsingFails(Input& input, Output& output)
  * tokens left in the input.
  * @param input
  * @param output
- * @return whether there has been a
+ * @return whether there is an unparsable flag or unexpected argument at the end
  */
 bool ArgParse::parsingAllFails(Input& input, Output& output)
 {
--- a/src/argparse.cpp
+++ b/src/argparse.cpp
@@ -6,7 +6,11 @@ using std::pair;
 using std::string;
 
 /**
- * @brief try to parse the arguments
+ * @brief try to parse the positional arguments and as many flags
+ * as possible. When the last positional argument has been read
+ * then flags are still read until the first unknown token.
+ * If unknown tokens at the end of input should be regarded as
+ * an error, use parsingAllFails()
  * @param input
  * @param output
  * @return return whether there has been an error (true = error, false = no error)
@@ -91,6 +95,13 @@ bool ArgParse::parsingFails(Input& input, Output& output)
     return false;
 }
 
+/**
+ * @brief run parsingFails() and assert that there are no unknown
+ * tokens left in the input.
+ * @param input
+ * @param output
+ * @return whether there has been a
+ */
 bool ArgParse::parsingAllFails(Input& input, Output& output)
 {
     return parsingFails(input, output) || unparsedTokens(input, output);
@@ -111,7 +122,7 @@ bool ArgParse::unparsedTokens(Input& input, Output& output)
 
 /**
  * @brief Accept boolean flags (e.g. --all --horizontal ...) at
- * any position
+ * any position between the (mandatory or optional) positional arguments
  * @param a list of flags
  * @return
  */
--- a/src/argparse.cpp
+++ b/src/argparse.cpp
@@ -2,6 +2,7 @@
 
 #include <iostream>
 
+using std::pair;
 using std::string;
 
 /**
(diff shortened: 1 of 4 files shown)


src/argparse.cpp
```diff
@@ -60,8 +79,76 @@ bool ArgParse::parsingFails(Input& input, Output& output)
...
     // have not been parsed.
     return false;
 }
+
+/**
+ * @brief run parsingFails() and assert that there are no unknown
+ * tokens left in the input.
+ * @param input
+ * @param output
+ * @return whether there has been a
```
blarz, line 103: Missing word?

src/argparse.cpp
```diff
@@ -39,18 +45,31 @@ bool ArgParse::parsingFails(Input& input, Output& output)
...
+                continue;
+            }
+            // otherwise, there is room for another optional
+            // argument or 'arg' is mandatory. So get another
+            // token from the input, and parse it to the
+            // current 'arg'
+            input >> valueString;
+        }
+        // in any case, we have this here:
+        // assert (!arg.optional_ || optionalArgumentsRemaining > 0);
```
blarz, line 68: This statement is left commented out only for documentation purpose?

src/argparse.cpp
```diff
@@ -60,8 +79,76 @@ bool ArgParse::parsingFails(Input& input, Output& output)
...
+        return true;
+    }
+    return false;
+}
+
+/**
+ * @brief Accept boolean flags (e.g. --all --horizontal ...) at
+ * any position between the (mandatory or optional) positional arguments
+ * @param a list of flags
+ * @return
```
blarz, line 127: Left empty on purpose?

tests/test_layout.py
```diff
@@ -155,6 +155,54 @@ def test_focus_wrap(hlwm, running_clients, running_clients_num):
             hlwm.call('focus down')
 
 
+@pytest.mark.parametrize("mode", ['setting', 'flag'])
+@pytest.mark.parametrize("setting_value", [True, False])
+@pytest.mark.parametrize("external_only", [True, False])
+@pytest.mark.parametrize("running_clients_num", [3])
```
blarz, line 161: This confused me, what is the use of this parameter? And the name does not match the parameter in the list of the test function? Could be that I don't know enough about this fixture and this intended

blarz (changes requested):

tests/test_layout.py
t-wissmann: This parameter is implicitly used by the `running_clients` fixture and controls, how many clients are created. (`running_clients` is then a list of the defined size containing the clients). @dnnr implemented it this way back then if I recall correctly.

src/argparse.cpp
t-wissmann: Yes. Should I put `this` here?

src/argparse.cpp
t-wissmann: Yes. I don't want any crash if the assertion fails.

blarz: Ok!!

src/argparse.cpp
blarz: No, wouldn't add much value IMHO. Leave it is as is :)

tests/test_layout.py
blarz: Ok then. I just wondered because in the test_dump case below, both are used.
But seems to work, so closing this remark. Thanks for clearing it up!

src/argparse.cpp
blarz: Fixes

blarz:
I know it doesn't make much sense in this case, but shouldn't something like

```hc focus -e -i down``` be possible too?

t-wissmann:
> I know it doesn't make much sense in this case, but shouldn't something like
> 
> `hc focus -e -i down` be possible too?

Yes, that should be possible too, but it seems I made a mistake and should take a look at it again.

blarz:
> > I know it doesn't make much sense in this case, but shouldn't something like
> > `hc focus -e -i down` be possible too?
> 
> Yes, that should be possible too, but it seems I made a mistake and should take a look at it again.

Funny thing is that with the mirror command having to flags works (```--vertical --horizontal``` that is)

t-wissmann:
> > > I know it doesn't make much sense in this case, but shouldn't something like
> > > `hc focus -e -i down` be possible too?
> > 
> > 
> > Yes, that should be possible too, but it seems I made a mistake and should take a look at it again.
> 
> Funny thing is that with the mirror command having to flags works (`--vertical --horizontal` that is)

That's because for `mirror`, the flags are after all the positonal args. I've replaced an `if` by a `while` and now it works (but I also should add a test for it).

tests/test_layout.py
t-wissmann: > Ok then. I just wondered because in the test_dump case below, both are used.
> But seems to work, so closing this remark. Thanks for clearing it up!

That's because I didn't know it before this PR that also works without explicitly mentioning `running_clients_num` in the function parameters.

t-wissmann:
There's now a test for `focus -i -e down` (`flag-shadow`) and also `focus -e -e down` (`flag-double`)

blarz:
> There's now a test for `focus -i -e down` (`flag-shadow`) and also `focus -e -e down` (`flag-double`)

Thanks!

blarz (approved):

PR #1046, opened 2020-11-22
New command 'mirror' for mirroring the frame tree

t-wissmann:
This adapts the 'mirror' implementation from the PR #93 to the new
codebase. Additionally, it accepts a positional argument along which
axis the layout is to be mirrored.

On the implementation side, it uses type_traits and a new 'is_finite'
predicate to avoid boilerplate code when instantiating the Converter
class to the enum class type Horizontal/Vertical/Both. There is still a
bit to write, but much less compared to the Converter-implementation for
the ClientPlacement class.

In order to make Converter applicable to type traits, a new template
parameter to Converter "typename = void" was necessary.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -519,6 +519,10 @@ rotate::
     Rotates the layout on the focused tag counterclockwise by 90 degrees. This
     only manipulates the alignment of frames, not the content of them.
 
+mirror ['--vertical'|'--horizontal']::
+    Mirrors the layout on the focused tag vertically, horizontally, or both.
+    This only manipulates the alignment of frames, not the content of them.
+
 set 'NAME' 'VALUE'::
     Sets the specified setting 'NAME' to 'VALUE'. Allowed values for boolean
     settings are 'on' or 'true' for on, 'off' or 'false' for off, 'toggle' to
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -519,9 +519,10 @@ rotate::
     Rotates the layout on the focused tag counterclockwise by 90 degrees. This
     only manipulates the alignment of frames, not the content of them.
 
-mirror ['--vertical'|'--horizontal']::
-    Mirrors the layout on the focused tag vertically, horizontally, or both.
-    This only manipulates the alignment of frames, not the content of them.
+mirror ['vertical'|'horizontal'|'both']::
+    Mirrors the layout on the focused tag vertically, horizontally, or both; the
+    default is 'horizontal'. This command only manipulates the alignment of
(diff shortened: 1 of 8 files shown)


doc/herbstluftwm.txt
```diff
@@ -519,6 +519,10 @@ rotate::
     Rotates the layout on the focused tag counterclockwise by 90 degrees. This
     only manipulates the alignment of frames, not the content of them.
 
+mirror ['--vertical'|'--horizontal']::
```
blarz, line 522: I know that you want to use the new ArgParse functionality,
but does it make sense to have the layout mirror direction as optional flags?

A positional argument vertical/horizontal/both may also be a solution, right?
Either way is fine. But mirror without any flags is a no-op then, right?

blarz: But playing around with it right now, pretty nice functionality anyway!

t-wissmann: > A positional argument vertical/horizontal/both may also be a solution, right?

Right, I never thought about that. I think `vertical`/`horizontal`/`both` (with `horizontal` being the default) is more intuitive

t-wissmann:
I've updated the `mirror` command to have positional arguments now. As a side effect, this PR is not independent of PR #1045.

blarz:
> I've updated the `mirror` command to have positional arguments now. As a side effect, this PR is not independent of PR #1045.

Awesome I'll have a look at it tonight. Can you update the documentation, it still refers to the flags. Otherwise it looks good when I flew over the code.

t-wissmann:
> Awesome I'll have a look at it tonight. Can you update the documentation, it still refers to the flags. Otherwise it looks good when I flew over the code.

I've updated the man page. Since the CI does not check the doc, I saved some cpu time and cancelled the travis CI build, so don't be worried by the CI failure for 6de9972. The ci for the previous commit of this PR (6a5e75d) succeeded.

blarz:
> > Awesome I'll have a look at it tonight. Can you update the documentation, it still refers to the flags. Otherwise it looks good when I flew over the code.
> 
> I've updated the man page. Since the CI does not check the doc, I saved some cpu time and cancelled the travis CI build, so don't be worried by the CI failure for [6de9972](https://github.com/herbstluftwm/herbstluftwm/commit/6de9972060faa2746b72356b35bf0dc320ddcace). The ci for the previous commit of this PR ([6a5e75d](https://github.com/herbstluftwm/herbstluftwm/commit/6a5e75d10efb5fe79fd55f8f4fcffa81b010f7e6)) succeeded.

Thanks! It'll be build now anyway because of mergify. Let's see what happens :)

blarz (approved):
Nice!

PR #1047, opened 2020-11-22
Make rule parsing a static member of RuleManager

t-wissmann:
This is another step towards a command that applies a single temporary
rule. Since this does not change hlwm's behaviour, the existing set of
tests in test_rules.py should be sufficient.


t-wissmann:
I have the claimed `apply_rule` command (without `s`) in draft form locally, but I don't know when I will find time to finish it. Hence, lets do the present change first to keep the diff then smaller.

blarz (approved):

PR #1049, opened 2020-11-23
Use ArgParse in cycle_all

t-wissmann:
Migrate cycle_all to the new ArgParse and add tests for the standard
error messages. There are already tests for cycle_all, so this also
covers the successful parsing of the flag and the positional argument.


blarz (approved):
Nice :)

PR #1051, opened 2020-11-28
Add the 'apply_tmp_rule' command

t-wissmann:
It applies a temporary rule to one particular or all clients without
modifying or evaluating the existing rule set. I made the
error messages of the rule parser generic in the command name where
possible. There are still hard-coded 'rule:'-prefixes in error messages
in setLabel(), addCondition(), and addConsequence() in the Rule class,
but I didn't want to blow up neither the argument list nor this PR.


blarz (approved):
Looks good to me!

Haven't tried it locally, but the tests seem to cover it all, so my confidence is high ;)

Issue #1053, opened 2020-11-30
No rule to make target install


fallazc:
I am trying to package herbstluftwm 0.9.0 for Slackware but I seems that the install target for the Makefile doesn't exist. The same setup used to work for version 0.7.2. Is there any equivalent to that?

Here is how I attempt to build it:
```
rm -rf $PKG
mkdir -p $TMP $PKG $OUTPUT
cd $TMP
rm -rf $PRGNAM-$VERSION
tar xvf $CWD/$PRGNAM-$VERSION.tar.gz
cd $PRGNAM-$VERSION

make PREFIX=/usr/ MANDIR=/usr/man DOCDIR=/usr/doc$PRGNAM-$VERSION || exit 1
make DESTDIR=$PKG PREFIX=/usr/ MANDIR=/usr/man DOCDIR=/usr/doc$PRGNAM-$VERSION install
```
Thanks

The-Compiler:
With 0.8.0, the build system was changed to cmake - see the [install instructions](https://github.com/herbstluftwm/herbstluftwm/blob/master/INSTALL).

fallazc:
Fixed it, so I had to be in the build folder before running the make install.
I think you should make mention of it in the `Installation steps` section

blarz:
> Fixed it, so I had to be in the build folder before running the make install.
> I think you should make mention of it in the `Installation steps` section

This was indeed a missing step, thank you for reporting this! I fixed it with #1055

PR #1055, opened 2020-12-01
Fix INSTALL formatting, add out-of-tree build dir

blarz:
Maybe clears up some confusion that was mentioned in #1053

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -47,13 +47,15 @@ The CMake build system is used. It is recommended to use an out-of-tree build
 directory. The compilation and installation is configured by the following CMake
 variables which you can set on the initial cmake call and alter with ccmake:
 
-CMAKE_INSTALL_PREFIX         = /usr/            # the prefix
-CMAKE_INSTALL_SYSCONF_PREFIX = $(DESTDIR)/etc/  # path to etc directory
+    CMAKE_INSTALL_PREFIX         = /usr/            # the prefix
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -47,13 +47,15 @@ The CMake build system is used. It is recommended to use an out-of-tree build
...
-CMAKE_INSTALL_PREFIX         = /usr/            # the prefix
-CMAKE_INSTALL_SYSCONF_PREFIX = $(DESTDIR)/etc/  # path to etc directory
+    CMAKE_INSTALL_PREFIX         = /usr/            # the prefix
+    CMAKE_INSTALL_SYSCONF_PREFIX = $(DESTDIR)/etc/  # path to etc directory
 
 Individual paths for special files can be set with the *DIR variables, typically
 relative to CMAKE_INSTALL_PREFIX. If you are building package, you would
 typically build regularly, and install with DESTDIR=./path/to/fakeroot.
 
+    mkdir build && cd build
```
blarz, line 57: Feels a bit redundant to have it this here and in the three other places below

t-wissmann: Right. But I think it's fine to have this multiple times. I remember that I struggled a lot with these out-of-tree builds because it felt so unfamiliar to manually create a new folder when (only) compiling an existing project.

t-wissmann (approved):
Thanks!

INSTALL
blarz: Alright, I'll just leave the PR as is then, thank you for the feedback!

PR #1057, opened 2020-12-04
Fix race condition in test_tags_restored_after_wmexec

t-wissmann:
This fixes a race condition in a test case using wmexec. The race
condition occurred in the debian build log

https://buildd.debian.org/status/fetch.php?pkg=herbstluftwm&arch=ppc64el&ver=0.9.0-1~bpo10%2B1&stamp=1607007505&raw=0

and consists of the magic string "hlwm started" showing up to early such
that the read_and_echo_output() within hlwm.call() consumed it
and not the read_hlwm_output() in the test case.

As a fix, we use the unchecked_call() for which we can deactivate the
consumption of hlwm output.

I've successfully reproduced the race condition by updating the test
case to:

    @pytest.mark.parametrize('n', range(0, 200))
    def test_tags_restored_after_wmexec(hlwm, hlwm_process, n):

and passing read_hlwm_output=False fixes it.

### DIFF
--- a/tests/test_ewmh.py
+++ b/tests/test_ewmh.py
@@ -136,7 +136,9 @@ def test_tags_restored_after_wmexec(hlwm, hlwm_process):
     hlwm.create_client()
 
     # Restart hlwm:
-    hlwm.call(['wmexec', hlwm_process.bin_path, '--verbose'])
+    p = hlwm.unchecked_call(['wmexec', hlwm_process.bin_path, '--verbose'],
+                            read_hlwm_output=False)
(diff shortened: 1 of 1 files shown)


blarz:
Updated the commit message from
> that the read_and_echo_output() within hlwm.unchecked_call() consumed it

to

> that the read_and_echo_output() within hlwm.call() consumed it

blarz (approved):
Makes sense! Interesting that this has taken so long to be triggered

t-wissmann:
Thanks!

PR #1058, opened 2020-12-04
Add basic python bindings

t-wissmann:
This initiates python bindings to herbstluftwm. The purpose of this is
to make it easier to write python scripts for hlwm. Moreover, as soon as
there are more useful functions in the bindings, we can also make use of
them in the test suite. Thus, we don't need a dedicated testsuite for
the bindings.

So far, the only functionality is call(), unchecked_call(), and
argument parsing.

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -14,9 +14,7 @@ def tox(tox_args, build_dir):
     """
     tox_env = os.environ.copy()
     tox_env['PWD'] = build_dir
-    # here, we pass --recreate to tox to enforce that the package
-    # of the hlwm python bindings is updated in the virtual environment
-    sp.check_call(f'tox --recreate {tox_args}', shell=True, cwd=build_dir, env=tox_env)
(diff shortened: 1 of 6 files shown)


t-wissmann:
Also, I'm not sure what's the best name for the package. I went for `herbstluftwm` because it is more intuitive to use than `herbstclient`, even though the package only interacts with `herbstclient`.

blarz:
> Also, I'm not sure what's the best name for the package. I went for `herbstluftwm` because it is more intuitive to use than `herbstclient`, even though the package only interacts with `herbstclient`.

I think that makes sense!

blarz (approved):
Let's get this thing rolling!

What about testing the bindings?
I didn't insist on having it in this PR, because it is implicitly tested by the testsuite.
But we should think about explicitly (unit) testing the bindings when more functionality is added.
What do you think?

t-wissmann:
Thanks :) The new PR #1060 comes with test cases

PR #1060, opened 2020-12-05
Add AttributeProxy object for python bindings

t-wissmann:
This simplifies the reading and writing of attributes via the python
bindings. Also, this adds:

  - Test cases for thew new AttributeProxy functionality
  - example.py showing the example usage (and we call it from the test
    suite to ensure that the examples indeed work)
  - an utility function for combining commands via chain/and/or


blarz (approved):
Nice!

PR #1061, opened 2020-12-05
Do not call applyLayout() from FrameLeaf

### LINKED ISSUE #1056: Crash on split explode
Under some conditions the command "split explode" causes the wm to crash.

## How to reproduce
1. Start the wm (or don't have any window open)
2. Execute the default "Mod-Return" twice (`hc keybind $Mod-Return spawn "${terminal:-xterm}"`
3. Execute "split explode" (either with the default `$Mod-Control-space` or with `herbstclient split explode`)
4. Crash

It is important not to change focus or do anything else. If you change focus from one terminal to the other before step 3 it will not crash anymore.

## Crash output
    Dez 02 19:31:05 XaverPC systemd-coredump[18478]: Process 16962 (herbstluftwm) of user 1000 dumped core.

    Stack trace of thread 16962:
    #0  0x00007f9ae00959d5 raise (libc.so.6 + 0x3d9d5)
    #1  0x00007f9ae007e8a4 abort (libc.so.6 + 0x268a4)
    #2  0x000055c42d8d5bd8 _ZSt20__replacement_assertPKciS0_S0_ (herbstluftwm + 0x51bd8)
    #3  0x000055c42d913467 _ZN9FrameLeaf13computeLayoutE9Rectangle (herbstluftwm + 0x8f467)
    #4  0x000055c42d912498 _ZN10FrameSplit13computeLayoutE9Rectangle (herbstluftwm + 0x8e498)
    #5  0x000055c42d912498 _ZN10FrameSplit13computeLayoutE9Rectangle (herbstluftwm + 0x8e498)
    #6  0x000055c42d91e6f8 _ZN7Monitor11applyLayoutEv (herbstluftwm + 0x9a6f8)
    #7  0x000055c42d91978e _ZN9FrameLeaf5splitE10SplitAlign10FixPrecDecm (herbstluftwm + 0x9578e)
    #8  0x000055c42d8fb487 _ZN9FrameTree12splitCommandE5InputRSo (herbstluftwm + 0x77487)
    #9  0x000055c42d8c834f _ZNSt17_Function_handlerIFiR9FrameTree5InputRSoEMS0_FiS2_S3_EE9_M_invokeERKSt9_Any_dataS1_OS2_S3_ (herbstluftwm + 0x4434f)
    #10 0x000055c42d94fa84 _ZNSt17_Function_handlerIFi5InputRSoEZN10TagManager12frameCommandESt8functionIFiR9FrameTreeS0_S1_EEEUlS0_S1_E_E9_M_invokeERKSt9_Any_dataOS>
    #11 0x000055c42d8e4a5e _ZN8Commands4callE5InputRSo (herbstluftwm + 0x60a5e)
    #12 0x000055c42d90e75c _ZNK10KeyManager14handleKeyPressEP9XKeyEvent (herbstluftwm + 0x8a75c)
    #13 0x000055c42d8aeb17 main (herbstluftwm + 0x2ab17)
    #14 0x00007f9ae00801e2 __libc_start_main (libc.so.6 + 0x281e2)
    #15 0x000055c42d8c6c1e _start (herbstluftwm + 0x42c1e)

## Version
    herbstluftwm 0.9.0
    xrandr support on
    xinerama support on
    
    System Fedora 33

t-wissmann:
In FrameLeaf::split(), the function FrameLeaf::setSelection() was called before
`selection` was (and could be) set correctly. Since setSelection() called
Monitor::applyLayout(), this led to a crash because the `selection` was out of
bounds.

The present change removes all invokations of Monitor::applyLayout() from
setSelection() (the only leftover use of applyLayout() in FrameLeaf/FrameSplit).
To ensure that the actual window geometries update correctly, I went
through all invocations of setSelection and added the applyLayout() where it
wasn't called before (namely, the commands focus_nth and cycle). The other
commands (cycle_all, remove, focus) called applyLayout() already.

Also, I removed the call `clients[selection]->window_focus()` because it is
simply wrong if the FrameLeaf is not focused. Instead, the applyLayout()
function should take care of the focus.

Beside adding a test case for the crash scenario of #1056, I've extended the
test cases for focus_nth and cycle to verify that focus indeed updates
correctly.

This fixes #1056.

### DIFF
--- a/src/frametree.cpp
+++ b/src/frametree.cpp
@@ -152,6 +152,7 @@ int FrameTree::cycleSelectionCommand(Input input, Output output) {
     if (count != 0) {
         frame->setSelection(MOD(frame->getSelection() + delta, count));
     }
+    get_current_monitor()->applyLayout();
     return 0;
 }
 
@@ -162,6 +163,7 @@ int FrameTree::focusNthCommand(Input input, Output output) {
         return HERBST_NEED_MORE_ARGS;
     }
     focusedFrame()->setSelection(atoi(index.c_str()));
+    get_current_monitor()->applyLayout();
     return 0;
 }
 
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -468,8 +468,6 @@ void FrameLeaf::setSelection(int index) {
         index = clients.size() - 1;
     }
     selection = index;
-    clients[selection]->window_focus();
-    get_current_monitor()->applyLayout();
 }
 
 int Frame::splitsToRoot(SplitAlign align) {
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -304,14 +304,29 @@ def test_cycle(hlwm, running_clients, running_clients_num, num_splits, cycle_del
     new_windex = int(hlwm.get_attr('tags.0.curframe_windex'))
     expected_index = (windex + cycle_delta + wcount) % wcount if wcount > 0 else 0
     assert expected_index == new_windex
+    # check that the right winid is focused
+    layout = hlwm.call(['dump', '', '@']).stdout
+    winids = re.findall('0x[a-fA-f0-9]*', layout)
+    if len(winids) > 0:
+        assert winids[new_windex] == hlwm.get_attr('clients.focus.winid')
 
 
 @pytest.mark.parametrize("running_clients_num", [0, 1, 5])
 @pytest.mark.parametrize("index", [0, 1, 3, 5])
 def test_focus_nth(hlwm, running_clients, running_clients_num, index):
+    # bring the clients in the right order
+    layout = '(clients vertical:0 '
+    layout += ' '.join(running_clients)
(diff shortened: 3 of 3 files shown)


t-wissmann:
I also remove the `window_focus()` because it makes no sense here. Also I noticed that `explode` does not behave correctly regarding frame focus (see comment in the test case).

tiosgz:
> This fixes #1060.

Fixes a PR? Do you mean #1056?

t-wissmann:
> > This fixes #1060.
> 
> Fixes a PR? Do you mean #1056?

Right, thanks! I was too much in a hurry when sending this PR (I also should still check some other things, so it will stay 'wip' for a bit)

t-wissmann:
I've also added tests for `cycle` and `focus_nth` that ensure that the `clients.focus` object is updated correctly. So I think this is ready to go now.

blarz (approved):

PR #1062, opened 2020-12-06
Enable stack protection compiler flags in Debug build

blarz:
Inspired by the flags used by Fedora in issue #1056

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -61,8 +61,8 @@ build_env = os.environ.copy()
 build_env.update({
     'CC': args.cc,
     'CXX': args.cxx,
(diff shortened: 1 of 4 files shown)


t-wissmann:
The `clients` array lives in the heap, not the stack, so it's probably something else. If I find more time, I can bisect the flags locally to see which flag triggers it.

It's definitely a good idea to add it to the ci :-)

blarz:
> The `clients` array lives in the heap, not the stack, so it's probably something else. If I find more time, I can bisect the flags locally to see which flag triggers it.

Alright, thank you :)

> It's definitely a good idea to add it to the ci :-)

I added it to the debugflags.cmake. If you don't like it, I can restrict it to CI only.

blarz:
Will be taken care of in #1063

PR #1063, opened 2020-12-07
Use -D_GLIBCXX_ASSERTIONS and other security flags

t-wissmann:
In the debug build, activate range checks for C++ datatypes. This flag
would have been prevented crash #1056 on fedora, that has these flags
per default. While at it, I've also added further stack protection flags
in the hope that this helps us to detect bugs earlier.

### DIFF
--- a/cmake/debugflags.cmake
+++ b/cmake/debugflags.cmake
@@ -18,11 +18,6 @@ set(FLAGS
     # range checks for C++ datatypes:
     -D_GLIBCXX_ASSERTIONS
 
-    # more stack protection:
-    -fstack-protector-strong
-    -fcf-protection
-    -fstack-clash-protection
-
     # TODO: Fix complaints, then enable -Wconversion (gcc and clang might behave
     # differently)
 
@@ -36,6 +31,13 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
         -Wduplicated-cond -Wduplicated-branches -Wlogical-op
 
         # TODO: Consider adding -fanalyzer here (new in GCC-10)
+        
+        
+
+        # more stack protection:
+        -fstack-protector-strong
+        -fcf-protection
+        -fstack-clash-protection
         )
 
     # Add GCC- *and* C++-specific flags
--- a/cmake/debugflags.cmake
+++ b/cmake/debugflags.cmake
@@ -15,6 +15,14 @@ set(FLAGS
     # Disable unused parameter warnings (most cases are on purpose):
     -Wno-unused-parameter
 
+    # range checks for C++ datatypes:
+    -D_GLIBCXX_ASSERTIONS
+
+    # more stack protection:
+    -fstack-protector-strong
+    -fcf-protection
+    -fstack-clash-protection
+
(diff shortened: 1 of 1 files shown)


t-wissmann:
The fedora flags are using `-Wp,-D_GLIBCXX_ASSERTIONS` instead of `-D_GLIBCXX_ASSERTIONS` and I don't know what the difference/relevance is. With `-D_GLIBCXX_ASSERTIONS` I can reproduce the crash of #1056 (and there's a similar story for this [project on github](https://github.com/psi4/psi4/issues/1255)). Fedora lists their [build flags here](https://fedoraproject.org/wiki/Changes/HardeningFlags28).

sham1:
> The fedora flags are using `-Wp,-D_GLIBCXX_ASSERTIONS` instead of `-D_GLIBCXX_ASSERTIONS` and I don't know what the difference/relevance is.

`-Wp` just passes that flag to the preprocessor and bypasses the compiler.

t-wissmann:
> `-Wp` just passes that flag to the preprocessor and bypasses the compiler. Because the compiler could change and interpret the flags by itself, it's just probably for more reliability.

Makes sense, thanks!

cmake/debugflags.cmake
```diff
@@ -15,6 +15,14 @@ set(FLAGS
     # Disable unused parameter warnings (most cases are on purpose):
     -Wno-unused-parameter
 
+    # range checks for C++ datatypes:
+    -D_GLIBCXX_ASSERTIONS
+
+    # more stack protection:
+    -fstack-protector-strong
+    -fcf-protection
+    -fstack-clash-protection
```
blarz, line 24: I think that flag is only available with gcc, that is why I put it further down in this file #1062
(But Travis seems to be asleep right now)

blarz (changes requested):

cmake/debugflags.cmake
t-wissmann: Ah, I see, thanks! I've moved the flags to the GNU-section

blarz: -fstack-protector-strong is available in both gcc and clang, only fstack-clash-protection is gcc only.
Not sure about cf-protection.

Sorry, should've been more specific

blarz:
Should we also add FORTIFY_SOURCE?

cmake/debugflags.cmake
t-wissmann: I think it's fine if we stay safe here (not sure whether there are any c++ compilers besider clang and gcc that possibly don't support it..).

t-wissmann:
> Should we also add FORTIFY_SOURCE?

I don't have any experience with it; if you can recommend it and if it does not hide any crashs, we can add it.

blarz:
> > Should we also add FORTIFY_SOURCE?
> 
> I don't have any experience with it; if you can recommend it and if it does not hide any crashs, we can add it.

I don't have any firsthand experience with it either. So maybe let's stick with the flags you introduce here for now.

cmake/debugflags.cmake
blarz: Alright!

blarz (approved):

PR #1064, opened 2020-12-07
Make 'split explode' preserve the client focus

t-wissmann:
This restores the (desirable) behaviour of 'split explode' from v0.7.2
where splitting preserved which client is focused.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -395,9 +395,9 @@ split 'ALIGN' ['FRACTION' ['FRAMEINDEX']::
     after the split. The other half will be occupied by the currently focused
     frame. After splitting, the originally focused frame will stay focused.
     One special 'ALIGN' mode is 'explode', which splits the frame in such a way
-    that the window sizes and positions are kept as much as possible. If no
-    'FRACTION' is given to 'explode' mode an optimal fraction is picked
-    automatically. Example:
+    that the window focus, window sizes, and positions are kept as much as
+    possible. If no 'FRACTION' is given to 'explode' mode an optimal fraction is
+    picked automatically. Example:
 
         * split explode
         * split bottom 0.5
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,8 @@ Current git version
     - When hiding windows, correctly set their WM_STATE to IconicState (we set
       it to Withdrawn state before, which means "unmanaged" and thus is wrong).
       This may require restarting pagers when upgrading hlwm live.
+    - Fix crash in 'split explode'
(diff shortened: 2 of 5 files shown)


doc/herbstluftwm.txt
```diff
@@ -395,9 +395,9 @@ split 'ALIGN' ['FRACTION' ['FRAMEINDEX']::
     after the split. The other half will be occupied by the currently focused
     frame. After splitting, the originally focused frame will stay focused.
     One special 'ALIGN' mode is 'explode', which splits the frame in such a way
-    that the window sizes and positions are kept as much as possible. If no
-    'FRACTION' is given to 'explode' mode an optimal fraction is picked
-    automatically. Example:
+    that the window focus, window sizes, and positions are kept as much as
+    possible. If no 'FRACTION' is given to 'explode' mode an optimal fraction is
```
tiosgz, line 399: I know this isn’t in the logical scope of this PR, but… it doesn’t make sense. (& it’s been here for a [really long time](https://github.com/herbstluftwm/herbstluftwm/commit/59ce5d823e3204c7e499601ad32e8769aab0647b))

The `to 'explode' mode` seems to be completely wrong there.

But IMO it also needs an explanation of what’s ‘optimal fraction,’ so I’ll probably fix it in a later PR (unless you want to do it now).

blarz: That would be great!

You mean the word "mode" isn't correct  here?

tiosgz: Whoops, just reread it & realised that there’s also `ALIGN` there (it seemed really odd that it was called `FRACTION`, but somehow it didn’t hit me the first time; hence I completely misunderstood it before.

I’ll have to have another look at it & then let you know what I actually meant & if (& what) there’s something wrong.

Anyway don’t let it block the PR, this is something that has nothing to do with it.

blarz: Alright, thank you!

blarz (approved):
Looks good to me!

doc/herbstluftwm.txt
t-wissmann: If you misread something first, then this is an indicator that something should be clarified or rephrased in the docs :-) If you have an idea, it would be great if you send a PR :-)

tiosgz: I guess the part that I've misread is fine. I was just quickly looking at what are the arguments named, & so I only looked at what's between the brackets.

Now that I've given it more time than a few seconds, I found out that it's correct (well, there's missing a bracket, but that's not what I was talking about at first).

PR #1067, opened 2020-12-12
Migrate CI to github actions

t-wissmann:
This converts the travis ci configuration to github actions. Again we
have three jobs: one with gcc, one with clang, both on a current 64 bit
system, and one on an old 32 bit system.

All three jobs run in docker containers that are pulled from dockerhub.
In the old travis setup, if the docker file changed, then the newer
docker file was used directly. This is not yet reflected in the github
actions, so we can't use pull requests to test changes in the docker
files, yet.

Moreover, the ci/build.py is really optimized towards being invoked only
once, as it was the case for the travis ci. With github actions, we will
need to invoke it multiple times per job (namely step by step), so
ci/build.py will need to change further in the future. I tried to be as
little invasive here as possible to keep the diff of ci/build.py small.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -1,51 +1,23 @@
 name: HLWM CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build-test-current:
     # Name the Job
     name: Build and test on current ubuntu
     # Set the type of machine to run on
-    runs-on: ubuntu-latest
+    container: hlwm/ci:focal
 
     steps:
       # Checks out a copy of your repository on the ubuntu-latest machine
       - name: Checkout code
         uses: actions/checkout@v2
 
-      - name: Install dependencies
-        run: |
-          sudo apt-get update
-          DEBIAN_FRONTEND=noninteractive sudo apt-get -y install \
-            ccache \
-            clang-10 \
-            clang-tidy-10 \
-            cmake \
-            g++-9 \
-            iwyu \
-            lcov \
-            libx11-dev \
-            libxext-dev \
-            libxinerama-dev \
-            libxrandr-dev \
-            ninja-build \
-            pkg-config \
-            python3.8 \
-            tox \
-            x11-xserver-utils \
-            xdotool \
-            xterm \
-            xvfb \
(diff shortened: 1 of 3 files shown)


.github/workflows/hlwm.yml
```diff
@@ -0,0 +1,111 @@
...
+          ci/build.py --build-dir=build --cmake --cxx=g++-9 --cc=gcc-9 --build-type=Debug --ccache=$HOME/.ccache --compile
+
+      - name: Test
+        run: |
+          ci/build.py --build-dir=build --run-tests
+
+      - name: Codecov report
+        continue-on-error: true
+        run: |
+          wget -O codecov-io.bash https://codecov.io/bash
```
blarz, line 46: This doesn't seem to work anymore:

  wget -O codecov-io.bash https://codecov.io/bash
  bash codecov-io.bash -f coverage.info
  shell: sh -e {0}
/__w/_temp/030faa66-a8fb-4057-a380-e8962ace7133.sh: 1: wget: not found
Error: Process completed with exit code 127.

blarz (changes requested):

t-wissmann:
Since this PR changes the mergify config, one has to manually merge this PR. This means that one needs to manually copy and paste the PR description to the commit message...

.github/workflows/hlwm.yml
t-wissmann: It seems wget is missing in our docker image. The best fix I can think of is to add it to the docker image (later?) because I don't see any other way to download the script.

blarz: Didn't we use curl in the past? Maybe that is installed in the Docker image already

t-wissmann: Looking at the Dockerfile, there is no indicator for me that curl is installed. In the past, that line was executed outside of docker, namely on the system provided by travis. I'm sure that the system provided by github actions also has wget or curl installed, but I don't know how to perform a single step of a job outside the `container` specified.

blarz: True. Let's fix that afterwards.

blarz (approved):
Very nice!

PR #1069, opened 2020-12-13
Rephrase docs for default 'split explode' FRACTION

tiosgz:
It confused me a little.


t-wissmann (approved):
Thanks, I think it's better!

PR #1072, opened 2020-12-14
Compare compiler binaries for ccache in ci

t-wissmann:
The ccache in the ci has a lot of cache misses. Maybe this comes from
the mtime of the compiler being different on each CI run. This now
compares the actualy compiler binary instead of the compiler's mtime.
Hopefully, this improves the cache hit rate.

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -50,7 +50,10 @@ if args.ccache:
     if conf.exists():
         conf.unlink()
 
-    # Set a reasonable size limit, and compare compiler based on content, not mtime
+    # Set a reasonable size limit, and compare compiler based on content, not mtime.
+    # Hash-verifying the compiler is required when building with
+    # clang-and-tidy.sh (because the script's mtime is not stable) and for
+    # other cases, the overhead is minimal):
     sp.check_call('ccache --max-size=500M -o compiler_check=content', shell=True)
 
     # Wipe stats before build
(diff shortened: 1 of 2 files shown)


ci/build.py
```diff
@@ -50,8 +50,8 @@ def tox(tox_args, build_dir):
     if conf.exists():
         conf.unlink()
 
-    # Set a reasonable size limit
-    sp.check_call('ccache --max-size=500M', shell=True)
+    # Set a reasonable size limit, and compare compiler based on content, not mtime
+    sp.check_call('ccache --max-size=500M -o compiler_check=content', shell=True)
```
dnnr, line 54: Don't we already enable that in line 73? That said, the code seems to have changed a bit since I last checked... this args.cmake is new to me.

t-wissmann: Oh right! I have only looked at the output of `ccache -p` in the ci (which is before line 73) and noticed this setting. Do you have an idea what else could cause the 100% cache miss rate?

dnnr: Without having looked at it more closely, I'd suggest trying to disable the hash_dir option. This is a common culprit in environments where every build happens in a randomized path. No idea if that applies to GHA, though.

t-wissmann: According to the log, the directory is always `/__w/herbstluftwm/herbstluftwm/build` (not sure whether the log is de-randomized). I've just set `hash_dir=false`, so let us see what happens.

PR #1073, opened 2020-12-15
Laxer ccache parameters

t-wissmann:
According to the GHA docs[1] the cache lookup "first searches for cache
hits for key and restore-keys in the branch containing the workflow
run."

However, it seems that in #1072, it always picks the cache from the
master branch. Hence, in this PR, I'm trying to adjust the ccache
parameters accordingly to make the cache in the 'master' branch as good
as possible.

[1] https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -41,6 +41,11 @@ jobs:
           path: .tox-cache
           key: focal-tox
 
+      - name: inspect ccache 1
+        run: |
+          find ~/.ccache | grep 8/f/64 || true
+          find ~/.ccache | wc -l
+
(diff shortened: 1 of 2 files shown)


t-wissmann:
I'm adding the debug output to the CI of the master branch in the hope to get more insights what's going on, and why the ccache is not restored correctly.

t-wissmann:
Now I understand: the cache is never ever updated. Once a cache entry is stored under a given key, it stays there and is never updated! So we need to alter the lookup keys smartly to get the behaviour we need.

PR #1074, opened 2020-12-15
Print parse errors for apply rule commands

t-wissmann:
The commands apply_tmp_rule and apply_rules now print parser error
messages to their output which is forwarded to the invoking
herbstclient. This makes debugging rules easier.

However, on new clients, the error message is still printed to
herbstluftwm's stderr.


t-wissmann:
Now I understand: the cache is *never ever* updated. Once a cache entry is stored under a given key, it stays there and is never updated! So we need to alter the lookup keys smartly to get the behaviour we need.

blarz (approved):
Makes sense!

PR #1076, opened 2020-12-19
Rename RootCommands to MetaCommands

t-wissmann:
MetaCommands is the more suitable name, because it indicates that is
very generic in the concrete object and command structure of the actual
herbstluftwm system.

Since I plan to create another class containing commands, the present
renaming is beneficial, because I think it will avoid name confusion.
Moreover, this name is not exposed to the user, so there are no
compatibility issues.


blarz (approved):
Looks good to me, couldn't find any leftovers to be renamed.

PR #1077, opened 2020-12-19
Add test cases for the 'shift' command

t-wissmann:
This adds tests for the functionality of the 'shift' command and also
clarifies the error message if no client is focused (the message was "No
neighbour found" before).


blarz (approved):
More tests, always good! Thanks!

Issue #565, opened 2019-07-21
Migrate layout commands from layout.cpp to frametree.cpp


t-wissmann:
This regards the commands

  - [x] "bring" `frame_current_bring`
  - [x] "split" `frame_split_command`
  - [x] "resize" `frame_change_fraction_command`
  - [ ] "focus_edge" `frame_focus_edge`
  - [x] "focus" `frame_focus_command`
  - [ ] "shift_edge" `frame_move_window_edge`
  - [x] "shift" `frame_move_window_command`
  - [x] "set_layout" `frame_current_set_client_layout`

PR #1078, opened 2020-12-19
Migrate the 'shift' command

t-wissmann:
- Migrate it to the new ArgParse and Completion API
  - The rest of the implementation is nearly unchanged (`frame` is renamed to
    `sourceFrame` and the assignment in the if condition is avoided).
  - Remove old implementation and completion
  - We need a slight change in frame_move_window_edge, analogously
    to frame_focus_edge in a1de01fba (#685) where 'focus' was migrated.
  - The test-case for focus_edge is reused to test 'shift_edge'
  - A leftover function declaration from layout.h is removed

This fixes one item in the list of #565.


blarz (approved):
Looks good to me, also tested locally

PR #1079, opened 2020-12-20
Only build json doc on WITH_DOCUMENTATION

t-wissmann:
Since tarballs contain the json doc anyway treat the json doc as the
other man pages and only build them if WITH_DOCUMENTATION is activated.
This avoids python as a build dependency for the tarballs.

If WITH_DOCUMENTATION is not set, then we still create an optional make
target which is called in the CI.

Also this marks the installation of the json doc optional, meaning that
it is installed if it has been generated.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -158,6 +158,6 @@ jobs:
           cd build
           cmake -GNinja -DWITH_DOCUMENTATION=NO -DENABLE_CCACHE=YES ..
           ninja -v -k10
-          ninja -v doc_json
+          ninja -v -t doc_json
           ccache -s
           find ~/.ccache | wc -l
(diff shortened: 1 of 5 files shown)


t-wissmann:
I'm not yet sure whether the documentation is installed correctly if `WITH_DOCUMENTATION=off`; because in that case the built documentation is still in the source directory and not in the build directory.

t-wissmann:
> I'm not yet sure whether the documentation is installed correctly if `WITH_DOCUMENTATION=off`; because in that case the built documentation is still in the source directory and not in the build directory.

With this I meant that running `make install` will not install the documentation if `WITH_DOCUMENTATION=off`. Also, since the installation of the json doc is now optional, installing does not fail for  `WITH_DOCUMENTATION=off` :-)

blarz (approved):
Looks good to me and makes sense for all the packagers out there

PR #1081, opened 2020-12-22
Fix typo of the word 'writable'

t-wissmann:
Since `writable` is the more standard spelling (and I'm unintentionally writing
it anyway) change everything to `writable`:

    sed -i 's,\(writ\)e\(able\),\1\2,ig' `git ls-files|grep -v ^ci/` 

(the `grep -v` is only to ignore the `ci` directory, which contains a symbolic
link which would get destroyed by `sed -i`). Grep proves that we caught all
occurrences:

    grep -i writeable `git ls-files`

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -158,5 +158,6 @@ jobs:
           cd build
           cmake -GNinja -DWITH_DOCUMENTATION=NO -DENABLE_CCACHE=YES ..
           ninja -v -k10
+          ninja -v doc_json
           ccache -s
           find ~/.ccache | wc -l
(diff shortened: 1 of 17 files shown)


blarz:
As far as I can see both forms are fine (according to various dictionaries).
My browser's spell checker seems to prefer writable over writeable :D

t-wissmann:
Oh indeed... it seems like `writeable` is the alternative spelling of `writable`.

t-wissmann:
We've always used `writeable` so far (in the man page and the output of the `attr` command). Should we change it to `writable`? I think we should make it consistent and we should make a decision before #1082 is merged :)

tiosgz:
If I may help you decide…
* you always use *-ize* (US)
* in programming are, AFAIK, preferred US forms & words
* according to [Wiki](https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#Dropped_%22e%22), *writeable* is rather UK English (& now that I’m writing it, it seems that my browser’s spell checker doesn’t know it even though I do have BrE dictionary)

t-wissmann:
Right, so I should actually make the inverse pull request where  I replace `writeable` by `writable`.

blarz (approved):
writable seems to make the most sense

PR #1082, opened 2020-12-22
json doc: Derive writable property from source code

t-wissmann:
This extends the doc/gendoc.py by a parser for the writable flag of
attributes. The parser looks for constructor arguments and invokations
of Attribute::setWritable(). Unfortunately, some of the latter were
indirect in loops, so I had to explicitly call setWritable() for every
attribute (happens in Client and Monitor). Since all attributes of the
Settings object are writable, I've hardcoded this in the gendoc.py.

Fortunately, we can be sure that we didn't miss a attribute, because the
tests in test_doc verify that the `writable` field in the json doc is
correct.


t-wissmann:
I have merged #1081 into this branch, so #1081 must be merged first.

blarz:
Updated a writeable in the commit message to writable ;)
There seem to be conflicts still with this PR

blarz (approved):
Please resolve the conflict, then this should be good to go

t-wissmann:
Thanks! This seems to be the main/only downside of the pr-squashing: one has to merge the same conflicts multiple times if one PR includes the other.

PR #1083, opened 2020-12-22
Document inheritance in json doc

t-wissmann:
The class inheritance is relevant for the json doc, because the
FrameTree has children of type `Frame`, and a `Frame` can be a
`FrameSplit` or a `FrameLeaf`.

I have no explicit test for this, but of course the test suite already
tests for the existence of the inherited attributes.


blarz (approved):
Looks good to me!

PR #1084, opened 2020-12-24
diff-json-doc: Collapse longer diffs

t-wissmann:
This puts the json diff in the PR comments into a collapsible block. The
diff is shown per default for diffs shorter than 100 lines and is
collapsed per default otherwise.


t-wissmann:
The improvement has already been applied to the diff in #1082 and #1083

PR #1085, opened 2020-12-24
diff-json-doc: Normalize trailing commas

t-wissmann:
Add trailing commas after fields, and not only between fields. This reduces the
diff size whenever a new field is added at the end of an object.


t-wissmann:
The improvement has already been applied to the diff in #1082 and #1083.

PR #1088, opened 2020-12-24
Add 'parent_frame' entry for each client

### LINKED ISSUE #6: object tree: frame id (or object) as attribute of client
People sometimes ask how they get the correct frame ID for a rule.

Having a frame id (or with frame object, the object) as attribute/child of a client object might make this easier.

t-wissmann:
The 'parent_frame' child of a client provides the frame object that
holds the client.

This fixes #6.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,7 +12,7 @@ Current git version
     order of existing tags.
   * New child object 'focused_client' for each tag object.
   * New child object 'focused_frame' for the tiling object of each tag object.
-  * New child object 'frame' for each client providing the frame the client sits in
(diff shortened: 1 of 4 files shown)


t-wissmann:
I'm not sure whether `frame` is a good name. Maybe something like `owning_frame` or `parent_frame` is better.

blarz:
> I'm not sure whether `frame` is a good name. Maybe something like `owning_frame` or `parent_frame` is better.

parent_frame sounds nice IMHO

t-wissmann:
Alright! I've renamed it to `parent_frame`.

blarz (approved):
Nice!

PR #1090, opened 2020-12-28
Basic window title drawing

t-wissmann:
The title of every client is drawn on the window decoration, if
title_height > 0. The drawing routine XDrawString does not support utf8
characters and only basic X fonts (no Xft). Both issues can be fixed in
the future.

This adds the new attribute type 'Font' and introduces new theme
attributes: title_font, title_height, title_color.

It is quite hard to add automated tests for it. Hence, for the time
being, there are only tests for the new font types.


NEWS
```diff
@@ -8,6 +8,8 @@ Release 0.9.1 on 2020-12-28
 ---------------------------
 
   * Client minimization (controlled by the attribute 'minimized' of every client)
+  * Client window titles (controlled by the theme attributes 'title_height',
```
blarz, line 11: Should probably open a new "Current git version" section for this

blarz (changes requested):

NEWS
t-wissmann: Right, Thanks! I've fixed it.

blarz:
> The title of every client is drawn on the window decoration, if
> title_height > 0. The drawing routine XDrawString does not support utf8
> characters and only basic X fonts (no Xft). Both issues can be fixed in
> the future.

This looks very cool, thank you!

> It is quite hard to add automated tests for it. Hence, for the time
> being, there are only tests for the new font types.

How could we test it? Can we check for a text property within the padding area or something?

The padding is obviously not displayed if `smart_window_surroundings` is enabled,
but that is probably what we want, right?

NEWS
blarz: Thank you!

t-wissmann:
> > It is quite hard to add automated tests for it. Hence, for the time
> > being, there are only tests for the new font types.
> 
> How could we test it? Can we check for a text property within the padding area or something?

The best I can think of is making a screenshot and counting pixels. (E.g. if `title_color` is red, then the red pixel count should increase when changing the window title from `x` to `X`)

> The padding is obviously not displayed if `smart_window_surroundings` is enabled,
> but that is probably what we want, right?

I didn't think about  `smart_window_surroundings` at all. Maybe it's even sensible to hide the title in the smart surroundings case

blarz:
> > > It is quite hard to add automated tests for it. Hence, for the time
> > > being, there are only tests for the new font types.
> > 
> > 
> > How could we test it? Can we check for a text property within the padding area or something?
> 
> The best I can think of is making a screenshot and counting pixels. (E.g. if `title_color` is red, then the red pixel count should increase when changing the window title from `x` to `X`)

That sounds cumbersome.. I hope we can find another way.

> > The padding is obviously not displayed if `smart_window_surroundings` is enabled,
> > but that is probably what we want, right?
> 
> I didn't think about `smart_window_surroundings` at all. Maybe it's even sensible to hide the title in the smart surroundings case

Yes, I think so, too.

blarz (approved):
Nice! Let's get this in and see how it progresses :)

PR #1091, opened 2020-12-28
Add xft font support for window titles

t-wissmann:
Use xft fonts in the window title rendering per default and only fall
back to plain x fonts if a XLFD is given explicitly.

Though xft supports utf8 per default, it does not seem to provide a font
fallback system if a particular unicode character is not contained in
the font selected. Hence, the best compromise here is to pick a font
supporting as many unicode characters as possible, e.g. Dejavu Sans.

This introduces a new dependencies on libxft and freetype which should
be standard on nowadays systems with X11.


t-wissmann:
I don't get why the ancient ubuntu image again lacks the `freetype` library...

blarz:
> I don't get why the ancient ubuntu image again lacks the `freetype` library...

The cmake output looks more like pkg_config stuff doesn't find anything in general?
Or am I reading it wrong?

t-wissmann:
> > I don't get why the ancient ubuntu image again lacks the `freetype` library...
> 
> The cmake output looks more like pkg_config stuff doesn't find anything in general?
> Or am I reading it wrong?

Oh right.. Do you have an idea why?

blarz:
> > > I don't get why the ancient ubuntu image again lacks the `freetype` library...
> > 
> > 
> > The cmake output looks more like pkg_config stuff doesn't find anything in general?
> > Or am I reading it wrong?
> 
> Oh right.. Do you have an idea why?

Not sure, it seems to work fine for the newer container..

Maybe pkg-config has trouble finding the i386 libs?

blarz:
> > > I don't get why the ancient ubuntu image again lacks the `freetype` library...
> > 
> > 
> > The cmake output looks more like pkg_config stuff doesn't find anything in general?
> > Or am I reading it wrong?
> 
> Oh right.. Do you have an idea why?

Can you try this:

```
$ git diff
diff --git a/ci/Dockerfile.ci-trusty b/ci/Dockerfile.ci-trusty
index bef7ac00..a1f7ce4d 100644
--- a/ci/Dockerfile.ci-trusty
+++ b/ci/Dockerfile.ci-trusty
@@ -17,7 +17,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install \
     libxml2-utils \
     libxrandr-dev:i386 \
     ninja-build \
-    pkg-config \
+    pkg-config:i386 \
     xterm \
     xvfb \
     && rm -rf /var/lib/apt/lists/*
```
I played around with it locally and it seems to resolve finding stuff with a -m32 build.

blarz:
And if that doesn't help, maybe PKG_CONFIG_PATH='/usr/lib/i386-linux-gnu/pkgconfig' in the `env` of the `build-old-32bit`

t-wissmann:
> > > > I don't get why the ancient ubuntu image again lacks the `freetype` library...
> > > 
> > > 
> > > The cmake output looks more like pkg_config stuff doesn't find anything in general?
> > > Or am I reading it wrong?
> > 
> > 
> > Oh right.. Do you have an idea why?
> 
> Can you try this:
> 
> ```
> $ git diff
> diff --git a/ci/Dockerfile.ci-trusty b/ci/Dockerfile.ci-trusty
> index bef7ac00..a1f7ce4d 100644
> --- a/ci/Dockerfile.ci-trusty
> +++ b/ci/Dockerfile.ci-trusty
> @@ -17,7 +17,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install \
>      libxml2-utils \
>      libxrandr-dev:i386 \
>      ninja-build \
> -    pkg-config \
> +    pkg-config:i386 \
>      xterm \
>      xvfb \
>      && rm -rf /var/lib/apt/lists/*
> ```
> 
> I played around with it locally and it seems to resolve finding stuff with a -m32 build.

Sounds promising! Can you simply commit it or send it as a PR? (the dockerhub pulls the dockerfile from the master branch of the herbstluftwm repository)

blarz:
> Sounds promising! Can you simply commit it or send it as a PR? (the dockerhub pulls the dockerfile from the master branch of the herbstluftwm repository)

Sure!

blarz:
Opened #1103 as I cannot push to master of this repo directly (which is totally fine)

t-wissmann:
> Opened #1103 as I cannot push to master of this repo directly (which is totally fine)

Thanks for investigating this! The build on https://hub.docker.com/repository/docker/hlwm/ci/builds has already been triggered.

blarz:
Nice, seems to work with trusty now. Not sure if the leak detected by LSAN is an actual leak or if we need to add to the suppression list.

t-wissmann:
> Nice, seems to work with trusty now.

Yaiy! :-)

>  Not sure if the leak detected by LSAN is an actual leak or if we need to add to the suppression list.

This really puzzles me, because even though the build type is "Debug", the back trace regarding the allocation of the lost bytes does not touch hlwm's source code: 

```
Direct leak of 320 byte(s) in 320 object(s) allocated from:
    #0 0x7f7ba4499bc8 in malloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dbc8)
    #1 0x7f7ba32d8ddd  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x17ddd)

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x7f7ba4499bc8 in malloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dbc8)
    #1 0x7f7ba32e22f4  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x212f4)

Indirect leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f7ba4499dc6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10ddc6)
    #1 0x7f7ba32e293c  (/lib/x86_64-linux-gnu/libfontconfig.so.1+0x2193c)

SUMMARY: AddressSanitizer: 608 byte(s) leaked in 322 allocation(s).
```

blarz:
> This really puzzles me, because even though the build type is "Debug", the back trace regarding the allocation of the lost bytes does not touch hlwm's source code:

If you're sure that this isn't a problem on hlwm's side, you can add libfontconfig to `ci/lsan-suppressions.txt`.
We already have an entry for leaking xrandr ;)

blarz:
> > This really puzzles me, because even though the build type is "Debug", the back trace regarding the allocation of the lost bytes does not touch hlwm's source code:
> 
> If you're sure that this isn't a problem on hlwm's side, you can add libfontconfig to `ci/lsan-suppressions.txt`.
> We already have an entry for leaking xrandr ;)

Googling around seems to bring up the same reasons why we suppressed xrandr leaks.
libfontconfig just does, so [projects like chromium suppress it, too](https://chromium.googlesource.com/chromium/src/build/+/master/sanitizers/lsan_suppressions.cc#22).

I totally forgot, but I already searched for this problem some time ago because of a Qt program where LSAN reported this and [Reddit also said "this is known"...](https://www.reddit.com/r/cpp/comments/41vm7r/is_clangs_leak_sanitizer_inaccurate_or_is_qt/)

t-wissmann:
> > > This really puzzles me, because even though the build type is "Debug", the back trace regarding the allocation of the lost bytes does not touch hlwm's source code:
> > 
> > 
> > If you're sure that this isn't a problem on hlwm's side, you can add libfontconfig to `ci/lsan-suppressions.txt`.
> > We already have an entry for leaking xrandr ;)
> 
> Googling around seems to bring up the same reasons why we suppressed xrandr leaks.
> libfontconfig just does, so [projects like chromium suppress it, too](https://chromium.googlesource.com/chromium/src/build/+/master/sanitizers/lsan_suppressions.cc#22).
> 
> I totally forgot, but I already searched for this problem some time ago because of a Qt program where LSAN reported this and [Reddit also said "this is known"...](https://www.reddit.com/r/cpp/comments/41vm7r/is_clangs_leak_sanitizer_inaccurate_or_is_qt/)

Thanks a lot! That's a relief; I was really frustrated with the CI for this PR. I've added the lsan-suppressions and we can now start to talk about the actual content :-)

The main change worth discussing is that I've replaced the cmake-specific X11-detection-part by multiple `pkg-config` calls. I feel more comfortable using the (I think) more standard `pkg-config`. What do you think? I can also move the pkg-config migration to another PR (I'll take a look at it tomorrow).

share/autostart
```diff
@@ -121,7 +121,10 @@ hc set frame_transparent_width 5
 hc set frame_gap 4
 
 hc attr theme.title_height 15
-hc attr theme.title_font '-*-fixed-medium-r-*-*-13-*-*-*-*-*-*-*'
```
blarz, line 124: Do we want to keep of all of those added lines?

src/CMakeLists.txt
```diff
@@ -85,14 +85,28 @@ cmake_dependent_option(WITH_XINERAMA "Use multi-monitor support via xinerama" ON
...
 ## dependencies X11 (link to Xext for XShape())
 target_include_directories(herbstluftwm SYSTEM PUBLIC
-    ${X11_X11_INCLUDE_PATH} ${X11_Xinerama_INCLUDE_PATH} ${X11_Xrandr_INCLUDE_PATH})
+    ${FREETYPE_INCLUDE_DIRS}
+    ${X11_INCLUDE_DIRS}
+    ${XFT_INCLUDE_DIRS}
+    ${XEXT_INCLUDE_DIRS}
+    ${XINERAMA_INCLUDE_DIRS}
+    ${XRANDR_INCLUDE_DIRS}
+    # ${X11_Xinerama_INCLUDE_PATH}
```
blarz, line 99: Can be removed? (Same for the line below)

INSTALL
```diff
@@ -18,6 +18,7 @@ Runtime dependencies:
 
     - bash (if you use the default autostart file)
     - libx11
+    - xft and freetype
```
blarz, line 21: Should we add the new requirements to NEWS, too? It is kind of "user visible" and will probably make the life of package maintainers easier

ci/lsan-suppressions.txt
```diff
@@ -1,2 +1,8 @@
 # Xrandr is leaking...
 leak:libXrandr
+
+# libfontconfig too according to
+# https://chromium.googlesource.com/chromium/src/build/+/master/sanitizers/lsan_suppressions.cc#22
+# http://crbug.com/39050
+# https://www.reddit.com/r/cpp/comments/41vm7r/is_clangs_leak_sanitizer_inaccurate_or_is_qt/
+leak:libfontconfig
```
blarz, line 8: Still seems to make problems with the CI, can you try `leak:libfontconfig.so.1` and see if that fixes it?

blarz (changes requested):

blarz:
> Thanks a lot! That's a relief; I was really frustrated with the CI for this PR. I've added the lsan-suppressions and we can now start to talk about the actual content :-)

I can imagine that this is getting on your nerves... I hope we can fix this soon.
It still breaks though. I pushed a commit testing suppression with the full lib name, not sure if it helps.

> The main change worth discussing is that I've replaced the cmake-specific X11-detection-part by multiple `pkg-config` calls. I feel more comfortable using the (I think) more standard `pkg-config`. What do you think? I can also move the pkg-config migration to another PR (I'll take a look at it tomorrow).

I am pretty indifferent to the pkg-config changes. Not sure what the preferred way for this is "in cmake-world"?
It feels a bit weird, as I thought cmake is the tool that takes care of this? Wouldn't it also work by adding `find_package(Foo REQUIRED)` directives instead of `pkg_check_modules(FOO foo)` you added? What was it that made you choose?

I'd appreciate a separate PR, maybe explaining the reasoning behind it. But it is not really that import to me :)

ipc-client/CMakeLists.txt
```diff
@@ -17,9 +17,9 @@ set_target_properties(herbstclient PROPERTIES
```
blarz, line 17: Should the POSIC read POSIX a couple of lines above?

blarz (changes requested):

ci/lsan-suppressions.txt
blarz: The lsan-suppression is fine, it is just that one test is failing, because it doesn't get the LSAN_OPTIONS env passed.
I'll add the fix here.

t-wissmann: Great, thanks!

ipc-client/CMakeLists.txt
t-wissmann: Yes, I've fixed it in 26363fd5935d82dd2d3225f65ed05e40205113d7 and pushed it directly to master.

share/autostart
t-wissmann: For me, it is always quite hard to get the format of font specifications right because there are strange subtleties in every application. That's why I thought it might be helpful to see a few examples. But maybe it's better to move it to the documentation (I don't know where yet)

ipc-client/CMakeLists.txt
blarz: Thanks!

t-wissmann:
> > The main change worth discussing is that I've replaced the cmake-specific X11-detection-part by multiple `pkg-config` calls. I feel more comfortable using the (I think) more standard `pkg-config`. What do you think? I can also move the pkg-config migration to another PR (I'll take a look at it tomorrow).
> 
> I am pretty indifferent to the pkg-config changes. Not sure what the preferred way for this is "in cmake-world"?
> It feels a bit weird, as I thought cmake is the tool that takes care of this? Wouldn't it also work by adding `find_package(Foo REQUIRED)` directives instead of `pkg_check_modules(FOO foo)` you added? What was it that made you choose?
> 
> I'd appreciate a separate PR, maybe explaining the reasoning behind it. But it is not really that import to me :)

I've extracted it to the PR #1108 and I'm also trying to explain why I found debugging easier here. I first tried `find_package(Freetype…)` but I had issues because I didn't know whether it is `Freetype` or `Freetype2` and I had the impression that this varied (either over time or between distros). It feels strange that the build system implements heuristics for library detection here, and it feels quite fragile to me. I'm wondering why `find_package` isn't just a wrapper around `pkg-config`; I mean, the pkg-config infrastructure, why not use it? (Of course I don't know how things on windows are; which seems a perfectly fine reason for cmake to avoid `pkg-config`, but it does not apply in our case of herbstluftwm of course)

t-wissmann:
I've looked at the docstrings in
* /usr/share/cmake-3.19/Modules/FindFreetype.cmake
* /usr/share/cmake-3.19/Modules/FindFontconfig.cmake
and adjusted this PR to pkg-config free PR (12f06b2e5fb57da6a746a2e3484af0af4944edf0). It now builds now my arch linux system. Let's see what the CI says (I see that the 32-bit already complains because it can't find `FindFontconfig.cmake`)

src/CMakeLists.txt
t-wissmann: I've removed it in #1108

INSTALL
t-wissmann: Yes, good idea! I've added them.

blarz:
> > > The main change worth discussing is that I've replaced the cmake-specific X11-detection-part by multiple `pkg-config` calls. I feel more comfortable using the (I think) more standard `pkg-config`. What do you think? I can also move the pkg-config migration to another PR (I'll take a look at it tomorrow).
> > 
> > 
> > I am pretty indifferent to the pkg-config changes. Not sure what the preferred way for this is "in cmake-world"?
> > It feels a bit weird, as I thought cmake is the tool that takes care of this? Wouldn't it also work by adding `find_package(Foo REQUIRED)` directives instead of `pkg_check_modules(FOO foo)` you added? What was it that made you choose?
> > I'd appreciate a separate PR, maybe explaining the reasoning behind it. But it is not really that import to me :)
> 
> I've extracted it to the PR #1108 and I'm also trying to explain why I found debugging easier here. I first tried `find_package(Freetype…)` but I had issues because I didn't know whether it is `Freetype` or `Freetype2` and I had the impression that this varied (either over time or between distros). It feels strange that the build system implements heuristics for library detection here, and it feels quite fragile to me. I'm wondering why `find_package` isn't just a wrapper around `pkg-config`; I mean, the pkg-config infrastructure, why not use it? (Of course I don't know how things on windows are; which seems a perfectly fine reason for cmake to avoid `pkg-config`, but it does not apply in our case of herbstluftwm of course)

Thanks for the PR and the explanation!

Seeing how the 32bit build is making problems without pkgconfig, I'd say we go for it.

INSTALL
blarz: Thanks!

share/autostart
blarz: I know. Maybe reduce it down to two? And is the default case a font that is always installed?

share/autostart
```diff
@@ -121,7 +121,10 @@ hc set frame_transparent_width 5
...
+hc attr theme.title_font 'Dejavu Sans:pixelsize=12'  # example using Xft
+# hc attr theme.title_font "Liberation Mono:pixelsize=12:antialias=false:autohint=false"
+# hc attr theme.title_font '-*-fixed-medium-r-*-*-13-*-*-*-*-*-*-*'
+hc attr theme.padding_top 2
```
blarz, line 127: Why is this added?

t-wissmann: I've restricted the font examples to two (one xft, one XLFD). The `text_height` only is the height of the text above it's baseline, and the `padding_top` adds space above the window content (but still below the title). So `padding_top` is used as `text_depth` here.

share/autostart
blarz: Looks good to me now

share/autostart
blarz: I see, thanks!

blarz (approved):
This PR was quite a ride! Looks good now :)

PR #1094, opened 2020-12-29
Support proper utf8 strings in window titles

t-wissmann:
This uses the XFontSet structure and XmbDrawString routines that are
capable of utf8 encoded strings. The setlocale() in the main is of great
importance, because otherwise, text drawing stops on the first non-ascii
character.

Unfortunately, XCreateFontSet() or XmbDrawString() seem to be
incompatible with the terminus font on my installation.

If the XCreateFontSet fails, then we fall back to plain XFontStruct and
XDrawString.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -380,13 +380,24 @@ void Decoration::redrawPixmap() {
                        inner.width,
                        inner.height - dec->last_actual_rect.height);
     }
-    if (s.title_height() > 0 && s.title_font->data().xFontStruct_) {
-        XSetForeground(g_display, gc, get_client_color(s.title_color));
-        XFontStruct* font = s.title_font->data().xFontStruct_;
-        int font_x_offset = s.padding_left() + s.border_width;
-        XSetFont(g_display, gc, font->fid);
-        XDrawString(g_display, pix, gc, font_x_offset, s.title_height(),
-                    client_->title_().c_str(), client_->title_().size());
+    if (s.title_height() > 0) {
+        FontData& fontData = s.title_font->data();
+        string title = client_->title_();
+        Point2D titlepos = {
+            static_cast<int>(s.padding_left() + s.border_width()),
+            static_cast<int>(s.title_height())
+        };
+        if (fontData.xFontSet_) {
+            XSetForeground(g_display, gc, get_client_color(s.title_color));
+            XmbDrawString(g_display, pix, fontData.xFontSet_, gc, titlepos.x, titlepos.y,
+                    title.c_str(), title.size());
+        } else if (fontData.xFontStruct_) {
+            XSetForeground(g_display, gc, get_client_color(s.title_color));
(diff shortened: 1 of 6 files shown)


t-wissmann:
windowtitle-x11-font-unicode

src/main.cpp
```diff
@@ -516,6 +517,10 @@ static void sigaction_signal(int signum, void (*handler)(int)) {
 int main(int argc, char* argv[]) {
     Globals g;
     parse_arguments(argc, argv, g);
+
+    if(!setlocale(LC_CTYPE, "") || !XSupportsLocale()) {
+        fputs("warning: no locale support\n", stderr);
```
blarz, line 522: Why fputs when in the the rest of the main we use std::cerr?

src/main.cpp
```diff
@@ -516,6 +517,10 @@ static void sigaction_signal(int signum, void (*handler)(int)) {
 int main(int argc, char* argv[]) {
     Globals g;
     parse_arguments(argc, argv, g);
+
+    if(!setlocale(LC_CTYPE, "") || !XSupportsLocale()) {
```
blarz, line 521: Nit pick: Can you add a space between if and the opening parentheses?

src/decoration.cpp
```diff
@@ -380,13 +380,24 @@ void Decoration::redrawPixmap() {
...
-                    client_->title_().c_str(), client_->title_().size());
+    if (s.title_height() > 0) {
+        FontData& fontData = s.title_font->data();
+        string title = client_->title_();
+        Point2D titlepos = {
+            static_cast<int>(s.padding_left() + s.border_width()),
+            static_cast<int>(s.title_height())
+        };
+        if (fontData.xFontSet_) {
+            XSetForeground(g_display, gc, get_client_color(s.title_color));
```
blarz, line 391: Could be moved before the if because it is used the same way in both branches?

blarz (changes requested):
Looks good, and the Firefox tabs don't have messed up characters anymore, yay ;)

Left some tiny comments

src/main.cpp
t-wissmann: Because I've stole the line from dwm :'-)... It's fixed now.

src/main.cpp
t-wissmann: Sure (same issue as for the `fputs`...)

src/decoration.cpp
t-wissmann: It should stay there, because for xft (PR #1091) it is different.

src/main.cpp
blarz: I see :D Thanks :)

src/main.cpp
blarz: Thanks!

src/decoration.cpp
blarz: Alright! Thanks for clearing that up

blarz (approved):
Looks good, thanks!

Issue #1093, opened 2020-12-28
BUGS file removal breaks build


The-Compiler:
1282c4f976f70e68d5878d67a32c5faac19009d5 breaks cmake:

```
Install the project...
-- Install configuration: "Release"
CMake Error at cmake_install.cmake:46 (file):
  file INSTALL cannot find
  "/home/florian/.cache/yay/herbstluftwm-git/src/herbstluftwm/BUGS": No such
  file or directory.
```

Also the file is still referenced in various docs as well:

````console
$ git grep BUGS                                              
CMakeLists.txt:install(FILES BUGS NEWS DESTINATION ${DOCDIR})
INSTALL:A list of known bugs is listed in BUGS. If you found other bugs or want to
NEWS:things like EWMH or rules aren't implemented yet (See BUGS file for a list of
doc/herbstclient.txt:BUGS
doc/herbstclient.txt:See the *herbstluftwm* distribution BUGS file.
doc/herbstluftwm.txt:BUGS
doc/herbstluftwm.txt:See the *herbstluftwm* distribution BUGS file.
src/client.cpp:         * twice.  see BUGS for more information
```

t-wissmann:
Sorry about this!

t-wissmann:
@blarz fixed it in #1095. :-)

PR #1095, opened 2020-12-29
Fix references to removed BUGS file

blarz:
Unbreaks install and removes all other references as reported in #1093.
Rather reference the Github issues page everywhere.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -206,7 +206,7 @@ void Client::window_focus() {
 
     if (this != lastfocus) {
         /* FIXME: this is a workaround because window_focus always is called
-         * twice.  see BUGS for more information
+         * twice.
          *
          * only emit the hook if the focus *really* changes */
         // unfocus last one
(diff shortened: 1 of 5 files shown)


src/client.cpp
```diff
@@ -206,7 +206,7 @@ void Client::window_focus() {
 
     if (this != lastfocus) {
         /* FIXME: this is a workaround because window_focus always is called
-         * twice.  see BUGS for more information
+         * twice.
```
blarz, line 209: Not sure what do here? Should we fill in the old info from BUGS into this comment?

t-wissmann: Right, we could insert
```
window_focus is called twice when switching desktops (before mapping
windows and after mapping them).
```
I'm also planning to clean the focus-mess up and move all the input focus handling to the X11-specific classes.

t-wissmann (approved):
Thanks. I think it's good to merge it, we can fix the one reference regarding `window_focus` later.

PR #1096, opened 2020-12-29
Test install target in CI run

blarz:
This should prevent issues like #1093

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -52,7 +52,7 @@ jobs:
 
       - name: Build
         run: |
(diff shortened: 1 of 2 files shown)


blarz:
Should go green in the CI as soon as #1095 is submitted.

t-wissmann (approved):
Thanks! When it's merged, we can also move the `--install` to a separate step in the workflow

blarz:
> Thanks! When it's merged, we can also move the `--install` to a separate step in the workflow

Makes sense.

I guess you need to merge this PR manually, as it modifies the workflow and mergify isn't allow to auto-submit this.

t-wissmann:
> Thanks! When it's merged, we can also move the `--install` to a separate step in the workflow

It is a separate step in PR #1099.

PR #1099, opened 2020-12-30
Make CI steps more fine-grained

t-wissmann:
This splits the build step in the github action workflow into smaller
steps of cmake, compilation, and installation. Since we now understand
how the cache in github actions work (cf #1075), I've removed the
printf-debug steps.

The --cxx etc flags define environment variables, and they should only
be used during cmake, but not during the ninja compilation. Hence, it
should be sufficient to pass build_env to cmake. The only environment
variable relevant during the build process is CLANG_TIDY_BUILD_DIR which
is now only set in the `build.py --compile` step.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -45,19 +45,17 @@ jobs:
           path: .tox-cache
           key: focal-tox
 
-      - name: Inspect ccache after restore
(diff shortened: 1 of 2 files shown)


blarz (approved):
Looks better now, good idea!

PR #1100, opened 2021-01-02
Make the lock_tag attribute writable

t-wissmann:
For some reason, the lock_tag attribute was not writable (even in the
old source code at tag v0.7.2). In fact, the lock_tag and unlock_tag
commands do nothing except setting the attribute.

While at it, I've added a bunch of test cases for the lock_tag attribute
and the respective commands.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,7 @@ Current git version
 
   * Client window titles (controlled by the theme attributes 'title_height',
     'title_color', 'title_font')
+  * The 'lock_tag' attribute is now writable.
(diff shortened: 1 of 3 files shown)


blarz:
Changed the commit message from

> For some reason, the lock_tag attribute was now writable

to *not* writable.

blarz (approved):
Makes sense and I like the tests!

t-wissmann:
> to _not_ writable.

Thanks! :-)

PR #1103, opened 2021-01-03
CI: Use i386 pkg-config on trusty

blarz:
Hopefully fixes problems with CMake/pkg-config not finding some libs.


t-wissmann (approved):
Thanks a lot!

PR #1104, opened 2021-01-03
Add minimization keybinding to default autostart

blarz:
#1032 added this feature and had that nice keybind command in the commit message.
It fits perfectly into the default autostart.

### DIFF
--- a/share/autostart
+++ b/share/autostart
@@ -85,6 +85,7 @@ hc keybind $Mod-r remove
 hc keybind $Mod-s floating toggle
 hc keybind $Mod-f fullscreen toggle
 hc keybind $Mod-Shift-f set_attr clients.focus.floating toggle
+hc keybind $Mod-m set_attr attr clients.focus.minimized toggle
 hc keybind $Mod-p pseudotile toggle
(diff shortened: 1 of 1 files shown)


tiosgz:
Guess you’ll have to change your commit details or hlwm will get a new contributor ;)

blarz:
> Guess you’ll have to change your commit details or hlwm will get a new contributor ;)
> 
> // I see it’s okay in the resulting commit in #1105

D'oh, I did the commit on my work laptop.. Thanks for noticing, I'll fix it.

t-wissmann:
mergify does a squash commit and I'd assume the user opening the PR will become the commit's author. I think it's a good idea to have it as a default binding. I personally have `Mod-m` on maximize.sh and `Mod-Shift-m` on the minimize command, because I use maximization more often. But we can also add it like this and swap the keybindings later.

t-wissmann:
For some reason, the github CI does not show up for this pr. Maybe it returns when we press 'update branch' (after there is something new in master)

blarz:
I will open a new PR with my regular account and will change to Mod-Shfit-m

PR #1105, opened 2021-01-03
Remove unused hc_send_command_once

blarz:
Doesn't seem to be used anywhere.


t-wissmann (approved):
I think it's fine to remove it. I once planned to ship a `libherbstclient` and thought that this could be of use there. But  it's more likely that we will use python bindings instead of libherbstclient :)

PR #1106, opened 2021-01-03
Add 'help' command

t-wissmann:
The 'help' command prints information on a given attribute or object.
The tests so far check that it prints something useful and matches it
against the 'doc' field in the json documentation.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,7 +9,7 @@ Current git version
 
   * Client window titles (controlled by the theme attributes 'title_height',
     'title_color', 'title_font')
-  * New command 'help' for online help on the object tree
+  * New command 'help' for live documentation on the object tree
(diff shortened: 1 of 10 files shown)


NEWS
```diff
@@ -9,6 +9,7 @@ Current git version
 
   * Client window titles (controlled by the theme attributes 'title_height',
     'title_color', 'title_font')
+  * New command 'help' for online help on the object tree
```
tiosgz, line 12: This sounds like a browser will be opened with some online help :( Is it a typo of *inline*?

t-wissmann: Hm right, maybe 'live documentation' is better (instead of 'online help').

tiosgz: Much better, thanks!

blarz (approved):
Looks good to me!

PR #1108, opened 2021-01-04
Use pkg-config for library detection

t-wissmann:
Instead of using the FindX11.cmake module that comes with cmake, we now
use pkg-config for the compiler and linker flags. This allows us to
explicitly state which libraries we use. For FindX11.cmake it just
tried to detect what's there, with little output. for pkg-config, cmake
now clearly prints which libraries are found, and which are missing.

This also adds the libraries freetype and xft that will be necessary
for xft font rendering.

### DIFF
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -1,5 +1,14 @@
 ## get project dependencies
 # Xlib
-find_package(X11 REQUIRED)
+find_package(PkgConfig)
+
+pkg_check_modules(X11 x11)
+pkg_check_modules(XRANDR xrandr)
+pkg_check_modules(XINERAMA xinerama)
+pkg_check_modules(XEXT xext)
+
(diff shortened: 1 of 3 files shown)


t-wissmann:
I've included the freetype and xft dependencies, to investigate whether the memory leaking happens already here.

blarz:
> I've included the freetype and xft dependencies, to investigate whether the memory leaking happens already here.

Seems like it is not the case. Should we leave them here or move it to the xft PR?

t-wissmann:
> > I've included the freetype and xft dependencies, to investigate whether the memory leaking happens already here.
> 
> Seems like it is not the case. Should we leave them here or move it to the xft PR?

Let's leave them here (I'm now mentioning it in this PR's description / commit message)

blarz (approved):

PR #1109, opened 2021-01-04
Add minimization keybinding to default autostart

blarz:
#1032 added this feature and had that nice keybind command in the commit message.
It fits perfectly into the default autostart.

### DIFF
--- a/share/autostart
+++ b/share/autostart
@@ -85,6 +85,7 @@ hc keybind $Mod-r remove
 hc keybind $Mod-s floating toggle
 hc keybind $Mod-f fullscreen toggle
 hc keybind $Mod-Shift-f set_attr clients.focus.floating toggle
+hc keybind $Mod-Shift-m set_attr clients.focus.minimized true
 hc keybind $Mod-p pseudotile toggle
 # The following cycles through the available layouts within a frame, but skips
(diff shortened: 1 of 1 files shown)


t-wissmann (approved):
Thanks :) It's a nice follow up to #1104.

t-wissmann:
Again, the github actions were not tirggered.

tiosgz:
> Again, the github actions were not tirggered.

Isn’t it because `share/` is ignored?
https://github.com/herbstluftwm/herbstluftwm/blob/9421d93e7473448325b4b2ad70e6848f0c1cdfbf/.github/workflows/hlwm.yml#L10-L13

t-wissmann:
> > Again, the github actions were not tirggered.
> 
> Isn’t it because `share/` is ignored?
> https://github.com/herbstluftwm/herbstluftwm/blob/9421d93e7473448325b4b2ad70e6848f0c1cdfbf/.github/workflows/hlwm.yml#L10-L13

Oh, good catch; thanks! So let's remove this ignore (the autostart is even tested by the test suite, so it should not be ignored).

t-wissmann:
I've fixed it in bf6c1a9380eb41c09340d97e2ce54660098697f1 and pushed it to master directly.

PR #1111, opened 2021-01-04
Add attribute watchers

### LINKED ISSUE #1098: Watching attributes (comments requested)
Right now, it is not possible to get notified when an attribute value changes. For example, when showing the current layout algorithm in the panel (pursued in issue #71), one does not know yet when an update of the panel is necessary.

Right now, I'm planning to add a command `watch` such that one can run:
```
hc watch tags.focus.tiling.focused_frame.algorithm
```
This will emit one hook whenever the attribute has a different value, and the format is probably:
```python
attribute_changed ATTRIBUTE_PATH NEW_VALUE
```
For example:
```
attribute_changed tags.focus.tiling.focused_frame.algorithm grid
```

There are two kinds of changes:
  * The value of the attribute changes (e.g. another layout is selected)
  * The links between objects change (e.g. another frame is selected)

Because of the second item, the existence of dynamic links (such as `focused_frame`), and dynamic attributes, the best implementation I can think of is plain polling. That is, we have to read all the watched attributes in every iteration of the main algorithm. This would imply that if a attribute changes forth and back within one command / event, it will not be noticed. (Maybe this is desirable, maybe not).

If you have comments/ideas, let me know.

t-wissmann:
This implements a commonly requested feature allowing to be notified
when an attribute changes. The implementation just saves the value of
the watched attributes and periodically checks if the value has been
changed.

There is still room for optimization: One optimization is to replace the
plain map from string to string to a more advanced structure that speeds
up scanForChanges(); another is to use the signals where possible.

This fixes #1098. Also, this implements the missing piece of #71.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -835,6 +835,11 @@ new_attr *bool*|*color*|*int*|*string*|*uint* 'PATH' ['VALUE']::
     with +my_+. If 'VALUE' is supplied, then it is written to the attribute (if
     this fails the attribute still remains).
 
+[[watch]]watch 'PATH'::
+    Watch the value of the given attribute 'PATH'. Whenever the value changes
+    from 'OLDVALUE' to 'NEWVALUE', a hook is emitted: +
+    +attribute_changed+ 'PATH' 'NEWVALUE' 'OLDVALUE'
+
 remove_attr 'PATH'::
     Removes the user defined attribute 'PATH'.
 
@@ -1659,6 +1664,11 @@ On special events, herbstluftwm emits some hooks (with parameters). You can
 receive or wait for them with link:herbstclient.html[*herbstclient*(1)]. Also custom hooks can be
 emitted with the *emit_hook* command. The following hooks are emitted by
 herbstluftwm itself:
+
+attribute_changed 'PATH' 'NEWVALUE' 'OLDVALUE'::
+    The attribute 'PATH' was changed from 'OLDVALUE' to 'NEWVALUE'. Requires
+    that the attribute 'PATH' has been passed to <<watch,the +watch+ command>>
+    before.
 
 fullscreen [on|off] 'WINID'::
     The fullscreen state of window 'WINID' was changed to [on|off].
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -838,7 +838,7 @@ new_attr *bool*|*color*|*int*|*string*|*uint* 'PATH' ['VALUE']::
 [[watch]]watch 'PATH'::
     Watch the value of the given attribute 'PATH'. Whenever the value changes
     from 'OLDVALUE' to 'NEWVALUE', a hook is emitted: +
-    +attribute_changed+ 'PATH' 'NEWVALUE' 'OLDVALUE'
+    +attribute_changed+ 'PATH' 'OLDVALUE' 'NEWVALUE'
 
 remove_attr 'PATH'::
     Removes the user defined attribute 'PATH'.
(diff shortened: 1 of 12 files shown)


tiosgz:
Two questions:
* `hc watch something` is a command that says hlwm to notify every `hc -i` of `something`, & so it doesn’t block whatever executed it? I.e. it doesn’t idle & watch the hook?
* is there any particular reason for the `new` `old` order? `tag_renamed` has `old` `new`

t-wissmann:
> * `hc watch something` is a command that says hlwm to notify every `hc -i` of `something`, & so it doesn’t block whatever executed it? I.e. it doesn’t idle & watch the hook?

Exactly. It does not block. Calling `watch PATH` is a declaration that someone (e.g. a panel) is interested in the changes of `PATH`. One still needs `hc -i` to be actually informed about changes.

> * is there any particular reason for the `new` `old` order? `tag_renamed` has `old` `new`

It felt strange to have the `old` value closer to the attribute name than the `new` one. I remember that we had a discussion on the order when `tag_renamed` was extended. I agree that `old` `new` makes more sense, but any order is fine for me. I've updated the PR to this `old` `new` order in eab4a38cb39d08977d5ecf1b3ff86e438927323e so if we decide to go back to `new` `old`, we can just revert that.

tiosgz:
> Exactly. It does not block. Calling `watch PATH` is a declaration that someone (e.g. a panel) is interested in the changes of `PATH`. One still needs `hc -i` to be actually informed about changes.

Okay, thanks for the info.
> It felt strange to have the `old` value closer to the attribute name than the `new` one. I remember that we had a discussion on the order when `tag_renamed` was extended. I agree that `old` `new` makes more sense, but any order is fine for me. I've updated the PR to this `old` `new` order in [eab4a38](https://github.com/herbstluftwm/herbstluftwm/commit/eab4a38cb39d08977d5ecf1b3ff86e438927323e) so if we decide to go back to `new` `old`, we can just revert that.

If anything, it should be consistent across all hooks, shouldn’t it? Otherwise I don’t care so much, it just seemed unusual to have first something that’s newer.

blarz:
> > It felt strange to have the `old` value closer to the attribute name than the `new` one. I remember that we had a discussion on the order when `tag_renamed` was extended. I agree that `old` `new` makes more sense, but any order is fine for me. I've updated the PR to this `old` `new` order in [eab4a38](https://github.com/herbstluftwm/herbstluftwm/commit/eab4a38cb39d08977d5ecf1b3ff86e438927323e) so if we decide to go back to `new` `old`, we can just revert that.
> 
> If anything, it should be consistent across all hooks, shouldn’t it? Otherwise I don’t care so much, it just seemed unusual to have first something that’s newer.

I like "new old" a bit better, feels more.. "natural"? But I agree with @tiosgz consistency matters more.

doc/herbstluftwm.txt
```diff
@@ -835,6 +835,11 @@ new_attr *bool*|*color*|*int*|*string*|*uint* 'PATH' ['VALUE']::
     with +my_+. If 'VALUE' is supplied, then it is written to the attribute (if
     this fails the attribute still remains).
 
+[[watch]]watch 'PATH'::
```
blarz, line 838: Why is watch twice in here? Is it syntax I don't know about?

blarz: Also should we add the new command to NEWS?

blarz (changes requested):
Looks good so far. But can you add some tests, too, maybe?

doc/herbstluftwm.txt
t-wissmann: I've updated the NEWS file.
The `[[watch]]` defines a label and it's referenced in the `HOOKS` section later (yielding a clickable link in the html doc).

blarz: Ah! I see, thanks!

t-wissmann:
> Looks good so far. But can you add some tests, too, maybe?

Oh sure.. I've forgot to run `git add` on the new `test_watchers.py` file...

blarz:
> > Looks good so far. But can you add some tests, too, maybe?
> 
> Oh sure.. I've forgot to run `git add` on the new `test_watchers.py` file...

Thanks, looking at it right now :)

t-wissmann:
> This change looks good to me.
> 
> Should we add an "unwatch" command, too? Now, or in a separate PR?

In a separate PR; I'm trying to keep the commits as small as possible.

blarz:
> > This change looks good to me.
> > Should we add an "unwatch" command, too? Now, or in a separate PR?
> 
> In a separate PR; I'm trying to keep the commits as small as possible.

Alright!

blarz (approved):

PR #1112, opened 2021-01-04
Let objects have documentation

t-wissmann:
This makes Object inherit from HasDocumentation and parses the
respective calls to setDoc() into the json document. The attr command
now prints this object documentation, and this allows us to add test
cases that it matches the doc in the json file.

Note that the object documentation is something different than the
documentation on these "child entries" because e.g. the documentation on
clients.focus is "the focused client" whereas the documentation on
Client is that it represents "a managed window".

### DIFF
--- a/doc/gendoc.py
+++ b/doc/gendoc.py
@@ -418,6 +418,7 @@ class ObjectInformation:
         self.member2info = {}  # mapping (classname,member) to ChildInformation or AttributeInformation
         self.member2init = {}  # mapping (classname,member) to its initializer list
         self.member2doc = {}  # mapping (classname,member) to its doc string
+        self.class2doc = {}  # mapping classname to its doc string
 
     def base_class(self, subclass, baseclass):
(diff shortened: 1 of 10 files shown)


blarz:
Shouldn't this trigger the json diff posting by the bot?

t-wissmann:
> Shouldn't this trigger the json diff posting by the bot?

I'm running `ci/diff-json-doc.py` locally because it's not yet part of github actions.

blarz:
> > Shouldn't this trigger the json diff posting by the bot?
> 
> I'm running `ci/diff-json-doc.py` locally because it's not yet part of github actions.

I see, that makes sense. Thanks for the diff!

blarz (approved):

t-wissmann:
One line had trailing whitespaces after I did the merge commit in the github web interface. I've fixed it but this removed the approval.

blarz (approved):

PR #1113, opened 2021-01-04
Correctly interpret a panel's WM_STRUT

### LINKED ISSUE #1110: Polybar is rendered on top of other windows after updating from 0.9.0 to 0.9.1
After updating to 0.9.1, polybar with `offset-y=12` and `override-direct=false` is rendered on top of  other windows

**0.9.1**

**0.9.0**

t-wissmann:
The entries of WM_STRUT need to be interpreted with respect to the
borders of the root window (not of the monitor). The present change now
translates WM_STRUT into a rectangle which is then used to allocate
space at the monitor edge.

Only if WM_STRUT is empty, the geometry of the panel window is used as
the rectangle.

Formerly, we ignored WM_STRUT and always used the panel window, which
broke the support for panels that deliberately leave some space around
the panel. I originally took the window geometry, because 1. it was
easier (but as it turned out wronger) than computing the rectangle from
WM_STRUT and 2. this check for the case where the panel geometry and the
WM_STRUT were contradictory, as in #815.

This fixes #1110.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,8 @@ Current git version
   * Client window titles (controlled by the theme attributes 'title_height',
     'title_color', 'title_font')
   * The 'lock_tag' attribute is now writable.
+  * Bug fixes:
+    - Detect panels that leave a bit of space between the panel and the monitor edge.
 
 Release 0.9.1 on 2020-12-28
 ---------------------------
--- a/src/panelmanager.cpp
+++ b/src/panelmanager.cpp
@@ -142,40 +142,34 @@ PanelManager::ReservedSpace PanelManager::computeReservedSpace(Rectangle mon)
         // we only reserve space for the panel if the panel defines
         // wmStrut_ or if the aspect ratio clearly indicates whether the
         // panel is horizontal or vertical
-        if (!p.wmStrut_.empty() || intersection.height > intersection.width) {
-            // vertical panels
-            if (intersection.x == mon.x) {
-                rs.left_ = intersection.width;
+        if (p.wmStrut_.empty()) {
+            // if the panel does not define wmStrut, then
+            // try to detect it automatically from the intersection
+            if (intersection.height > intersection.width) {
+                // vertical panels
+                if (intersection.x == mon.x) {
+                    rs.left_ = intersection.width;
+                }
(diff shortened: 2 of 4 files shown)


blarz:
This makes stuff weird with my xfce4-panel, when I set it to "vertical" (which I normally don't but did for testing this).

The panel is shown on left side of the screen, vertical, looks correct.
But no clients are shown on that tag on that monitor with the panel.

blarz:
> This makes stuff weird with my xfce4-panel, when I set it to "vertical" (which I normally don't but did for testing this).
> 
> The panel is shown on left side of the screen, vertical, looks correct.
> But no clients are shown on that tag on that monitor with the panel.

To be specific, this happens only if I have the panel on the second monitor, which is to the right.

t-wissmann:
OK, I'll check it again.

t-wissmann:
> This makes stuff weird with my xfce4-panel, when I set it to "vertical" (which I normally don't but did for testing this).
> 
> The panel is shown on left side of the screen, vertical, looks correct.
> But no clients are shown on that tag on that monitor with the panel.

I've reimplemented the pad-computation and now I'm confident that it's correct. I've also added lots of test cases for it (one of them revealing a `start_y` vs `end_y` typo). It seems to work with all panel positions of `lxpanel`. Unfortunately, xfce4-panel does not support xinerama, so I can't test it in `Xephyr`. If the tests don't convince you yet, can you test it whether it works for you?

blarz:
> > This makes stuff weird with my xfce4-panel, when I set it to "vertical" (which I normally don't but did for testing this).
> > The panel is shown on left side of the screen, vertical, looks correct.
> > But no clients are shown on that tag on that monitor with the panel.
> 
> I've reimplemented the pad-computation and now I'm confident that it's correct. I've also added lots of test cases for it (one of them revealing a `start_y` vs `end_y` typo). It seems to work with all panel positions of `lxpanel`. Unfortunately, xfce4-panel does not support xinerama, so I can't test it in `Xephyr`. If the tests don't convince you yet, can you test it whether it works for you?

Thanks! Tests convinced me but I tested with xfce4-panel anyway and it works with vertical now, too :)

blarz (approved):

PR #1115, opened 2021-01-05
Update monitor paddings if auto_detect_panels changes

t-wissmann:
Automatically update the padding attributes of every monitor if the
setting auto_detect_panels is changed.


blarz (dismissed):
Makes sense!

t-wissmann:
Oh, even the merge commit dismisses the PR approval. (which makes sense because it allowed me change the conflicting files arbitrarily).

blarz:
> Oh, even the merge commit dismisses the PR approval. (which makes sense because it allowed me change the conflicting files arbitrarily).

That makes sense, yes. Re-approval incoming ;)

blarz (approved):

PR #1116, opened 2021-01-05
Fix minor things in watcher code

blarz:
Remove (currently?) unused vector, fix whitespace
and add a test for missing arguments.


t-wissmann (approved):
Thanks! I've used vector while implementing it; so it's fine to remove it.

PR #1118, opened 2021-01-06
Remove early and now unused winterbreeze concepts

t-wissmann:
This removes unused concepts from the early winterbreeze code:

  - HookManager: we now have it as 'watchers'. The name hook (from
    pre-winterbreeze times) was a bit unfortunate, so I think 'watchers'
    is the better name.
  - Entries in the enum class Type for objects: These names are never
    used, and we simply take the name of the C++ classes in the json doc.
  - Also, the attribute system is now more 'statically' typed, so we
    don't need a type 'Attribute'. It's only use was in the class
    Attribute itself. Removing it there made Attribute an abstract
    class, and by this we implicitly ensure every attribute has a type.
  - Hence, we can remove the ATTRIBUTE_... prefix for the remaining
    types.


blarz (approved):
I enjoy the approval of PRs removing lots of dead code!

PR #1119, opened 2021-01-06
Fix typo in xftwidth

### LINKED ISSUE #1117: [panel.sh] typo in textwidth name for guix
https://github.com/herbstluftwm/herbstluftwm/blob/299b6d54ed46f1ec7d9d0f14dc2c54c0611bb18c/share/panel.sh#L33-L34

I guess it should be `xft` but I’m not a guixer so can’t be sure. @samtinel could you fix it please? (asking you coz you were the one to [add it](https://github.com/herbstluftwm/herbstluftwm/commit/8c78994ae73c3371cb9598bced8735c8a1ebf14e) & so you should know)

t-wissmann:
Insert the correct name of the xftwidth textwidth replacement. Now the
content of $textwidth matches the argument to `which`.

Thanks to Bohdan Potměkleč (tiosgz) for reporting. This fixes #1117.

### DIFF
--- a/share/panel.sh
+++ b/share/panel.sh
@@ -31,7 +31,7 @@ if which textwidth &> /dev/null ; then
 elif which dzen2-textwidth &> /dev/null ; then
     textwidth="dzen2-textwidth";
 elif which xftwidth &> /dev/null ; then # For guix
-    textwidth="xtfwidth";
+    textwidth="xftwidth";
 else
     echo "This script requires the textwidth tool of the dzen2 project."
     exit 1


tiosgz:
The reason why I didn’t PR it & instead asked samtinel was that I can’t be sure it’s `xft` and not `xtf` in guix, no matter which makes more sense or that there’s a `xftwidth` elsewhere.

So, do you have an evidence which form is right?

blarz (approved):

blarz:
> The reason why I didn’t PR it & instead asked samtinel was that I can’t be sure it’s `xft` and not `xtf` in guix, no matter which makes more sense or that there’s a `xftwidth` elsewhere.
> 
> So, do you have an evidence which form is right?

Googling around definitely gives more hits for xftwidth https://guix.gnu.org/packages/xftwidth-20170402/
The only hit for xtf is the panel.sh of this repo :D

tiosgz:
> Googling around definitely gives more hits for xftwidth https://guix.gnu.org/packages/xftwidth-20170402/
> The only hit for xtf is the panel.sh of this repo :D

Okay, you convinced me, together with that the packages are defined in one place: https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/xdisorg.scm#n1644 (dzen is right before xftwidth)

PR #1120, opened 2021-01-07
Don't expose name 'HSFont' in json doc

t-wissmann:
The class name 'HSFont' should stay C++ internal. The HS-prefix only
exists because the type 'Font' already exists in Xlib. Instead, just
call it 'font' in the json doc.

At some point in the future, this mapping from internal C++ types to
user-readable type names should be specified in the C++ source code.


t-wissmann:
In principle, this PR is ready, but I'd like to use this PR to test the changes of #1121.

t-wissmann:
> In principle, this PR is ready, but I'd like to use this PR to test the changes of #1121.

The diffing seems to work, so let's try it here :-)

blarz (approved):
Alright!

PR #1122, opened 2021-01-07
www: Generate object tree documentation

t-wissmann:
This adds a script that transforms the json documentation into a
user-friendly asciidoc document. This asciidoc document is added to the
website (under "documentation" -> "objects"). The goal is that in the
future, we can use this generated document as a replacement for the
hand-written "objects" section in the man page.

For testing, run `make` in the www folder and open `object-doc.html`.


blarz (approved):
Resulting document looks good to me, I didn't look too closely at the format-doc.py

PR #1124, opened 2021-01-08
New 'geometry' attribute for every monitor

t-wissmann:
The geometry (i.e. the rectangle) is the central property of a monitor,
so it deserves to be an attribute. This introduces 'rectangle' as a type
for attributes and adds some test cases for it.


t-wissmann:
So far, we have called it `rectangle` instead `geometry` in the docs, but I think `geometry` is the more appropriate term (`rectangle` is only the type, and not was it does).

blarz:
> So far, we have called it `rectangle` instead `geometry` in the docs, but I think `geometry` is the more appropriate term (`rectangle` is only the type, and not was it does).

Agreed!

blarz (dismissed):
Looks good and works for me!

blarz:
Seems like #1126 messed the build up for this PR

t-wissmann:
> Seems like #1126 messed the build up for this PR

One include was missing, let's see what iwyu thinks about it :)

blarz (approved):

PR #1125, opened 2021-01-08
Fix bug when removing tag with minimized clients

### LINKED ISSUE #1123: display bug when merging minimized windows
Found a minor bug: when I merge a tag containing a minimized window, it appears to be floating, until I focus it with the mouse.

Steps to reproduce:

1. `herbstclient add test`
2. `herbstclient add merged`
3. `herbstclient use test`
4. open a terminal
5. `herbstclient set_attr  clients.focus.minimized true`
6. `herbstclient use merged`
7. `herbstclient merge_tag test`
8. the terminal first appears to be floating, but becomes tiled again as soon as it is focused with the mouse.

t-wissmann:
When merging a tag into a currently visible tag, one must not make all
moved clients visible, because they may be in minimized state. Instead,
let the tag itself take care of the client visibility.

This fixes #1123, thanks to @apm256 for reporting this!


src/tagmanager.cpp
```diff
@@ -146,9 +146,7 @@ int TagManager::removeTag(Input input, Output output) {
     Monitor* monitor_target = find_monitor_with_tag(targetTag);
```
blarz, line 146: That comment above this line is not really correct anymore is it?

t-wissmann: I've updated the comment.

blarz: Thanks!

blarz (approved):
Nice, thanks!

PR #1127, opened 2021-01-08
Move Input/Output to new commandio module

t-wissmann:
The types module should provide the core type definitions and this has
nothing to do with the types Input and Output which are used in the
interface of commands.


src/utils.h
```diff
@@ -9,6 +9,7 @@
 #include <sstream>
 #include <string>
 
+#include "commandio.h"
 #include "types.h"
```
blarz, line 13: Is this still needed?

src/settings.h
```diff
@@ -4,6 +4,7 @@
 #include <string>
 
 #include "attribute_.h"
+#include "commandio.h"
 #include "globals.h"
 #include "object.h"
 #include "types.h"
```
blarz, line 10: Is this still needed?

blarz (changes requested):
In some places types.h include could probably be removed. Not sure why IWYU doesn't complain

t-wissmann:
I think you're right! I've removed both includes and resolved the merge conflicts.

src/settings.h
blarz: Thanks!

blarz (approved):

PR #1128, opened 2021-01-08
In-source doc for tags

t-wissmann:
This transfers the docs on Tag and TagManager from the man page to the
source code.


src/tagmanager.cpp
```diff
@@ -30,6 +30,15 @@ TagManager::TagManager()
         Ewmh::get().updateCurrentDesktop();
         tag_set_flags_dirty();
     });
+
+    setDoc(
+        "The tags (or virtual desktops or workspaces). This contains "
```
blarz, line 35: The second sentence is missing something right? "This contains has an entry"?

blarz (changes requested):

src/tagmanager.cpp
t-wissmann: I've removed `has`. I was writing this part from scratch because there was no direct doc on `TagManager` in the original man page.

blarz: Thanks!

blarz (approved):

PR #1129, opened 2021-01-08
In-source doc for clients

t-wissmann:
This transfers the docs on Client and ClientManager from the man page to
the source code.


src/client.cpp
```diff
@@ -98,6 +98,39 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
     visible_.setDoc("whether this client is rendered currently");
     parent_frame_.setDoc("the frame contaning this client if the client is tiled");
     setDoc("a managed window");
+
```
blarz, line 101: Can we add setDoc for "minimized" here, too? I think we totally forgot about the minimized attribute in the docs.

blarz (changes requested):

src/client.cpp
t-wissmann: I've added it.

blarz: Nice, thanks!

blarz (approved):

PR #1130, opened 2021-01-08
In-source doc for monitors

t-wissmann:
This transfers the docs on Monitor and MonitorManager from the man page
to the source code.


src/monitor.cpp
```diff
@@ -64,7 +64,19 @@ Monitor::Monitor(Settings* settings_, MonitorManager* monman_, Rectangle rect_,
 
     stacking_window = XCreateSimpleWindow(g_display, g_root,
                                              42, 42, 42, 42, 1, 0, 0);
-
+    setDoc("The monitor is a rectangular part on the screen that holds "
+           "precisely one tag at a time. The pad attributes reserve "
+           "space on the monitor\'s edge for panels, so this space "
+           "(given in number of pixels) is never occuped by tiled clients.");
```
blarz, line 70: occupied?

blarz (changes requested):

src/monitor.cpp
t-wissmann: Fixed, thanks!

blarz (approved):

PR #1131, opened 2021-01-08
In-source doc for theme

t-wissmann:
This transfers the docs on Theme related classes from the man page
to the source code. Also this introduces setChildDoc() to avoid the
ambiguity of setDoc() on ChildMember_<> objects.


blarz:
The commit message doesn't seem to fit here.

src/theme.cpp
```diff
@@ -83,6 +107,9 @@ DecTriple::DecTriple()
     for (auto it : children) {
         it->scheme_changed_.connect([this]() { this->triple_changed_.emit(); });
     }
+    active.setChildDoc("configures the decoration the focused client");
```
blarz, line 110: "of" missing?

src/theme.cpp
```diff
@@ -83,6 +107,9 @@ DecTriple::DecTriple()
...
+    normal.setChildDoc("configures the decoration of focused clients");
```
blarz, line 111: Is it really the focused client?

blarz (changes requested):

src/theme.cpp
t-wissmann: Fixed.

src/theme.cpp
t-wissmann: Fixed.

src/theme.cpp
tiosgz: I know they make things clearer, but are the negatives really needed? Imagine you decide to add another special decoration; do you want to add another word here? Or otherwise: is the condition *explicitly* that the client is unfocused & non-urgent, or is it the fallback?

Sorry for being so perfectionist, I cannot help myself :(

src/theme.cpp
blarz: Thanks!

src/theme.cpp
t-wissmann: I didn't know a better way to express it and didn't just want to say 'normal' here. But I think I now have a better doc: "the default decoration scheme for clients"

tiosgz: I like the new description much better, thank you!

blarz: Thanks!

blarz:
> The commit message doesn't seem to fit here.

The PR looks good to me, can you update the commit message body? I don't think you are updating Monitor docs in this ;)

t-wissmann:
> > The commit message doesn't seem to fit here.
> 
> The PR looks good to me, can you update the commit message body? I don't think you are updating Monitor docs in this ;)

Ah, I forgot :) It's updated now!

blarz (approved):

PR #1136, opened 2021-01-13
Move example.py out of herbstluftwm Python binding package

Flowdalic:
Any Python file under python/herbstluftwm will be installed via
distutils. This means that example.py will also potentially land in
the system-wide site-packages directory, even though it is not
strictly part of herbstluftwm's Python bindings.

For example on Gentoo:

/usr/lib/python3.8/site-packages/herbstluftwm/example.py
/usr/lib/python3.8/site-packages/herbstluftwm/__init__.py
/usr/lib/python3.8/site-packages/herbstluftwm/__pycache__/__init__.cpython-38.opt-1.pyc
/usr/lib/python3.8/site-packages/herbstluftwm/__pycache__/example.cpython-38.pyc
/usr/lib/python3.8/site-packages/herbstluftwm/__pycache__/example.cpython-38.opt-1.pyc
/usr/lib/python3.8/site-packages/herbstluftwm/__pycache__/__init__.cpython-38.pyc
/usr/lib/python3.8/site-packages/herbstluftwm/__pycache__/__init__.cpython-38.opt-2.pyc
/usr/lib/python3.8/site-packages/herbstluftwm/__pycache__/example.cpython-38.opt-2.pyc

Instead, (small) code examples often should be placed in the
appropiate location under /usr/share/doc.


t-wissmann:
I see, thanks! Can you please also adjust the import in line 2 of `tests/test_python_binds.py`?

Don't worry about the "Diff object tree"; please ignore it.

t-wissmann:
I've updated the according filepath in the test suite (I didn't know how to fix the `import` but with #1168 it became much easier)

t-wissmann (approved):
Thanks!

PR #1145, opened 2021-01-16
Use generated object doc in the man page

t-wissmann:
Use the asciidoc file generated by format-doc.py in herbstluftwm.txt. To
make it fit into the man page and to make the include mechanism work
also for the website, a few adjustments were necessary.

Unfortunately, if a include fails, then asciidoc does not fail but only
prints a warning.


doc/CMakeLists.txt
```diff
@@ -57,10 +74,13 @@ if (WITH_DOCUMENTATION)
     gen_manpage(herbstluftwm 1)
     gen_manpage(herbstluftwm-tutorial 7)
 
+
```
blarz, line 77: Superfluous newline?

doc/format-doc.py
```diff
@@ -199,21 +198,15 @@ def run(self, clsname, path=[]):
 def main():
     parser = argparse.ArgumentParser(description='Generate object documentation')
     parser.add_argument('jsondoc', help='the hlwm-doc.json file')
+    parser.add_argument('--objects-intro',
```
blarz, line 201: Is this parameter used anywhere? I cannot find it

blarz (changes requested):

doc/CMakeLists.txt
t-wissmann: [x] removed.

doc/format-doc.py
t-wissmann: That's a leftover from a previous approach. I've removed it.

blarz: Thanks!

blarz (approved):

PR #1147, opened 2021-01-17
Add colors to the object doc

t-wissmann:
This adds colors to the attribute names and types and child names in the
object doc. Also, this inserts a colon between the attribute
specification and the doc string.

Moreover, 'code' is an html tag not a class, so main.css had to be
adjusted.

### DIFF
--- a/doc/format-doc.py
+++ b/doc/format-doc.py
@@ -139,14 +139,14 @@ class ObjectDocPrinter:
         reference_cls_doc = self.reference_to_class_doc(clsname, path)
         if reference_cls_doc is not None:
             identifier, text = reference_cls_doc
-            print(f'(see <<{identifier},{text}>>)')
+            print(f'For attributes and children, see <<{identifier},{text}>>')
             return
         # otherwise, print it here:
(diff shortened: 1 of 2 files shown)


blarz:
Sometimes there is a whitespace between the colon and the attribute, is this intended?

> active : configures the decoration of the focused client.

but 
> string reset: writing this resets all attributes to a default value

for example

t-wissmann:
> Sometimes there is a whitespace between the colon and the attribute, is this intended?
> 
> > active : configures the decoration of the focused client.
> 
> but
> 
> > string reset: writing this resets all attributes to a default value
> 
> for example

`active` is a child entry whereas `reset` is an attribute. The whitespace issue for child entries existed before, but I can fix it along with this PR.

t-wissmann:
When trying to make the colon usage for child entries consistent, I had to adjust further parts, but I hope the output is OK now.

blarz:
> When trying to make the colon usage for child entries consistent, I had to adjust further parts, but I hope the output is OK now.

Looks very good now, thank you for fixing this!

blarz (approved):

PR #1152, opened 2021-01-23
Fix frame objects in 'mirror' and 'rotate'

### LINKED ISSUE #1151: mirror-command sets wrong client_count if frame is accessed by index
## Info

herbstluftwm 0.9.1 (285f0f54 built on 2021-01-21)

## Problem

The mirror-command swaps two frames including the windows contained in those frames (which is the correct behavior). However, after executing the mirror-command the attribute `client_count` of both frames is wrong if its accessed by index with the following command (in this example index 0):

```sh
herbstclient get_attr tags.focus.tiling.root.0.client_count
```

However, if frame with index 0 is focused and the client_count is accessed by `focused_frame` then the client_count is correct:

```sh
herbstclient get_attr tags.focus.tiling.focused_frame.client_count
```

## Steps to reproduce

1. Create two frames. One left with index 0 and one right with index 1.
2. Open two terminals in frame 0 and one terminal in frame 1.
3. Verify the index and client_count of both frames as instructed below.
4. Mirror both frames with `herbstclient mirror`
5. Verify the index and client_count of both frames again as instructed below

It can be seen that the client_count for frame 0 is still 2 and for frame 1 its still 1 if accessed by index, although it should be reversed.

## client_count verification

Follow this steps once for each frame to verify the frame-index and `client_count`. `idx` refers in the following to the index of the `focused_frame`.

```sh
herbstclient get_attr tags.focus.tiling.focused_frame.index
herbstclient get_attr tags.focus.tiling.focused_frame.client_count
herbstclient get_attr tags.focus.tiling.root.idx.client_count
```

t-wissmann:
The commands 'mirror' and 'rotate' updated the child frame pointers
explicitly without updating the pointers for the frame objects. Which
caused the frame object tree to be wrong.

The present change fixes this and adds a test for it.

Thanks to @astier for an excellent bug report on this!

Fixes #1151.

### DIFF
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -1213,11 +1213,16 @@ def test_focused_frame_child(hlwm):
 
 
 def verify_frame_indices(hlwm, root='tags.focus.tiling.root', index=[]):
+    """traverse all frame objects under the given 'root' frame object
+    and verify that their 'index' attribute has the correct value according to
+    their path in the frame tree.
+    """
     path = root.rstrip('.') + '.' + '.'.join(index)
     assert hlwm.get_attr(path.rstrip('.') + '.index') == ''.join(index)
     if '0' in hlwm.list_children(path):
+        # if 'path' is a frame split object
         verify_frame_indices(hlwm, root=root, index=index + ['0'])
-        verify_frame_indices(hlwm, root=root, index=index + ['0'])
+        verify_frame_indices(hlwm, root=root, index=index + ['1'])
 
 
 @pytest.mark.parametrize("old,new", [
--- a/tests/test_layout.py
+++ b/tests/test_layout.py
@@ -1212,6 +1212,14 @@ def test_focused_frame_child(hlwm):
         assert hlwm.get_attr('tags.0.tiling.focused_frame.index') == focused_index
 
 
+def verify_frame_indices(hlwm, root='tags.focus.tiling.root', index=[]):
+    path = root.rstrip('.') + '.' + '.'.join(index)
+    assert hlwm.get_attr(path.rstrip('.') + '.index') == ''.join(index)
+    if '0' in hlwm.list_children(path):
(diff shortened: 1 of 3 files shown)


tests/test_layout.py
```diff
@@ -1212,6 +1212,14 @@ def test_focused_frame_child(hlwm):
         assert hlwm.get_attr('tags.0.tiling.focused_frame.index') == focused_index
 
 
+def verify_frame_indices(hlwm, root='tags.focus.tiling.root', index=[]):
```
blarz, line 1215: Could you add a short docstring on what and how this function does what it does?
I found it not that obvious (but that could be because I didn't have enough coffee just yet)

tests/test_layout.py
```diff
@@ -1212,6 +1212,14 @@ def test_focused_frame_child(hlwm):
...
+    path = root.rstrip('.') + '.' + '.'.join(index)
+    assert hlwm.get_attr(path.rstrip('.') + '.index') == ''.join(index)
+    if '0' in hlwm.list_children(path):
+        verify_frame_indices(hlwm, root=root, index=index + ['0'])
+        verify_frame_indices(hlwm, root=root, index=index + ['0'])
```
blarz, line 1220: Why is this called twice with the same arguments?

t-wissmann: Because it actually should be `'1'` in the second line. (The `if '0'`... only checks that it's a frame split).

tests/test_layout.py
t-wissmann: I've added a docstring. :-)

tests/test_layout.py

tests/test_layout.py
blarz: Thank you, it is so much clearer now :)

blarz (approved):

PR #1153, opened 2021-01-23
Add test cases for decorations and fonts

t-wissmann:
This adds a screenshot-based test suite for the drawing of decorations.
The idea is that after changing the theme or other parameters of the
window decorations, we take a screenshot and tell from the pixels
whether everything is drawn correctly.

So far, this contains tests for the main, inner, and outer border color
and tests for the font-rendering.

Also this adds another x11-sync to x11's sync_with_hlwm(), because
syncing with hlwm itself only makes sense if all the client information
is already present at the xserver.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -578,6 +578,33 @@ def x11_connection(xvfb):
     display.close()
 
 
+class RawImage:
+    def __init__(self, data, width, height):
+        """
+        A raw image is an array of size width*height containing
+        (r, g, b) triples, line by line
+        """
+        self.data = data
+        self.width = width
+        self.height = height
+        assert len(data) == width * height
+
+    def pixel(self, x, y):
+        """return (r, g, b) triple for a pixel"""
+        return self.data[x + self.width * y]
+
+    def color_count(self, rgb_triple):
+        count = 0
+        for pix in self.data:
+            if pix == rgb_triple:
+                count += 1
+        return count
+
+    @staticmethod
+    def rgb2string(rgb_triple):
+        return '#%02x%02x%02x' % rgb_triple
+
+
 @pytest.fixture()
 def x11(x11_connection):
     """ Short-lived fixture for interacting with the X11 display and creating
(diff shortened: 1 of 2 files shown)


t-wissmann:
This should boost up the coverage a lot. Also I'm curious whether the CI works fine, because I do not fully understand all these colorspace notions and already noticed that some colors were slightly altered when testing it in xvfb locally.

tests/test_decorations.py
```diff
@@ -0,0 +1,111 @@
...
+import pytest
+
+font_pool = [
+    '-*-fixed-medium-r-*-*-13-*-*-*-*-*-*-*',
+    'Dejavu Sans:pixelsize=12:bold'
+]
+
+
+def test_window_border_plain(hlwm, x11):
+    color = (0x9f, 0xbc, 0x12)
```
blarz, line 11: Some colors are given in hex, some in decimals. Should we unify the format in this test suite?
And: what are those colors, do they come from the defaults or default autostart?

blarz (commented):
This looks very good!
Just a quick comment while skimming over this change.

I'll have a more detailed look at it tonight hopefully.

tests/test_decorations.py
t-wissmann: The colors were chosen such that all integers within a test case are unique and not 0. Sometimes the hex is more convenient, sometimes decimals are. Hex-values correspond to the string representation of colors (e.g. `#9fbc12` here), but decimals are printed by pytest if an assertion fails. So I've used hex in test cases that worked out of the box and I've used decimals in test cases that needed much effort until they went through :-)

t-wissmann: Also, sometimes I have the impression that the colors are altered slightly, but I'm not sure (that's what I meant with the colorspace comment)

tests/conftest.py
```diff
@@ -697,7 +724,43 @@ def create_client(self, urgent=False, pid=None,
...
+            # raw.data is a array of pixel-values, which need to
+            # be interpreted using the colormap:
+            colorDict = {}
+            for pixel in raw.data:
+                colorDict[pixel] = None
+            colorPixelList = list(colorDict)
+            colorRGBList = attr.colormap.query_colors(colorPixelList)
+            for pixelval, rgbval in zip(colorPixelList, colorRGBList):
+                # Useful debug output if something blows up again (which is likely)
+                # print(f'{pixelval} -> {rgbval.red}  {rgbval.blue} {rgbval.green}')
```
blarz, line 747: I know this is just for debugging, but shouldn't it be red, green, blue instead of red blue green?

Could lead to some confusion while debugging.

tests/test_decorations.py
blarz: I see, thanks for explaining!

blarz (changes requested):
One tiny thing in the commented out section, but other than that this looks good to me

tests/conftest.py
t-wissmann: Indeed! Thanks, I've reordered it.

blarz: Thanks!

blarz (approved):
Cool! Happy that there is finally some way to test this

blarz:
Job ran into a timeout
```
[gw0] [ 97%] PASSED tests/test_x11.py::test_client_moveresizes_itself Timeout (0:04:00)!
Thread 0x00007f24411e9700 (most recent call first):
  File "/__w/herbstluftwm/herbstluftwm/.tox/py38/lib/python3.8/site-packages/execnet/gateway_base.py", line 400 in read
  File "/__w/herbstluftwm/herbstluftwm/.tox/py38/lib/python3.8/site-packages/execnet/gateway_base.py", line 432 in from_io
  File "/__w/herbstluftwm/herbstluftwm/.tox/py38/lib/python3.8/site-packages/execnet/gateway_base.py", line 967 in _thread_receiver
  File "/__w/herbstluftwm/herbstluftwm/.tox/py38/lib/python3.8/site-packages/execnet/gateway_base.py", line 220 in run
  File "/__w/herbstluftwm/herbstluftwm/.tox/py38/lib/python3.8/site-packages/execnet/gateway_base.py", line 285 in _perform_spawn
```

I retriggered it

PR #1155, opened 2021-01-30
Build doc in CI

t-wissmann:
This creates a new cmake target for the doc and calls it in another ci
build job. While at it, I've fixed a race condition in our cmake setup,
which was caused because the json-target was used before gen_json_doc()
created it (thus the order change in doc/CMakeLists.txt).


.github/workflows/hlwm.yml
```diff
@@ -210,3 +210,38 @@ jobs:
...
+  build-doc:
+    name: Build Documentation and website
+    runs-on: ubuntu-latest
+    env:
+      BUILD: build-${{ github.run_number }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: install asciidoc
+        run: sudo apt-get install --no-install-recommends asciidoc xsltproc docbook-xsl
```
blarz, line 223: Should we just add this to ci/Dockerfile.ci-focal?

.github/workflows/hlwm.yml
```diff
@@ -210,3 +210,38 @@ jobs:
...
+      - name: restrict cmake to doc
+        run: |
+          cat > CMakeLists.txt <<EOF
```
blarz, line 226: Why is this necessary? Because we cannot do a "docs-only build"?

.github/workflows/hlwm.yml
t-wissmann: Yes, then we could have a 'full build' there.

.github/workflows/hlwm.yml
t-wissmann: The default ubuntu-latest of github actions does not have the X11-libraries installed and so the `cmake` fails. But I've one had the x libraries installed, `cmake` succeeds and `make all_doc` would only build the doc.

.github/workflows/hlwm.yml
t-wissmann: I mean: we could do this eventually. I still think it's ok to have a separate job for the time being.

blarz: Sure!

.github/workflows/hlwm.yml
blarz: Makes sense. So as soon as the dependencies are in the Dockerfile we could just use a regular build and do a all_doc there, sometime in the future.

blarz (approved):

blarz:
.github/workflows was touched in this PR, so mergify wants you to merge manually for safety reasons ;)

t-wissmann:
Thanks :)

PR #1158, opened 2021-02-07
stack: Remove unused methods

blarz:
Coverage and grep indicate that those methods are not used anymore.

### DIFF
--- a/src/stack.cpp
+++ b/src/stack.cpp
@@ -135,25 +135,10 @@ void Stack::extractWindows(bool real_clients, function<void(Window)> yield) {
     }
 }
 
-void Stack::restack() {
-    if (!dirty) {
-        return;
-    }
-    vector<Window> buf;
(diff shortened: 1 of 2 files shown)


src/stack.cpp
```diff
@@ -135,28 +135,13 @@ void Stack::extractWindows(bool real_clients, function<void(Window)> yield) {
...
 
-void Stack::restack() {
-    if (!dirty) {
-        return;
-    }
-    vector<Window> buf;
-    extractWindows(false, [&buf](Window w) { buf.push_back(w); });
-    XRestackWindows(g_display, buf.data(), buf.size());
-    dirty = false;
-    Ewmh::get().updateClientListStacking();
```
t-wissmann, line 146: We should call `updateClientListStacking` in the Monitor's `restack()`...

t-wissmann (approved):
Thanks! My comment is more a TODO and unrelated to this PR :)

PR #1160, opened 2021-02-07
Extend tag manager test suite

blarz:
Fix some low hanging fruit. There is more in there.


t-wissmann (approved):
Thanks! Looks good to me

PR #1162, opened 2021-02-07
Use link time optimization in CI build

blarz:
Maybe finds some hidden problems and/or dead code.

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -73,7 +73,7 @@ if args.cmake:
     build_env.update({
         'CC': args.cc,
         'CXX': args.cxx,
-        'CFLAGS': '--coverage -Werror -fsanitize=address,leak,undefined',
+        'CFLAGS': '--coverage -Werror -fsanitize=address,leak,undefined -flto=auto -ffat-lto-objects',
         'CXXFLAGS': '--coverage -Werror -fsanitize=address,leak,undefined',
     })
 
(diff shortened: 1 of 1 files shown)


blarz:
Compilation seems to work. With gcc I see:
```../src/root.cpp: In member function ‘__ct_base ’:
../src/root.cpp:28:1: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
sys	0m31.103s
   28 | Root::Root(Globals g, XConnection& xconnection, IpcServer& ipcServer)
      | ^
```

Not sure how critical that is...

t-wissmann:
> Compilation seems to work. With gcc I see:
> 
> ```
> ../src/root.cpp:28:1: note: variable tracking size limit exceeded with ‘-fvar-tracking-assignments’, retrying without
> sys	0m31.103s
>    28 | Root::Root(Globals g, XConnection& xconnection, IpcServer& ipcServer)
>       | ^
> ```
> 
> Not sure how critical that is...

Hm not sure what `retrying without` means. Does it mean it compiles as if `-flto` was not passed?

dnnr:
Unfortunately, this doesn't seem to provide any benefits in terms of improving code quality. Unless there's any other good reason for it, let's not do this.

blarz:
Closing this.

PR #1167, opened 2021-02-10
Focused client must be above fullscreen window

t-wissmann:
On a non-focused monitor, the focused window should always be above
every fullscreen window. The new test failed before the present change.

Also, this removes an old comment (the shadow hack is now handled in the
loop before).


tests/test_stack.py
```diff
@@ -222,3 +222,28 @@ def test_tag_floating_state_on(hlwm, focus_idx):
     # we didn't do any raising since
     hlwm.call('floating off')
     assert helper_get_stack_as_list(hlwm) == stack_before
+
+
+def test_focused_on_other_monitor_above_fullscreen(hlwm):
+    # on the unfocused monitor, put a fullscreen window
+    # and focus another window. then the focused window should
+    # be visible!
```
blarz, line 230: I would add a "after switching monitor focus" somewhere in here, this is kind of crucial to this test, right?

blarz (changes requested):
I tested it and it totally makes sense!

I would appreciate it if you could adjust the introduction comment of the newly added test a bit, would make it easier (for me ;)) to understand what is happening.

tests/test_stack.py
t-wissmann: it is only crucial that it's not on the 'focused' monitor, but it does not matter whether (a) the monitor focus is switched or (b) the other monitor was never focused.  (I just chose (a) here because `jumpto` makes the test easier to write)

t-wissmann: I've refined the test case to also include (b). On master, both test cases (a and b) fail, and the present PR fixes both :)

blarz: Thanks for extending the tests! :)

blarz (approved):
Totally makes sense!

PR #1170, opened 2021-02-14
tagmanager: Test move_index with relative index

blarz:


tests/test_tags.py
```diff
@@ -79,6 +79,24 @@ def test_move_focused_client_to_new_tag(hlwm, move_command):
...
+def test_move_focused_client_by_relative_index(hlwm):
+    hlwm.call('add foobar')
+    assert hlwm.get_attr('tags.0.client_count') == '0'
+    assert hlwm.get_attr('tags.1.client_count') == '0'
+
+    winid, _ = hlwm.create_client()
+    assert hlwm.get_attr('tags.0.client_count') == '1'
+    assert hlwm.get_attr('tags.1.client_count') == '0'
+
+    hlwm.call('move_index +1')
```
t-wissmann, line 91: I think this is not a good example for a relative index because the target tag's absolute index is 1, too.

t-wissmann (changes requested):
thanks for adding a test; I think one should add another tag in between to really test that the parameter to `move_index` is really used as a relative index.

tests/test_tags.py
blarz: You're absolutely right, I'll update the test with a better one

t-wissmann (approved):
Thanks for improving :-)

PR #1171, opened 2021-02-15
Re-use pre-built doc in tarball builds

t-wissmann:
In source tarballs, the built documentation is already shipped along
with the tarball. The CMakeLists of doc now detects a prebuilt
documentation, and in this case simply copies the documentation to the
build directory (without requiring asciidoc to be installed).

This implies that one does not need to (and in fact should not) disable
`WITH_DOCUMENTATION` when building a tarball without having asciidoc
installed.

In the CI job for the docs, the tarball is now built and tested. In the
ancient ubuntu build, this tarball is used instead of the source from
the git repo. To be sure that everything worked, the job verifies that
the man page has been installed to the DESTDIR.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -180,7 +180,8 @@ jobs:
           cd herbstluftwm-*/build/
           mkdir -v -p $DESTDIR
           ninja -v install
-          find $DESTDIR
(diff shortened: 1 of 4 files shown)


t-wissmann:
I think it is ready now :-) (and can be used to create the tarball for the next&soon release)

Issue #992, opened 2020-08-14
undesirable interaction between compositor and windowborder drawing


bklaase:
When using VTE based terminals (termite, xfce4-terminal, etc) I get the behavior, that when a compositor is enabled, the borders become transparent. Urxvt (not being VTE based) is fine.

I found this (old) discussion on the matter in the termite issuetracker:
https://github.com/thestinger/termite/issues/200

Is this the same problem? Most wm's (bspwm, xfce4, openbox) don't exibit the issue, but some do (windowchef, xmonad, hlwm).

t-wissmann:
I once had this issue but I'm not experiencing it anymore
![term-border-transparent](https://user-images.githubusercontent.com/9048813/90269503-0543b380-de59-11ea-8f70-41b0e87dfc69.png)
I'm using picom 8-2 (picom 8, archlinux package version 2), and as I've just noticed, I don't have any special configuration, so I assume the default picom config from the arch package is used.

t-wissmann:
Ah, picom is still respecting `~/.config/compton.conf`, and this is mine: https://github.com/t-wissmann/dotfiles/blob/master/config/compton.conf

bklaase:
Thank you!
I will compare and let other know here what setting causes, or fixes the behavior!

bklaase:
Alas, same picom, same distro and even with your config:
![image](https://user-images.githubusercontent.com/8513117/90277547-1b0ba580-de66-11ea-89fb-30a95aadb68e.png)
![image](https://user-images.githubusercontent.com/8513117/90277640-42627280-de66-11ea-81b8-93104294b726.png)

Perhaps a driver thing. (intel gpu)... perhaps I'll test with nvidia and radeon cards later.

To be honest, my largest gripe was with borders getting ugly on max layout. When reading the man I found 
`hide_covered_windows` and those issues went away. It's even better now then having terminal content shine through, so I am happy :)

I'll close this, thanks for thinking along!

t-wissmann:
Definitely can confirm that I had this too. Maybe it's really a driver thing. By the way, hlwm is already doing what's proposed in https://github.com/chjj/compton/issues/240#issuecomment-62245919 -- namely to use a windows color-scheme for color conversion.

Barbarossa93:
[...]

PR #1178, opened 2021-02-17
Fix transparency of 32 bit client's border

### LINKED ISSUE #1173: [BUG] Semi-transparent borders with picom
To continue from [here](https://github.com/herbstluftwm/herbstluftwm/issues/992), there is currently an issue with certain windows that have a depth of 32bit causing the borders to appear semi-transparent when picom is running. Programs affected by this are zathura, vte-based terminals such as termite, and likely others. According to this [picom](https://github.com/yshui/picom/issues/285) issue, this problem is caused by the window manager not specifying a premultiplied value for the border color. For example, dwm applied with the alpha patch appears to fix the problem. However, it's not clear if this is indeed a problem with the window manager or with intel drivers as discussed [here](https://github.com/yshui/picom/issues/586). Interestingly, testing between urxvt with depth 24 and 32, there is no problem, suggesting that it's entirely possible this is an intrinsic problem in vte-based terminals, zathura, and/or picom.

Sorry to pester you @t-wissmann, but have you had a chance to figure out what the dwm alpha patch is doing?

EDIT:
Removing `xf86-video-intel` (the 2D acceleration Xorg intel drivers) on arch linux made no difference. I suppose this could still mean that its a problem with intel, but I have a hunch that it's less likely to be so

t-wissmann:
When using a compositor (e.g. compton or picom), the border of some
windows is sometimes (undesirably) semitransparent for applications with
alpha settings (e.g. sakura or katarakt). This only happens for certain
graphics, I can reproduce it on intel graphics but not on nvidia
chipsets. The issue is probably that the highest significant byte of the
color is not initialized.

In the present patch, this byte is set to 0xff, thus fixing the
byte controling alpha-value to 'fully opaque'. This also used by dwm[1]
and wmutils[2] to fix this issue.

While at it, the return type of get_client_color() is turned into
'unsigned long' because this is the type of xcol.pixel.

This fixes #1173 / #992 

[1] https://dwm.suckless.org/patches/alpha/dwm-fixborders-6.2.diff
[2] https://github.com/wmutils/opt/issues/48#issuecomment-667541248

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -296,16 +296,18 @@ void Decoration::redraw()
     }
 }
 
-unsigned int Decoration::get_client_color(Color color) {
+unsigned long Decoration::get_client_color(Color color) {
     XColor xcol = color.toXColor();
     if (colormap) {
         /* get pixel value back appropriate for client */
         XAllocColor(g_display, colormap, &xcol);
-        return xcol.pixel;
+        // explicitly set the alpha-byte to 0xff (fully opaque)
+        return xcol.pixel | (0xffu << 24);
     } else {
         /* get pixel value back appropriate for main color map*/
         XAllocColor(g_display, DefaultColormap(g_display, g_screen), &xcol);
-        return xcol.pixel;
+        // explicitly set the alpha-byte to 0xff (fully opaque)
+        return xcol.pixel | (0xffu << 24);
     }
 }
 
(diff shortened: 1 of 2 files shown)


t-wissmann:
This also fixes the issue for @Barbarossa93 (https://github.com/herbstluftwm/herbstluftwm/issues/1173#issuecomment-780697808) so I think it's good to test in `master` further :-)

Issue #1066, opened 2020-12-09
shift command should move between monitors


t-wissmann:
The `shift` command should move windows to a neighboured monitor if it reaches the monitor boundary. Have configured it manually https://github.com/t-wissmann/dotfiles/blob/master/config/herbstluftwm/autostart#L208..L233 but I think it should be the default behaviour (just like it is already for `focus`)

PR #1179, opened 2021-02-17
Move the tiling part of 'shift' to FrameTree

t-wissmann:
In order to pave the way for #1066, move all the frame-specific code of
the `shift` command to the FrameTree class (where it actually belongs).

No tests are needed because test_layout.py has already enough tests for
the `shift` command.


blarz (approved):
Looks good to me!

PR #1180, opened 2021-02-17
Prepare Colors for alpha values

t-wissmann:
Extend Color by a alpha value and extend the parser and printer to print
colors in the #rrggbbaa format. The alpha value is not used anywhere at
the moment, but it will come in handy for decoration drawing.

Also, this adds a new test suite file dedicated to all the type parsers
and printers.

### DIFF
--- a/src/decoration.h
+++ b/src/decoration.h
@@ -10,6 +10,7 @@
 class Client;
 class Settings;
 class DecorationScheme;
+class XConnection;
 
 class Decoration {
 public:
--- a/src/decoration.h
+++ b/src/decoration.h
@@ -10,7 +10,6 @@
 class Client;
 class Settings;
 class DecorationScheme;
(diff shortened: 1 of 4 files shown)


src/decoration.h
```diff
@@ -10,6 +10,7 @@
 class Client;
 class Settings;
 class DecorationScheme;
+class XConnection;
```
blarz, line 13: Why is this needed in this change?

blarz (changes requested):
Just one tiny thing

src/decoration.h
t-wissmann: Thanks! That was a mistake (I was already at the next step...). I've removed the line!

blarz: Thanks!

blarz (approved):

PR #1181, opened 2021-02-19
Use XConnection in Decoration

t-wissmann:
Instead of using the globals g_display etc, the Decoration class now
only takes the display pointer from the XConnection class.
Also, instead of the 'default screen', the screen number is taken from
the XConnection class.

Since this XConnection is also used in the destructor of decoration, we
should wait with resetting the global Root* variable and first shut all
submodules down.


blarz (approved):
Change seems fine, tested locally, too

PR #1184, opened 2021-02-20
Initialize rectangle before XGetGeometry()

t-wissmann:
If a window disappears right before XGetGeometry() is called, then the
rectangle is not initialized. Hence, just initialize everything before
to be on the save side.

This was detected by address sanitization, reporting that jumps depended
on uninitialized memory, and it was triggered by the qt application
katarakt that creates two windows and hides them instantly again.

### DIFF
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -124,9 +124,9 @@ bool XConnection::checkotherwm() {
 }
 
 Rectangle XConnection::windowSize(Window window) {
-    unsigned int border, depth;
-    int x, y;
-    unsigned int w, h;
+    unsigned int border = 0, depth = 0;
+    int x = 0, y = 0;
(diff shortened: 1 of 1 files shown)


blarz:
Can we add a unit test for this?

t-wissmann:
Hm, difficult to test. All we can do is created and destroy a window quickly and see that hlwm does not crash. In the test suite we start hlwm with `--exit-on-xerror`, so we would need to disable this flag for the test case, because passing an non-existing window id to `XGetGeometry()` triggers an xerror.

blarz:
If it is not reproducible deterministically, it is maybe not worth it. Just wondered if we could easily add one.

But it seems like a good idea to have hlwm running with *SAN compiled in (at least as a developer).

PR #1185, opened 2021-02-20
Make default colors more transparent and quiet

t-wissmann:
Stop using the #9fbc00 color as the default border color, since it is a
quite aggressive color. I loved this color, when it was introduced by
pwmt.org applications, but I've stopped using it in my own settings some
years ago and I have the impression that many people don't like it when
starting with hlwm (and the default hlwm config).

Instead, the border color now is the same (dark) green tone as in the
frame border, and is optimized such that it also looks nice with or
without transparency.

### DIFF
--- a/share/autostart
+++ b/share/autostart
@@ -118,10 +118,10 @@ hc keybind $Mod-i jumpto urgent
 # theme
 hc attr theme.tiling.reset 1
 hc attr theme.floating.reset 1
-hc set frame_border_active_color '#222222'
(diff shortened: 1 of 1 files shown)


t-wissmann:
Here's how it looks (requires the true-transparency PR, which will come next). The background pattern is not included yet, maybe we should create some pattern and ship it with hlwm.

hlwm-true-transparency
hlwm-same-without-transparency

blarz (approved):
I like it!

PR #1186, opened 2021-02-20
Support true transparency in all decorations

t-wissmann:
This adds true transparency support in all frame and client decorations
(using the xrender X11 extension), where the level of transparency is
controlled by the alpha value of the colors. The only instance where the
alpha value does not work yet is at Xft fonts.

Per default, hlwm tries to initialize true transparency on startup (this
can be prevented with --no-transparency).

The test cases for the decoration colors now run in all combinations of
xrender being available at the server and hlwm trying to use it.
For xrender to work, we need to start xvfb with color depth 24 (rather
than 8), and this also results in the screenshots being in a slightly
different format.

On the technical side, the XConnection class now provides a singleton
object getter, which can be used to obtain the pointers to Visual and
Colormap. These structures are either the 'default' or special structures
if true transparency is used.

The new library dependency xrender should be available on all systems,
but we can think of making it an optional dependency, since
XConnection::tryInitTransparency() is the only function using the
library.

I've also tested it on my intel graphics laptop, and it also works fine
(in particular it keeps the fix of #1178 intact).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,7 @@ Current git version
   * Colors contain alpha-values (format #RRGGBBAA)
   * Bug fixes:
     - Fix mistakenly transparent borders of argb clients
+  * New dependency: xrender
 
 Release 0.9.2 on 2021-02-17
 ---------------------------
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,9 @@ older versions.
 Current git version
 -------------------
 
+  * True transparency support for frame and client decorations (requires a
+    compositor like picom, compton, or xcompmgr)
+  * Colors contain alpha-values (format #RRGGBBAA)
   * Bug fixes:
     - Fix mistakenly transparent borders of argb clients
 
--- a/cmake/dependencies.cmake
+++ b/cmake/dependencies.cmake
@@ -7,6 +7,9 @@ pkg_check_modules(XRANDR REQUIRED xrandr)
 pkg_check_modules(XINERAMA xinerama)
 pkg_check_modules(XEXT REQUIRED xext)
 
+# for transparency support
+pkg_check_modules(XRENDER REQUIRED xrender)
+
 # for xft support:
 pkg_check_modules(XFT REQUIRED xft)
 pkg_check_modules(FREETYPE REQUIRED freetype2)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -517,23 +517,42 @@ def xvfb(request):
     # also we add '-noreset' such that the server is not reset
     # when the last client connection is closed.
     #
-    # the optional parameter is the resolution. If you want another resolution,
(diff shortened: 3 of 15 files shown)


tests/conftest.py
```diff
@@ -848,9 +880,10 @@ def __init__(self, server='Xephyr', screens=[(800, 600)], extra_args=[]):
         # ---------------------------
         self.screen_rects = []
         current_x_offset = 0
+        depth = 24
         for idx, (width, height) in enumerate(screens):
             self.screen_rects.append([current_x_offset, 0, width, height])
-            geo = '{}x{}x8'.format(width, height)
+            geo = '{}x{}x{}'.format(width, height, depth)
```
blarz, line 886: Would an f-string make it more readable?

NEWS
```diff
@@ -7,6 +7,9 @@ older versions.
 Current git version
 -------------------
```
blarz, line 9: Missed that in the previous PR: but should we note the new default color theme?

tests/conftest.py
t-wissmann: I think it's fine this way because `'{}x{}x{}'` is very short.

NEWS
t-wissmann: Not sure whether it's relevant because it's only a new default setting and not a new feature/functionality.

tests/conftest.py
blarz: True, it's just my personal taste, but both versions are fine, yes!

NEWS
blarz: I don't know either. Not sure what the usual workflow for users actually is, when checking out a new hlwm release. I usually do a vimdiff of my autostart and the default one in shared/ and pick the cherries :)

It's not that important to me, just wanted to see what's your opinion about it :)

t-wissmann:
I've also tested it on my intel graphics laptop, and it also works fine, in particular it keeps the fix of #1178 intact. (I've added this also to the commit description).

And with this PR the semi-transparent colors from #1185 kick in:
hlwm-true-transparency

cmake/dependencies.cmake
```diff
@@ -7,6 +7,9 @@ pkg_check_modules(XRANDR REQUIRED xrandr)
 pkg_check_modules(XINERAMA xinerama)
 pkg_check_modules(XEXT REQUIRED xext)
 
+# for transparency support
+pkg_check_modules(XRENDER REQUIRED xrender)
```
blarz, line 11: Does this need to be added to the NEWS section (same as you did for xft etc.)

t-wissmann: I've added it :)

blarz: Thanks!

blarz (approved):
Nice!

Issue #1183, opened 2021-02-20
[Feature request] smart_frame_surroundings should remove the frame background, if frame_bg_transparent is set off


juacq97:
The `smart_frame_surroundings` option should remove the background of the frames when `frame_bg_transparent` is set off. The main usage IMO for the frame background is to differentiate between focused and unfocused frames, and this is not needed when there's just one frame on screen.

t-wissmann:
I'm not sure what the actual problem is you would like to solve or improve. Is it that you don't want to see the frame's background color shining through a semi-transparent terminal window? (at least this is what annoys me regularly...) If this is the issue also you would like to solve, then maybe the new PR #1187 solves this.

juacq97:
When you have just one container, and if you have the frame_transparency set off, then you'll see a colored rectangle when empty. Of course, that's the intended behavior, because you want to see the colores rectangle and to differentiate focused frames. 
But if there's just one frame, that behaviour should disapear, I mean, the frame should become transparent only when there's just one frame, like how the border become transparent

t-wissmann:
> When you have just one container, and if you have the frame_transparency set off, then you'll see a colored rectangle when empty. Of course, that's the intended behavior, because you want to see the colores rectangle and to differentiate focused frames.

Then, why don't you just activate transparency again?

> But if there's just one frame, that behaviour should disapear, I mean, the frame should become transparent only when there's just one frame, like how the border become transparent

I don't get this sentence. Are you requesting the following?
"If there is only one frame, then it should be entirely transparent"

juacq97:
> "If there is only one frame, then it should be entirely transparent"

That's exactly what I'm requesting 😄, if `frame_transparency` is set off, and there's only one frame, it should be entirely transparent.

I need to stop using google translate to write my comments

t-wissmann:
Ah, great that we could sort it out! :-)

> I need to stop using google translate to write my comments

Oh, yes, you definitely should stop using google translate for whole sentences, because they seem to change the 'logic'. Instead, feel free to practise writing English in the issues here :-)

By the way, I was confused a lot by `only when` in your sentence because `if` and `only if` have *opposite* meanings:

  * `X if Y` means: `Y implies X` (`Y ⇒ X`) 
  * `X only if Y` means: `X implies Y` (`X ⇒ Y`) (which is equivalent to ["never ... except when"](https://www.collinsdictionary.com/de/worterbuch/englisch/only-if))

[...]

PR #1187, opened 2021-02-20
Cut holes into frames behind tiled clients

t-wissmann:
For semi-transparent clients, it is unpleasant to see the frame
decoration shining through. Hence, we now cut holes into the frame's
decoration at the area where clients are. If no true transparency is
activated, no difference should be noticeable, and for true transparency,
one sees the wallpaper or desktop window shining through.

Internally, the last 'outer' geometry of every client is used, and so
this also interacts nicely with pseudotiled windows and the grid layout.

This should improve the cosmetics and implicitly also solve #1183.

### DIFF
--- a/src/layout.h
+++ b/src/layout.h
@@ -20,6 +20,7 @@ class Client;
 typedef std::function<void(Client*)> ClientAction;
 
 class HSTag;
+class FrameDecoration;
 class FrameLeaf;
 class FrameSplit;
 class Settings;
@@ -137,6 +138,7 @@ public:
     DynAttribute_<int> selectionAttr_;
     DynAttribute_<LayoutAlgorithm> algorithmAttr_;
 private:
+    friend class FrameDecoration;
     friend class FrameTree;
     // layout algorithms
     TilingResult layoutLinear(Rectangle rect, bool vertical);
--- a/src/x11-utils.cpp
+++ b/src/x11-utils.cpp
@@ -9,11 +9,10 @@
 using std::vector;
 
 /**
- * \brief   cut a rect out of the window, s.t. the window has geometry rect and
- * a frame of width framewidth remains
+ * \brief cut the given rectangles out of the window using XShape
  */
-void window_cut_rect_hole(Window win, int width, int height,
-                          const vector<Rectangle>& holes) {
+void window_cut_rect_holes(Window win, int width, int height,
+                           const vector<Rectangle>& holes) {
     // inspired by the xhole.c example
     // http://www.answers.com/topic/xhole-c
     Display* d = g_display;
--- a/src/x11-utils.cpp
+++ b/src/x11-utils.cpp
@@ -6,18 +6,19 @@
 
 #include "globals.h"
 
+using std::vector;
+
 /**
  * \brief   cut a rect out of the window, s.t. the window has geometry rect and
  * a frame of width framewidth remains
  */
-void window_cut_rect_hole(Window win, int width, int height, int framewidth) {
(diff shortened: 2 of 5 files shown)


t-wissmann:
For illustration, see the following screenshot and notice that one sees the blue frame border everywhere, but the red frame background does not shine through the semi-transparent client decorations:
no-frame-behind-clients

t-wissmann:
For illustration, this is essentially the same layout before the change in this PR:
before-patch

blarz:
Change looks good, the screenshots look promising as well. Can we add a test (based on screenshots)?

src/layout.h
```diff
@@ -137,6 +138,7 @@ class FrameLeaf : public Frame, public FrameDataLeaf {
     DynAttribute_<int> selectionAttr_;
     DynAttribute_<LayoutAlgorithm> algorithmAttr_;
 private:
+    friend class FrameDecoration;
```
blarz, line 141: Why is this needed?

src/x11-utils.h
```diff
@@ -2,12 +2,14 @@
...
+#include <vector>
 
+#include "rectangle.h"
 #include "x11-types.h"
 
-// cut a rect out of the window, s.t. the window has geometry rect and a frame
-// of width framewidth remains
-void window_cut_rect_hole(Window win, int width, int height, int framewidth);
+// cut the specified rectangles out of the window
+void window_cut_rect_hole(Window win, int width, int height,
```
blarz, line 11: Should we update the name to window_cut_rect_holes?

src/layout.h
blarz: Ah nevermind, tried it out and saw the compiler error

src/x11-utils.h
t-wissmann: I don't care too much :)

src/x11-utils.cpp
```diff
@@ -6,18 +6,19 @@
 
 #include "globals.h"
 
+using std::vector;
+
 /**
  * \brief   cut a rect out of the window, s.t. the window has geometry rect and
```
blarz, line 12: This comment is here and in the .h file and needs to be updated, too

t-wissmann:
> Change looks good, the screenshots look promising as well. Can we add a test (based on screenshots)?

Hm actually it should work. I didn't initially because these screenshots-testcases are quite fiddly most of the time. Also, I simply don't know what I get if I take a screenshot from a shape window, but maybe we can add test cases that ensure that we don't cut away too much.

src/x11-utils.h
t-wissmann: I've renamed it while  updating the comment.

src/x11-utils.cpp
t-wissmann: [x] fixed.

blarz: Thanks!

src/x11-utils.h
blarz: Thank you :)

blarz:
> > Change looks good, the screenshots look promising as well. Can we add a test (based on screenshots)?
> 
> Hm actually it should work. I didn't initially because these screenshots-testcases are quite fiddly most of the time. Also, I simply don't know what I get if I take a screenshot from a shape window, but maybe we can add test cases that ensure that we don't cut away too much.

I always appreciate more tests if possible. We can postpone it for now and add them in a separate PR and submit this one now, or we'll keep it open a little longer. Your choice. I'd also approve it in the current state.

blarz (approved):

PR #1188, opened 2021-02-22
Add screenshot-based tests for frame decoration

t-wissmann:
This adds tests for the XShape functionality in the frame decorations.
This covers both the setting frame_bg_transparent /
frame_transparent_width and the new functionality of hiding the frame
content behind clients (PR #1187 / f7ba8f9940).


blarz (approved):
Very cool, thank you for adding these!

PR #1189, opened 2021-02-25
Remove undefined function

t-wissmann:
The function list_padding is declared in MonitorManager but not defined.
It is however defined in Monitor.


blarz (approved):
Nice find. Sometimes I think that some linter/static analyzer should've informed us about this..

Issue #136, opened 2018-11-12
Remove argv_duplicate()


t-wissmann:
Since we now have ArgList, we can migrate all uses of `argv_duplicate` and then remove it.

t-wissmann:
In `src/`, `argv_duplicate` has been removed with the rewrite of spawn/wmexec in adf2613ba0ffbca (#1330). (there is still a definition in `ipc-client`, but this can stay since it is pure C and can't benefit from the `ArgList` class)

PR #1190, opened 2021-02-26
Add converter for objects such as Monitor

t-wissmann:
This introduces a new converter class for the parsing of descriptions of
objects, e.g. Monitors, Tags, Clients. So far, only the converter for
Monitor* is implemented. If a command gets a monitor as a parameter, we
can now use ArgParse which internally uses the Converter class.

This new converter is used in the commands remove_monitor and tag_status
and their test cases have slightly been adjusted and extended.

### DIFF
--- a/tests/test_monitor.py
+++ b/tests/test_monitor.py
@@ -100,8 +100,8 @@ def test_remove_monitor_invalid_args(hlwm):
 
 
 def test_cannot_remove_nonexistent_monitor(hlwm):
-    call = hlwm.call_xfail('remove_monitor 1')
-    assert call.stderr == 'remove_monitor: Monitor "1" not found!\n'
+    hlwm.call_xfail('remove_monitor 1') \
+        .expect_stderr('No such monitor: 1')
     assert hlwm.get_attr('monitors.count') == '1'
 
 
--- a/tests/test_monitor.py
+++ b/tests/test_monitor.py
@@ -89,6 +89,8 @@ def test_remove_monitor_invalid_args(hlwm):
     hlwm.call_xfail('remove_monitor') \
         .expect_stderr('Expected one argument, but got only 0')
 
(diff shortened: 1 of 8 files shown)


tests/test_monitor.py
```diff
@@ -85,9 +85,23 @@ def test_remove_monitor(hlwm):
...
+    hlwm.call('add_monitor 800x600+40+40 tag2 monitor2')
+
+    hlwm.call_xfail('remove_monitor foobar') \
+        .expect_stderr('No such monitor: foobar')
+
+    hlwm.call_xfail('remove_monitor 2') \
+        .expect_stderr('No such monitor: 2')
+
+
 def test_cannot_remove_nonexistent_monitor(hlwm):
```
blarz, line 102: Isn't this test a bit redundant now?

blarz (commented):
Just a tiny question regarding tests

tests/test_monitor.py
t-wissmann: Right,.. I've removed it and added asserts in the `test_remove_monitor_invalid_args`

blarz:
How is the ArgParse distribution state btw? Is it used "everywhere"? I think a saw a issue for removing the old version of it?

blarz:
> How is the ArgParse distribution state btw? Is it used "everywhere"? I think a saw a issue for removing the old version of it?

Ah, I confused that one, I meant  #136

tests/test_monitor.py
blarz: Thanks!

blarz (approved):

t-wissmann:
> How is the ArgParse distribution state btw? Is it used "everywhere"? I think a saw a issue for removing the old version of it?

It's not used everywhere but I'd like to do so.

PR #1191, opened 2021-02-27
Run-Time Converter for Client*

t-wissmann:
This implements RunTimeConverter for Client* and uses it in the commands
jumpto and bring which are moved to GlobalCommands.

Since there are multiple variables and functions called 'complete',
RunTimeConverter::complete is renamed to RunTimeConverter::completeEntries
to make the variable names a bit more unique.

The test cases with error handling are slightly extended and moved to
test_global_commands.py, but the existing 'positive' test cases for
bring and jumpto are still in test_clients and test_layout in order to
keep the diff small.

### DIFF
--- a/src/clientmanager.cpp
+++ b/src/clientmanager.cpp
@@ -67,7 +67,7 @@ void ClientManager::injectDependencies(Settings* s, Theme* t, Ewmh* e) {
     ewmh = e;
 }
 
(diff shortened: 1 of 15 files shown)


blarz (approved):
Looks good to me!

t-wissmann:
Thanks!

PR #1192, opened 2021-02-28
layout: Remove unused getGloballyFocusedFrame

blarz:


t-wissmann (approved):
It's a relief that we can get rid of this :-)

PR #1193, opened 2021-03-01
Remove unused lock_number_change return value

blarz:
Change it to void, nobody uses the returned string.

### DIFF
--- a/src/monitormanager.cpp
+++ b/src/monitormanager.cpp
@@ -488,9 +488,9 @@ void MonitorManager::unlock() {
     lock_number_changed();
 }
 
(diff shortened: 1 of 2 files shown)


blarz:
Stumbled over this while looking through the coverage report ;)

t-wissmann (approved):
Thanks!

Actually, the `monitors_locked` should be an unsigned long, such that it is not possible be assigned a negative number.

blarz:
> Thanks!
> 
> Actually, the `monitors_locked` should be an unsigned long, such that it is not possible be assigned a negative number.

I see, so a bit of a redundant check then :D

PR #1194, opened 2021-03-03
Integrate completion into ArgParse

t-wissmann:
This introduces the new type CallOrComplete that combines the invocation
and argument completion of a command.

Effectively,  this will make commands functions with reasonable
parameter types and names and the CallOrComplete-function acts as
glueing code or binding to the user-accessible command interface.

To illustrate the new functionality, it is applied to the commands
bring, focus, shift, tag_status, remove_monitor whose test suite
regarding parse errors and completion is respectively extended.

The word-list suggestion parameter is not yet used, but it will be
useful for commands like cycle_all (completion for -1 and +1) or resize
(completion for +0.02 and -0.02), which will be handled in separate
commits.


src/command.h
```diff
@@ -91,6 +91,36 @@ class CommandBinding {
...
+
+    template <typename ClassName>
+    CommandBinding(ClassName* object,
+                   void(ClassName::*member)(CallOrComplete))
+        : CommandBinding(std::bind(member, object,
+                            std::placeholders::_1))
+    {
+    }
+
     // FIXME: Remove after C++ transition
```
blarz, line 124: Not related to this PR, but Is this possible yet?

blarz (approved):
This is pretty cool (not sure if I understood all of it, but tests still pass, so..)

src/command.h
t-wissmann: There are still a few in main.cpp, but I hope I can migrate them soon.

t-wissmann:
> This is pretty cool (not sure if I understood all of it, but tests still pass, so..)

Thanks :) (The complexity is a real downside, so I'm not sure whether I would use it for all commands, because for complex commands like `rule` it maybe makes things more complicated than necessary)

PR #1195, opened 2021-03-03
Use ArgParse completion in cycle_all

t-wissmann:
Move the cycle_all functionality into a function with typed parameters
that is called by the CallOrComplete-function. This slightly extends the
test cases for error handling and tab-completion of the cycle_all
command.


t-wissmann:
This includes the PR #1194 and thus should be merged afterwards.

blarz (approved):
Looks good to me!

PR #1196, opened 2021-03-03
Make setting monitors_locked an unsigned long

t-wissmann:
The lock counter was always used as a non-negative integer and should
also be such on the type level. In the original settings code, all
integers where signed, but now it is finally updated.

Also it is not necessary anymore to call lock_number_changed() manually,
because this happens automatically through signals.

Both changes are tested by the new test cases for lock/unlock.


blarz (approved):
Nice, thanks!

Issue #838, opened 2020-04-19
Feature Request: ability to replace previous window manager


dario23:
some window managers (i verified for openbox) have a mechanism to replace the previously running window manager using some kind of signalling to the previous window manager. i think the code doing this in openbox is in the [`replace_wm` function](https://github.com/danakj/openbox/blob/master/openbox/screen.c#L90).

i'd be happy to look into this more and/or open a PR if there's nothing in principle against adding such functionality.

t-wissmann:
In principle such a feature is a nice to have for hlwm (and if we implement it we should also be so kind and implement the other part of being replaced by another WM).
I was always wondering how this WM replacement works, and I'm wondering whether this works at all when the WM is the magic process of the X-Session (when the magic process terminals, the X-Session is terminated).

For the scenario of WM replacement while the WM is the magic process, herbstluftwm already provides the `wmexec` command that allows to exec into another window manager (openbox supports this too)

t-wissmann:
Two other points:
1. I'm wondering whether there is a protocol for this WM replacement process
2. If the WM is not the magic process, then you can simply kill it and start hlwm...

dario23:
the [change introducing this functionality into xfwm4](https://git.xfce.org/xfce/xfwm4/commit/?h=xfce-4.4&id=4ed599c097aad946513f46b670b966dce2c1c0e2) mentions ICCCM2 compliance.

roughly that seems to say that

  * window managers should acquire the `WM_S<n>` selection ([sec. 4.3](https://tronche.com/gui/x/icccm/sec-4.html#s-4.3))
  * the protocol around transfering ownership is described in [sec. 2](https://tronche.com/gui/x/icccm/sec-2.html#s-2)

hannahswain:
It would be wonderful to see this happen! This is the recommended command in XFCE's [Other Window Manager](https://wiki.xfce.org/howto/other_window_manager) docs.

dario23:
i am not looking into this actively anymore, since for my use-case (testing if things work in another WM quickly/doing one or two things that work better with openbox) is also possible using [openbox's equivalent to the `wmexec` call](http://openbox.org/wiki/Help:Actions#Restart), i.e. the restart action allows you to specify another executable to restart into.

i still think this would be worthwile to have, as not all other WMs have a wmexec command, but some of them still support the more-or-less standardized `--replace` protocol.

hannahswain:
[...]

PR #1197, opened 2021-03-04
Create EWMH object long before Root object

t-wissmann:
For the implementation of #838, it is beneficial to have the EWMH object
set up much earlier during bootup: 1. in order to have the WM window
available much earlier and 2. to implement the window manager
replacement in the EWMH object where it belongs.

A few new tests verify that the EWMH object is still set up correctly.


blarz (approved):
Seems sensible to me, although I have no knowledge about the whole EWMH stuff

PR #1199, opened 2021-03-04
Migrate pad command to MonitorManager and ArgParse

t-wissmann:
This moves the 'pad' command to MonitorManager and makes use of the
ArgParse utility for argument parsing and completion.

The arguments are a bit nasty because they are either an integer or the
empty string meaning that the pad should not change. This is reflected
by the new class EmptyOrInt, which is only used here and thus in the cpp
file only. Unfortunately, this new class takes a lot of lines, compared
to the "plain C" version before. On the positive side, we now have tab-
completion for the pad values and hopefully robust argument parsing
compared to atoi().

The new test cases hopefully covers all scenarios.


blarz (approved):
Looks good to me!

PR #1200, opened 2021-03-04
Migrate simple monitor commands to MonitorManager

t-wissmann:
This moves the commands

    focus_monitor, cycle_monitor, shift_to_monitor, monitor_rect

to MonitorManager and makes use of ArgParse. The existing test cases for
these commands are extended such that parse errors and tab completion
are also covered.

For these commands, the type-safety and automation of ArgParse comes in
quite handy.


src/main.cpp
```diff
@@ -166,7 +166,7 @@ unique_ptr<CommandTable> commands(shared_ptr<Root> root) {
         {"remove_monitor", { monitors, &MonitorManager::removeMonitorCommand }},
         {"move_monitor",   monitors->byFirstArg(&Monitor::move_cmd, &Monitor::move_complete) } ,
         {"rename_monitor", monitors->byFirstArg(&Monitor::renameCommand, &Monitor::renameComplete) },
-        {"monitor_rect",   monitor_rect_command},
```
blarz, line 169: Remove monitor_rect_command in .cpp file in line 367?

blarz (changes requested):

src/main.cpp
blarz: Or is it still needed?

t-wissmann: Thanks! I simply forgot to remove it. This was probably the reason why the coverage decreased by 0.1%.

blarz: Makes sense :D Looks good now, thanks!

blarz (approved):
Very nice!

PR #1201, opened 2021-03-05
Make frame attributes writable

### LINKED ISSUE #7: object tree: Writable custom attributes
With frame objects, there are some custom attributes (notably `layout`) for which it would make sense to make it writable - but that's currently not possible.

t-wissmann:
The frame attributes of FrameLeaf and FrameSplit are now writable and
allow to adjust the frame locally. As usual, the test cases consume more
lines than the actual source code accomplishing this change.

In the source code, the getTag() function seemed to be unused.
Unfortunately, I had to remove the 'const' qualifier from getFraction()
because otherwise the DynAttribute_ constructor failed to match.

This is another step towards a full implementation of frame objects (#2)
and this fixes #7.


tests/test_layout.py
```diff
@@ -1433,3 +1433,151 @@ def test_focus_shift_completion(hlwm):
...
+        x11.sync_with_hlwm()
+        width1 = win1.get_geometry().width
+        width2 = win2.get_geometry().width
+
+        hlwm.attr.tags.focus.tiling.root[attr] = value
+        assert hlwm.attr.tags.focus.tiling.root[attr]() == value
+        assert hlwm.call('dump').stdout.strip() != layout
+
+        x11.sync_with_hlwm()
+        # both updated attributes must lead to an increas of width of
```
blarz, line 1531: increase

NEWS
```diff
@@ -13,6 +13,8 @@ Current git version
   * Do not draw frame background behind clients (so for semi-transparent
     client decorations, one does not see the frame decoration behind but the
     wallpaper instead)
+  * The frame attributes ('selection', 'algorithm', 'fraction', 'split_type')
```
blarz, line 16: Merge conflicts right now

blarz (changes requested):
Looks good, just one typo and NEWS is having a merge conflict

tests/test_layout.py
t-wissmann: Thanks! Fixed

NEWS
t-wissmann: I've resolved them.

blarz: Thanks!

tests/test_layout.py

blarz (approved):

PR #1202, opened 2021-03-05
Converter for integers checks entire input string

### LINKED ISSUE #944: When calling stoi() we should always check that all characters are read

t-wissmann:
In the Converter for int and unsigned long, the parser now checks that
the entire input string is processed and no unparsable suffix is left at
the end.

This fixes #944.

### DIFF
--- a/src/types.cpp
+++ b/src/types.cpp
@@ -26,7 +26,12 @@ void Converter<Direction>::complete(Completion& complete, const Direction* relat
 template<>
 unsigned long Converter<unsigned long>::parse(const string& source)
 {
-    long value = std::stol(source);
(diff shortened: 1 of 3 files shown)


blarz (approved):
Looks good to me!

PR #1204, opened 2021-03-05
Make 'default_frame_layout' a LayoutAlgorithm setting

### LINKED ISSUE #24: Commands which accept layouts should accept names
Currently commands which take a layout expect a hard to remember integer - we should be able to use layout names instead (or both, to not break backwards compatibility)

t-wissmann:
This makes the setting 'default_frame_layout' have the type
LayoutAlgorithm, so one can now write human-readable names to it rather
than indices. For backwards-compatibility, indices are still accepted
for parsing, but they are neither printed nor completed. The indices are
also removed from the man page to avoid the confusion reported in #24.

On the technical side, the ValueList of a 'Finite<>' type now also holds
a flag that tells whether parsing should interpret integers as indices
in the list of values. This flag is only activated for LayoutAlgorithm
and is 'false' in all other cases.

The test cases are extended accordingly.

This fixes #24.

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -75,9 +75,11 @@ def test_default_frame_layout_value_too_high(hlwm):
         .expect_stderr('set: Invalid value "99" for setting "default_frame_layout": .*out of range')
 
 
-def test_default_frame_layout_value_negative(hlwm):
+def test_default_frame_layout_value_invalid_value(hlwm):
     hlwm.call_xfail('set default_frame_layout -23') \
         .expect_stderr('set: Invalid value "-23" for setting "default_frame_layout": .*Expecting.*vertical')
+    hlwm.call_xfail('set default_frame_layout foobar') \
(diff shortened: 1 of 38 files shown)


blarz:
Yes I love it, this always confused me because I had to look up the number of the max_layout ;)

tests/test_settings.py
```diff
@@ -72,12 +72,61 @@ def test_cycle_value_loop(hlwm):
```
blarz, line 72: Can you add a test for an invalid layout algorithm string? E.g. 
```set default_frame_layout foobar```

blarz: Ah there already is one with ```test_default_frame_layout_value_negative```.
Maybe we can rename the test? It doesn't fit that nicely anymore IMHO?

t-wissmann: I've renamed it. (And I'm testing for `-1` and `foobar`.

blarz: Thanks!

blarz (approved):
Cool!

PR #1205, opened 2021-03-05
Test herbstclient command line parsing

t-wissmann:
This extends the tests for most command line flags of the herbstclient
command. Also, herbstclient now prints an error message if too few
parameters are passed.

### DIFF
--- a/tests/test_herbstclient.py
+++ b/tests/test_herbstclient.py
@@ -152,7 +152,7 @@ def test_version_without_hlwm_running():
 
 def test_quiet_without_hlwm_running():
     for flag in ['-q', '--quiet']:
-        cmd = [HC_PATH, flag, '-q', 'echo', 'test']
+        cmd = [HC_PATH, flag, 'echo', 'test']
         proc = subprocess.run(cmd,
                               stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE,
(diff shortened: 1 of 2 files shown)


tests/test_herbstclient.py
```diff
@@ -139,3 +139,83 @@ def test_lastarg_only(hlwm, repeat):
...
+                              universal_newlines=True)
+        assert proc.returncode != 0
+        assert proc.stdout == ''
+        assert proc.stderr == ''
+
+
+def test_quiet_with_hlwm_running(hlwm):
+    # even with --quiet, herbstclient must produce output
+    for flag in ['-q', '--quiet']:
+        cmd = [HC_PATH, flag, '-q', 'echo', 'test']
```
blarz, line 168: Why is -q in the parameters, shouldn't flag cover this already?

blarz: Same for the test above

blarz (changes requested):

tests/test_herbstclient.py
t-wissmann: I've looked at the coverage report and `-q` was not handled. I think the issue was the wrong parameter (it's fixed by #1206). Also, this case ensures that the stderr is really empty (i.e. that there is no other error).

blarz: Yes, but I thought the for-loop is there for iterating over -q and --quiet?
Now cmd is called with ['-q', '-q'] and then ['--quiet', 'q']. 

Maybe I'm overlooking or misunderstanding something

t-wissmann: Aah, you're completely right. I understood your question entirely wrong (and thus started looking for other test cases where `--quiet` is mentioned...). I've fixed it, sorry for the confusion :)

blarz: Thanks! But a nice side-effect of uncovering a different testing error :D

blarz (approved):

PR #1206, opened 2021-03-05
Fix herbstclient test case

t-wissmann:
In the case of hc_quiet, herbstclient was invoked with one parameter
"--quiet true" instead of with two parameters "--quiet" and "true".


blarz (approved):
Makes sense!

PR #1207, opened 2021-03-05
Remove unused functions from ArgList and Attribute

t-wissmann:
The code coverage showed that these are not used.


blarz (approved):
Bye bye!

PR #1209, opened 2021-03-05
Option --replace to replace existing window manager

t-wissmann:
Herbstluftwm now uses the ICCCM-specified way of checking whether there
is an existing window manager, namely by checking the ownership
particular 'selection'. The owner of this selection is the active window
manager. This change introduces the new features:

    1. herbstluftwm does not start if the selection is already owned
    2. if herbstluftwm is started with --replace, try to become the
       owner of the selection
    3. if herbstluftwm looses the ownership, then quit, because there is
       some other window manager booting up.

In the test case, we only check that herbstluftwm can replace itself,
but I've tested it locally and the replacement works (forth and back)
with xfwm4, openbox, fvwm2, pekwm.

This change also adds a test case for the 'quit' command because it is
essentially half the test of '--replace'.

This implements #838.

### DIFF
--- a/src/ewmh.cpp
+++ b/src/ewmh.cpp
@@ -2,9 +2,9 @@
 
 #include <X11/Xatom.h>
 #include <X11/Xlib.h>
+#include <unistd.h>
 #include <algorithm>
 #include <cstdio>
-#include <unistd.h>
 
 #include "client.h"
 #include "globals.h"
--- a/src/ewmh.cpp
+++ b/src/ewmh.cpp
@@ -4,6 +4,7 @@
 #include <X11/Xlib.h>
 #include <algorithm>
 #include <cstdio>
+#include <unistd.h>
 
 #include "client.h"
 #include "globals.h"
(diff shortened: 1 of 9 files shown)


src/ewmh.cpp
```diff
@@ -147,6 +154,60 @@ long Ewmh::windowGetInitialDesktop(Window win)
...
+        {
+            // if there was an error in XChangeWindowAttributes, then
+            // the current_wm disappeared in the mean time.
+            current_wm = 0;
+        }
+    }
+    // make ourselves the owner of the selection.
+    if (!XSetSelectionOwner(X_.display(), wmSnAtom, windowManagerWindow_, CurrentTime))
+    {
+        HSDebug("Can not aquire ownership on screen %d\n", X_.screen());
```
blarz, line 180: Typo: acquire

t-wissmann: Fixed

blarz: Thanks!

blarz (approved):
Very nice!

PR #1210, opened 2021-03-07
Manage desktop windows in global stack

t-wissmann:
Instead of lowering desktop windows when they appear, manage desktop
windows in the global stack explicitly. Unfortunately, I did not find a
way to query the window stack. So for debugging and testing the output
of the `stack` command is extended by the desktop windows' winid and
window instance/class.

This solves an issue here where the Xfce's desktop window covered all
normal windows if I run `herbstluftwm --replace` in a running Xfce
session.

### DIFF
--- a/src/desktopwindow.h
+++ b/src/desktopwindow.h
@@ -2,6 +2,7 @@
 #define __HERBSTLUFT_DESKTOPWINDOW_H_
 
 #include <X11/X.h>
+#include <functional>
 #include <memory>
 #include <vector>
 
@@ -18,6 +19,7 @@ public:
     static void registerDesktop(Window win);
     static void lowerDesktopWindows();
     static void unregisterDesktop(Window win);
+    static void foreachDesktopWindow(std::function<void(DesktopWindow&)> loopbody);
 private:
 
     // members:
--- a/src/desktopwindow.h
+++ b/src/desktopwindow.h
@@ -17,7 +17,6 @@ public:
     bool below() const;
     Window window() const;
     static void registerDesktop(Window win);
-    static void lowerDesktopWindows();
     static void unregisterDesktop(Window win);
     static void foreachDesktopWindow(std::function<void(DesktopWindow&)> loopbody);
 private:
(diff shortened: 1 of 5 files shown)


tests/test_stack.py
```diff
@@ -191,10 +191,37 @@ def test_stack_tree(hlwm):
...
+      - Floating-Layer
+      - Tiling-Layer
+      - Frame Layer
+        - Window {framewindows[0]}
+    - Desktop Windows
+      - {winid} TestDesktop TestDeskClass
+'''
+    stack = hlwm.call('stack')
+    print("expected:")
+    print(expected_stack)
```
blarz, line 221: Do you want to keep those print statements?
The assert statement should print the diff between expected and actual anyway, right?

src/desktopwindow.h
```diff
@@ -18,6 +19,7 @@ class DesktopWindow {
     static void registerDesktop(Window win);
     static void lowerDesktopWindows();
```
blarz, line 20: Can be removed?

blarz (changes requested):

tests/test_stack.py
t-wissmann: I forgot to remove the print-statement, but I had added it because the diff wasn't helpful at all. During writing, I forgot to apply `x11.winid_str(w)` in line 206 and so the content of 'Frame Layer' was wrong. The output then was:
```
>       assert stack.stdout == expected_stack
E       assert '  -\n    - M...stDeskClass\n' == '  -\n    - M...stDeskClass\n'
E             -
E               - Monitor 0 with tag "default"
E                 - Focus-Layer
E                 - Fullscreen-Layer
E                 - Floating-Layer
E                 - Tiling-Layer
E                 - Frame Layer...
E         
E         ...Full output truncated (6 lines hidden), use '-vv' to show

tests/test_stack.py:220: AssertionError
```
Do I have to activate something to obtain a proper diff?

Anyway, I have removed the print-statement, thanks for spotting :)

src/desktopwindow.h
t-wissmann: Yes :) Thanks!

blarz: Thanks!

tests/test_stack.py
blarz: The full diff only comes when you run pytest with -vv.
What I usually do, when a test fails I run it explicitly with  ```pytest -vv test_foo.py:test_foo```

blarz (approved):
Thanks!

PR #1211, opened 2021-03-07
Migrate 'raise' command to GlobalCommands and ArgParse

t-wissmann:
The raise command expects as the parameter either a client or the window
id of an unmanaged window. This is reflected by the type
Either<Client*,WindowID> for the new template class Either. Because of
the genericity of the Converter of Either, the error message might be a
bit cryptic.


blarz (approved):
The either thing looks pretty cool

PR #1212, opened 2021-03-08
cycle_monitor: Make argument optional

The-Compiler:
The manpage says:

    cycle_monitor [DELTA]
        Cycles monitor focused by DELTA.  DELTA defaults to 1.

but after 28ea7f66ac9a889827f61206cdbc0743dd2e63c5, `hc cycle_monitor` says:

    cycle_monitor: Expected one argument, but got only 0 arguments.
    cycle_monitor: not enough arguments

because in #1200, it was accidentally marked as mandatory rather than optional.

### DIFF
--- a/tests/test_monitor.py
+++ b/tests/test_monitor.py
@@ -367,9 +367,16 @@ def test_list_padding_invalid_monitor(hlwm):
 
 @pytest.mark.parametrize("mon_num,focus_idx", [
     (num, focus) for num in [1, 2, 3, 4, 5] for focus in [0, num - 1]])
-@pytest.mark.parametrize("delta", ['-1', '+1'])
-@pytest.mark.parametrize("command", ['cycle_monitor', 'focus_monitor'])
-def test_cycle_monitor(hlwm, mon_num, focus_idx, delta, command):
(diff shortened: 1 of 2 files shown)


tests/test_monitor.py
```diff
@@ -367,9 +367,16 @@ def test_list_padding_invalid_monitor(hlwm):
 
 @pytest.mark.parametrize("mon_num,focus_idx", [
     (num, focus) for num in [1, 2, 3, 4, 5] for focus in [0, num - 1]])
-@pytest.mark.parametrize("delta", ['-1', '+1'])
-@pytest.mark.parametrize("command", ['cycle_monitor', 'focus_monitor'])
-def test_cycle_monitor(hlwm, mon_num, focus_idx, delta, command):
+@pytest.mark.parametrize("command, delta", [
+    ('cycle_monitor', '-1'),
+    ('cycle_monitor', '+1'),
+    ('cycle_monitor', '1'),
```
The-Compiler, line 373: This is another new testcase, albeit unrelated to the issue I think it makes sense to test this too.

t-wissmann (approved):
Nice, thanks!

PR #1214, opened 2021-03-08
Objects for panels

t-wissmann:
This makes PanelManager an object containing children for all panels. A
panel's window attribute is the first attribute to use the type WindowID
(whose converter has been used before already).

The documentation of a class is only included in the man page if there
is a path that leads to such an object. For panels, we can't provide
such a path in general. As a fix, instead of the 'name' of an entry in
an object, the json doc now mentions a 'name_pattern'. This
'name_pattern' can be put into the documentation, but the test cases in
test_doc.py must not treat this name literally.

### DIFF
--- a/src/panelmanager.cpp
+++ b/src/panelmanager.cpp
@@ -8,12 +8,33 @@ using std::make_pair;
 using std::string;
 using std::vector;
 
-class Panel {
+class Panel : public Object {
 public:
-    Panel(Window winid, PanelManager& pm) : winid_(winid), pm_(pm) {}
-    Window winid_;
+    Panel(Window winid, PanelManager& pm)
+        : winid_(this, "winid", winid)
+        , windowInstance_(this, "instance", {})
+        , windowClass_(this, "class", {})
+        , pm_(pm)
+        , size_(this, "geometry", {})
+    {
+        auto hint = pm.xcon_.getClassHint(winid);
+        windowInstance_ = hint.first;
+        windowClass_ = hint.second;
+        winid_.setDoc("the ID of the panel window");
+        windowInstance_.setDoc("the window instance (first entry of WM_CLASS)");
+        windowClass_.setDoc("the window class (second entry of WM_CLASS)");
+        size_.setDoc("the size and position of the window");
+        setDoc("a panel is an unmanaged window that reserves space at "
+               "the edge of the monitor it is on. The space depends on "
+               "the _NET_WM_STRUT defined by the panel. If it is however "
+               "not defined explicitly, then the amount of reserved space "
+               "is infered from the window geometry.");
+    }
+    Attribute_<WindowID> winid_;
+    Attribute_<string> windowInstance_;
+    Attribute_<string> windowClass_;
     PanelManager& pm_;
-    Rectangle size_;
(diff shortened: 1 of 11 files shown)


t-wissmann:
Hm. I don't get why there is this error message
```
../src/panelmanager.h: In member function ‘long unsigned int PanelManager::getCount()’:
../src/panelmanager.h:51:57: error: useless cast to type ‘long unsigned int’ [-Werror=useless-cast]
   51 |         return static_cast<unsigned long>(panels_.size());
      |                       
```
for this function:
```C++
    unsigned long getCount() {
        return static_cast<unsigned long>(panels_.size());
    };
```
In `indexingobject.h` there is the same function, but it does not trigger this error:
```C++
    unsigned long sizeUnsignedLong() {
        return static_cast<unsigned long>(data.size());
    }
```

blarz:
> Hm. I don't get why there is this error message
> 
> ```
> ../src/panelmanager.h: In member function ‘long unsigned int PanelManager::getCount()’:
> ../src/panelmanager.h:51:57: error: useless cast to type ‘long unsigned int’ [-Werror=useless-cast]
>    51 |         return static_cast<unsigned long>(panels_.size());
>       |                       
> ```
> 
> for this function:
> 
> ```c++
>     unsigned long getCount() {
>         return static_cast<unsigned long>(panels_.size());
>     };
> ```
> 
> In `indexingobject.h` there is the same function, but it does not trigger this error:
> 
> ```c++
>     unsigned long sizeUnsignedLong() {
>         return static_cast<unsigned long>(data.size());
>     }
> ```

That is indeed odd. Maybe the template in indexingobject makes gcc not look any closer?
(Removing ```-Werror=useless-cast```?)

src/panelmanager.cpp
```diff
@@ -8,12 +8,33 @@ using std::make_pair;
...
+        windowClass_ = hint.second;
+        winid_.setDoc("the ID of the panel window");
+        windowInstance_.setDoc("the window instance (first entry of WM_CLASS)");
+        windowClass_.setDoc("the window class (second entry of WM_CLASS)");
+        size_.setDoc("the size and position of the window");
+        setDoc("a panel is an unmanaged window that reserves space at "
+               "the edge of the monitor it is on. The space depends on "
+               "the _NET_WM_STRUT defined by the panel. If it is however "
+               "not defined explicitly, then the amount of reserved space "
+               "is infered from the window geometry.");
```
blarz, line 31: typo: inferred

tests/test_doc.py
```diff
@@ -38,6 +38,22 @@ def create_tag_with_all_links(hlwm):
     return 'tags.0'
 
 
+def create_panel(hlwm):
+    """
+    Connect to the display; we only need x11 here, so
+    let us not introduce it as a dependency for the all
+    test cases but instead connect to the X display manually.
+
+    Unfortunately, we never have the chance to proplerly close
```
blarz, line 47: typo: properly

blarz (changes requested):
Looks good, tested with xfce4-panel. Just two typos and this is good to go

blarz:
> Thanks! I've fixed both typos.

Thanks!

blarz (approved):

PR #1215, opened 2021-03-09
Remove -Wuseless-cast

t-wissmann:
The issue is that we have to make casts from platform dependently sized
types such as size_t to platform independent types such as unsigned
long. Such a cast is sometimes useless and sometimes important,
depending on the platform, so we can't omit it.

This should make #1214 build again which contained such a "useless" cast.


blarz (approved):
Maybe not the most useful compiler warning, agreed

PR #1216, opened 2021-03-09
Add 'lower' command

t-wissmann:
Analogously to the 'raise' command, there should be a 'lower' command
that lowers a managed or unmanaged window in the stack. For example,
this can be used to manually lower a desktop window.

Fortunately, the test cases for 'raise' could either extended or easily
adapted to the 'lower' command.

### DIFF
--- a/tests/test_meta_commands.py
+++ b/tests/test_meta_commands.py
@@ -556,7 +556,10 @@ def test_raise_lower_winid_missing(hlwm):
         .expect_stderr('lower: not enough arguments\n')
 
 
-def test_raise_invalid_winid(hlwm):
+def test_raise_lower_invalid_winid(hlwm):
+    hlwm.call_xfail('raise foobar') \
+        .expect_stderr('Invalid format, expecting')
+
     hlwm.call_xfail('lower foobar') \
         .expect_stderr('Invalid format, expecting')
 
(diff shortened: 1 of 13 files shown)


tests/test_meta_commands.py
```diff
@@ -548,13 +548,16 @@ def test_integer_out_of_range(hlwm):
...
-def test_raise_winid_missing(hlwm):
+def test_raise_lower_winid_missing(hlwm):
     hlwm.call_xfail('raise') \
         .expect_stderr('raise: not enough arguments\n')
 
+    hlwm.call_xfail('lower') \
+        .expect_stderr('lower: not enough arguments\n')
+
 
 def test_raise_invalid_winid(hlwm):
```
blarz, line 559: Shouldn't we test both cases in here? You replaced raise with lower, so raise is not tested with an invalid winid

blarz (changes requested):

tests/test_meta_commands.py
t-wissmann: Oh, right, thanks! I've fixed it.

blarz (dismissed):
Looks good to me (needs some resolving before it can be merged)

blarz (approved):

PR #1217, opened 2021-03-09
Avoid global variables and use XConnection instead

t-wissmann:
This avoids the usage of g_root, g_display, g_screen at many places and
instead uses XConnection, if easily accessible. Though g_display and
g_root are still used a lot, the present change at least allows the
removal of g_screen.


blarz (approved):
Cool!

Issue #1203, opened 2021-03-05
Make set/get/cycle_value compatible with attributes


t-wissmann:
There is no attribute-command that does what `cycle_value` does for settings. Instead of coming up with a new command name, I think it is more user-friendly to just make `cycle_value` also work for attributes (and maybe even `set` and `get` because these are much better names than `set_attr`/`get_attr`)

PR #1218, opened 2021-03-09
Make 'cycle_value' work with attributes

t-wissmann:
The 'cycle_value' command now expects the path of an attribute instead
of a settings name and only falls back to the old behaviour if no
matching attribute is found.

This is a first (and in fact most important) step of #1203.


blarz (approved):
Looks good to me!

PR #1221, opened 2021-03-10
Rename types.h (and .cpp) -> converter.h (and .cpp)

t-wissmann:
The types.h file mainly contains the Converter definition, and has
nothing to do with the Type enum (which is defined in entity.h). Thus, I
think it is more appropriate to call the file converter.h.


blarz (approved):
Makes sense!

Issue #599, opened 2020-01-02
Herbstluftwm and lightdm


arcolinuxz:
ArcoLinux uses lightdm as our standard display manager on 15 desktops.

We were wondering if you can help with the strange "look" when we start up herbstluftwm and how to overcome it.

**After login we see a black screen with a red border.** 

Can we get rid of that somehow? 
We do want to keep lightdm as display manager.

Check video at 40:46
https://www.youtube.com/watch?v=RzA8zgbq-nE

bazeeel:
yes i have that too with lightdm but ...if i start the herbstluftwm inside .xinitrc  then no more that strange black frame/window with red borders.

arcolinuxz:
Could you please assist?

bl00dy1837:
I don't know lightdm, I use .xinitrc.

For testing press your Mod Key (default ALT) + Return. Then should xterm start.

Read: https://herbstluftwm.org/tutorial.html or a tutorial as video https://www.youtube.com/watch?v=81fL98ZN7_g .

bazeeel:
> Could you please assist?

if not exist just create .xinitrc 
file
inside write exec herbsluftwm --locked

arcolinuxz:
> For testing press your Mod Key (default ALT) + Return. Then should xterm start.

Why do I need to start up xterm?
Do do what?

bl00dy1837:
> > For testing press your Mod Key (default ALT) + Return. Then should xterm start.
> 
> Why do I need to start up xterm?
> To do what?

Only to test if herbstluftwm works. After starting herbstluftwm default config there is only a black screen.

herbstluftwm is a tilling manager. there is no desktop/application menu. You need to define it by yourselfe.

I use as topbar "demu2" and to start applications "rofi".

tiosgz:
Sorry for responding so late… is this still actual for you?

The black screen is the main frame before you unlock HLWM: https://github.com/herbstluftwm/herbstluftwm/blob/78be1fa7d934fb700ce576f3c828aa82f71241cb/share/autostart#L157

You could maybe reduce the time partly by rearranging the autostart so that only visual changes happen before unlocking, & by `chain`ing the commands (but that may reduce readability if you aren’t careful — which is something you care about, do I remember correctly?)

I don’t recommend unlocking immediately (or not locking at all), coz, well, it usually reduces performance (see `lock`/`unlock` in the manpage).

Barbarossa93:
[...]

PR #1222, opened 2021-03-11
Start --locked from the .desktop file

tiosgz:
As noticed by @Barbarossa93 in #599, this isn't the default although
the default autostart does `unlock`

### DIFF
--- a/share/herbstluftwm.desktop
+++ b/share/herbstluftwm.desktop
@@ -2,5 +2,5 @@
 Encoding=UTF-8
 Name=herbstluftwm
 Comment=Manual tiling window manager
-Exec=herbstluftwm
+Exec=herbstluftwm --locked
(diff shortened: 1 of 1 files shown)


tiosgz:
Alternatively it could be added at the beginning of the autostart to make the `lock`s & `unlock`s balanced even when re-running the autostart, but that could cause problems to those who use `--locked` as well.

t-wissmann (approved):
Thanks! I actually planned to get rid of `--locked` in the long run by automatically locking & unlocking when the autostart is running. But until that, it's fine to add it I think!

As you already mentioned, it is quite hard to keep the `lock` and `unlock` balanced (e.g. if you run `reload` then it is not locked at the moment, but I think this should be solved with the automatic locking when it is implemented..)

PR #1224, opened 2021-03-11
Migrate use, use_index to ArgParse

t-wissmann:
This makes use of ArgParse in the commands 'use' and 'use_index', and
makes them part of the GlobalCommands class. They both need access to
TagManager and MonitorManager, so they don't really belong to neither of
those classes. Since they modify the global state, they fit quite well
in GlobalCommands.

For the implementation, TagManager now provides a run-time converter
for tags, but it only can be used for 'use'.

Also, this adds test cases for both commands and thus also for the
run-time converter.


t-wissmann:
The tests should be fixed as soon as #1226 is merged.

tests/test_global_commands.py
```diff
@@ -134,3 +137,83 @@ def test_cycle_value_invalid_arg(hlwm):
     # calling it a second time fails
     hlwm.call_xfail(command) \
         .expect_stderr('Invalid value for "clients.my_foo": ')
+
+
+def test_use_tag(hlwm):
```
blarz, line 142: There is a similar test in test_tags.py. Should we remove it?

blarz (changes requested):

tests/test_global_commands.py
t-wissmann: I've merged the test cases into one the test case `test_use_tag` in `test_global_commands.py`

blarz: Thanks!

blarz (approved):

PR #1227, opened 2021-03-11
Migrate tag commands to argparse

t-wissmann:
This migrates the commands 'move', 'move_index', and 'rename' to
ArgParse. Both commands have test cases already.


t-wissmann:
This PR requires `Converter<HSTag*>`, so I had to merge #1224 into this, and so this PR also requires the posix-escaping PR (#1226) and #1224 to be merged first.

blarz (approved):

PR #1228, opened 2021-03-11
Add doc on the syntax of fonts

t-wissmann:
This just describes the two ways of specifying a font.

There are three parsers used in fontdata.cpp, but the third one
(XLoadQueryFont) also uses XLFD and only is a fallback if the
XCreateFontSet() does not work, but I don't think it is worth
mentioning that in the user documentation.


src/typesdoc.cpp
```diff
@@ -64,6 +64,22 @@ TypesDoc::TypesDoc()
...
+         "A font specification (font family with modifiers regarding size, "
+         "weight, etc.) in one of the following formats:\n"
+         "\n"
+         "- Fontconfig description. This supports antialiased fonts,\n"
+         "  for example:\n"
+         "  * 'Dejavu Sans:pixelsize=12'\n"
+         "  * 'Bitstream Vera Sans:size=12:bold'\n"
+         "\n"
+         "- X logical font description (XLFD), as provided by the\n"
+         "  xfontsel tool. No antialiasing is supported, here but this\n"
```
blarz, line 78: Shift the comma?
"No antialiasing is supported here, but..."

blarz (changes requested):

src/typesdoc.cpp
t-wissmann: Yes! it's fixed.

blarz: Thanks!

blarz (approved):

PR #1229, opened 2021-03-11
Fix doc/gendoc.py extractor to strip comments

t-wissmann:
Formerly, the doc/gendoc.py keept the comment tokens in the token tree.
This resulted in a comment being treated as the function body of a
function definition and hence some of the documentation was missed.

Also, this was not noticed because the classname2examplepath array was
not extended in test_doc.py. The present change fixes the main issue and
the test case, and even makes it a bit stricter.

### DIFF
--- a/doc/gendoc.py
+++ b/doc/gendoc.py
@@ -23,6 +23,10 @@ class TokenRe:
 
 
 def extract_file_tokens(filepath):
+    """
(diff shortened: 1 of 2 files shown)


t-wissmann:
I've self-approvied it to see the positive effect on the other PRs #1230 and #1228.

PR #1230, opened 2021-03-11
Add types doc for the 'names' type

t-wissmann:
There is not much to say here, because all Finite<> types
use 'names' / 'n' as the attribute type, e.g. LayoutAlgorithm
or SplitType.


t-wissmann:
Makes only sense after #1229 is merged.

blarz (approved):

blarz (approved):

PR #1234, opened 2021-03-12
Migrate merge_tag to ArgParse

t-wissmann:
This migrates 'merge_tag' to ArgParse and adds a few more test cases.


t-wissmann:
The diff includes #1233, which should be merged first.

src/tagmanager.cpp
```diff
@@ -127,31 +128,26 @@ int TagManager::tag_add_command(Input input, Output output) {
...
-    if (!tagToRemove) {
-        output << input.command() << ": Tag \"" << tagNameToRemove << "\" not found\n";
-        return HERBST_INVALID_ARGUMENT;
-    }
-
-    if (!targetTag) {
-        output << input.command() << ": Tag \"" << targetTagName << "\" not found\n";
-        return HERBST_INVALID_ARGUMENT;
-    }
+void TagManager::removeTagCommand(CallOrComplete invoc) {
```
blarz, line 131: Why is the method called removeTagCommand when the command is actually called merge tag?

t-wissmann: This happend to be historically, but you're right, I should rename it to something like  `mergeTagCommand`, shouldn't I?

blarz: It would make sense to do it now, while you're moving stuff in here anyway, yes :)

I would approve of a renaming. It confused me quite a bit, while reading the commit message subject
and thinking "wait a minute, nothing in the diff says "merge" :D

t-wissmann: (I've renamed it)

blarz: Cool! Thank you :)

blarz (dismissed):

blarz:
Needs another conflict resolving. What a merge fest today :D

t-wissmann:
> Needs another conflict resolving. What a merge fest today :D

:) It's resolved.

blarz (approved):
Let's... merge this :D

Issue #1235, opened 2021-03-12
HLWM conflicts with REAPER's floating dialogue windows


theotrst:
Hey, 
I got some problems (i.e. system crashes and I have to restart manually) when I start REAPER (Audio Editor, www.reaper.fm/) when I open any REAPER's dialog floating windows. I don't know why only with this program I have this issue (maybe someone has found a solution for this). But I can solve this by setting a rule. I know it is possible to create a rule to selectively disable floating dialogs windows by default. 

Should I add somthing like

> hc disable attr clients.focus.floating toggle

but how to make this rule only for a single program?

many thanks,
R

t-wissmann:
> I got some problems (i.e. system crashes and I have to restart manually) when I start REAPER (Audio Editor, [www.reaper.fm/](http://www.reaper.fm/)) when I open any REAPER's dialog floating windows. I don't know why only with this program I have this issue (maybe someone has found a solution for this).

Does herbstluftwm crash? What version are you using? (Output of `herbstclient version` or `herbstluftwm --version`). Can you run the following and reproduce the crash in Xephyr? (Ctrl+Shift grabs mouse & keyboard)
```
Xephyr -resizeable :13 & sleep 2; DISPLAY=:13 herbstluftwm --verbose
```

> But I can solve this by setting a rule. I know it is possible to create a rule to selectively disable floating dialogs windows by default.
> 
> Should I add somthing like
> 
> > hc disable attr clients.focus.floating toggle
> 
> but how to make this rule only for a single program?

First you need to find out the window class of reaper. you can do this by running
```
xprop | grep 'WM_CLASS'
```
and then click on the reaper window. The window class is the *second* entry of WM_CLASS. If it is for example `Reaper` you can add the rule:
```
hc rule class=Reaper floating=false
```
This disables the floating-setting for the reaper window. However, I doubt that this will solve any crashes.

theotrst:
Thank you for your answer.
I run the 0.8.3 version

I tried to run REAPER within Xephyr: I opened Reaper, and try to open a dialog windows (usually I try ABOUT). I try to make a photo of Xephyr's report:

> 
> ~\ >>> hc version
> herbstluftwm 0.8.3
> Copyright (c) 2011-2014 Thorsten Wißmann
> Released under the Simplified BSD License
> xrandr support: on
> xinerama support: on
[...]

PR #1236, opened 2021-03-13
Only resend ConfigureNotify on geometry change

t-wissmann:
This fixes an issue (#1235) which consisted of hlwm hanging because of a
loop between hlwm and the reaper application (reaper.fm).

Apparently, the issue is that `XA_WM_NORMAL_HINTS` is updated by the
reaper window, and this causes the tiling layout to be re-computed,
which results in a ConfigureNotify sent to the window. This event seems
to make reaper set `XA_WM_NORMAL_HINTS` again and so the whole process
starts again.

The solution is to only send the ConfigureNotify if the geometry really
changed. I don't know how to write a test case for it, but any test case
more or less should check that this still sends enough ConfigureNotify
events as requested by ICCCM (section 4.1.5.).

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -225,6 +225,7 @@ void Decoration::resize_outline(Rectangle outline, const DecorationScheme& schem
         // right when the window size changed.
         outline = scheme.inner_rect_to_outline(inner);
     }
+    bool windowGeometryChanged = inner != last_inner_rect;
     last_inner_rect = inner;
     inner.x -= outline.x;
     inner.y -= outline.y;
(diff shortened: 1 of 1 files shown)


blarz:
Love the branch name :D :notes:

t-wissmann:
I think we should merge it and see whether it causes other issues, and then react.

blarz:
> I think we should merge it and see whether it causes other issues, and then react.

I wasn't sure on how to go forward with this PR and was just about to ask. I'll approve it then :)

blarz (approved):
All our times have come

Issue #1239, opened 2021-03-16
Bugged position of menus in Qt apps


tiosgz:
This seems to be new with #1236.

Qt apps use an outdated menu position if the window is floating. Example:

- set tag to floating
- open VLC
- open a menu to check that it’s placed correctly; close the menu
- move the window
- check the menu again – it’s misplaced
- set tag to tiling
- check the menu again – it’s okay
- split the tag & move the window to the other frame
- the menu is still okay

blarz:
I can confirm. The menu is still at the old position after moving the floating VLC window.

blarz:
Resizing the window however fixes the menu position again.

blarz:
And it really was introduced with #1236 as @tiosgz said. I reverted to HEAD~2 and this doesn't happen.
(not sure how to fix it though)

t-wissmann:
I have it on my radar and will look at it on the weekend.

PR #1240, opened 2021-03-21
Remember last reported geometry in client directly

t-wissmann:
Instead of deciding in Decoration when to send a configure, just
remember the last configure event in every client so we know directly
whether it is worth sending a new configure event or not.

This fixes the bug #1239 where moving a client with the mouse did not
send the configure event when the mouse dragging was over. (This caused
the menus in Qt applications to be positioned wrong.). The bug was
originally introduced by #1236, thanks to @tiosgz for reporting it!

For easier testing, this remembered window geometry is exposed via the
attribute system. To simplify writing tests for it, there is a new type
'Rectangle' in the python bindings.

### DIFF
--- a/python/herbstluftwm/types.py
+++ b/python/herbstluftwm/types.py
@@ -1,0 +1,50 @@
+import re
+"""
+Module containing types used in the communication with herbstluftwm;
+primarily, types used in attributes.
+"""
+
+
+class Rectangle:
+    """
+    A rectangle on the screen, defined by its size and its distance to
+    the top left screen corner.
+    """
+    def __init__(self, x=0, y=0, width=0, height=0):
+        self.x = x
+        self.y = y
+        self.width = width
+        self.height = height
+
+    def __str__(self):
+        return f'Rectangle({self.x}, {self.y}, {self.width}, {self.height})'
+
+    def __repr__(self):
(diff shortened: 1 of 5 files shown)


blarz:
Definitely fixes  #1239 for me here, is the bug, thanks!
Did you verify that #1235 is also still fixed?

t-wissmann:
> Definitely fixes #1239 for me here, is the bug, thanks!
> Did you verify that #1235 is also still fixed?

No, I didn't check yet because I didn't want to install reaper on my laptop.

tiosgz:
Thank you ❤

blarz:
> > Definitely fixes #1239 for me here, is the bug, thanks!
> > Did you verify that #1235 is also still fixed?
> 
> No, I didn't check yet because I didn't want to install reaper on my laptop.

I just downloaded it and started it (without installing it) and opened some random dialog.
If that is what already triggered the bug in #1235 I don't see it and this change is good to go I'd say :)

blarz (approved):

t-wissmann:
Thanks for testing it!

theotrst:
I tested Reaper now (building the git-version)... it seems this fixes the problem!!!

t-wissmann:
Thanks for checking! :-)

PR #1241, opened 2021-03-21
Move pointer grabbing to XMainLoop via signals

t-wissmann:
XMainLoop cares about the Xlib event queue, so it is natural to do all
queue operations there (e.g. discarding things from the queue).
Moreover, this helps separating the 'hlwm logic' from the X11-specific
code.

So, instead of grabbing the pointer in MouseManager, only adjust the
link clients.dragged and make XMainLoop react to a change of
clients.dragged via signals.

There is already a test suite on the mouse grabbing. As a bonus, this
removes four usages of the global variable g_display.


blarz (approved):
Makes sense to me!

PR #1242, opened 2021-03-21
Fix types doc on fonts

t-wissmann:
There are two issues in the documentation on the font type that are
fixed by adjusting the documentation that comes out of the json file:

  - nested bullet items are now indented correctly.
  - everything within single-quotes is now treated as-is, so no
    special characters can be interpreted by asciidoc.

The diff on the man page is the following. Note how the indentation and
the asterisks of the example fonts are correct now.

```
            •   Fontconfig description. This supports antialiased
                fonts, for example:

-       •   Dejavu Sans:pixelsize=12
+               •   Dejavu Sans:pixelsize=12

-       •   Bitstream Vera Sans:size=12:bold
+               •   Bitstream Vera Sans:size=12:bold

            •   X logical font description (XLFD), as provided by the
                xfontsel tool. No antialiasing is supported here, but
                this is usually superior for bitmap fonts. For
                example:

-       •   --fixed-medium-r---13------- for a standard bitmap font
-           available on most systems. For attributes and children,
-           see types.bool
+               •   -*-fixed-medium-r-*-*-13-*-*-*-*-*-*-* for a
+                   standard bitmap font available on most systems.
+                   For attributes and children, see types.bool

        •   int: Type representing signed integers. For attributes
            and children, see types.bool
```


blarz (approved):
Looks good and doesn't seem to break anything else
(but I only skimmed over the rest of the documentation)

PR #1243, opened 2021-03-21
Add types doc for regex and window id

t-wissmann:
This documents the syntax of the remaining two attribute types.


src/typesdoc.cpp
```diff
@@ -104,11 +104,33 @@ TypesDoc::TypesDoc()
...
     uint_.init(Type::ULONG);
     uint_.setDoc("Type representing unsigned (i.e. non-negative) integers");
+
+    windowid_.init(Type::WINDOW);
+    windowid_.setDoc(
+        "The window id is the number of a window. This can be a managed window "
+        "(i.e. client) or an unmanaged window (e.g. a panel, a menu, "
+        "or a desktop window).\n"
+        "The default format is 0xHEX where HEX is a hexadecimal number "
+        "(digits 0-9 and a-z) but it can also be specified in the decimal "
```
blarz, line 129: This should be a-f

blarz (changes requested):

src/typesdoc.cpp
t-wissmann: Oh, absolutely! Thanks, it's fixed now :)

blarz: Thanks!

blarz (approved):

PR #1244, opened 2021-03-24
Always reply to a ConfigureRequest with ConfigureNotify

t-wissmann:
Depending on the context, it is mandatory to send back a configure
request, even if the window geometry did not change.

This fixes the issue that gnome-terminal and xfce4-terminal are blank on
start up, because they are waiting for the ConfigureNotify.

The bug was introduced by #1240 which is part of a sequence of changes
regarding issue #1235. The present change is just a quick fix, in the
long run this should be solved more elegantly.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -432,9 +432,11 @@ void Client::updatesizehints() {
 
 
 
-void Client::send_configure() {
+void Client::send_configure(bool force) {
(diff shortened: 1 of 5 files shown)


blarz (approved):
Thanks for the quick fix, fixes my issues here

PR #1247, opened 2021-03-26
Test that smart placement uses the decoration

t-wissmann:
Another test case that the smart placement really uses the decoration's
geometry (i.e. 'outer geometry') for the smart placement algorithm. I've
nearly introduced an issue related to this in another change, so it is
good to have it tested explicitly.

### DIFF
--- a/tests/test_rules.py
+++ b/tests/test_rules.py
@@ -1,5 +1,6 @@
 import pytest
 import re
+from herbstluftwm.types import Rectangle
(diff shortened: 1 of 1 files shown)


blarz (approved):
Test looks good, thanks!

t-wissmann:
Thanks :) Now this hopefully shows that I didn't miss anything in #1248 :)

PR #1248, opened 2021-03-26
Only emit Attribute_::changed() on real change

t-wissmann:
The signal changed() present for all attributes was always emitted when
the attribute was written, even if the same value is assigned to it.
This issue only existed for direct assignments (operator=) in the source
code. For re-assignments via the attribute commands, the signal was only
emitted if the value really changed.

Changing the semantics of the signal has of course much impact, but when
using it so far, I always thought it has the 'only on change' semantics
(just as the name suggests). Moreover, it makes sense because we can
then use signals in the X11-modules to automatically forward information
to X11 whenever attributes (properly) change.

The present change uncovered a bug: in the smart placement, the variable
Decoration::last_scheme was used implicitly, even though it was only
initialized by accident (one of the signals called applyLayout()). Now,
Decoration::inner_to_outer() picks a reasonable scheme if last_scheme is
not yet initialized. The test case for this is introduced in #1247.


blarz (approved):
Fixed a small typo in the comment, but other than that this looks good!

Issue #831, opened 2020-04-19
List floating clients of (current) tag


tiosgz:
It would be nice to have a command to list floating clients (their IDs) of a tag (possibly as a part of `dump` — but how?). This is good e.g. for
* listing all clients of current tag (without calling `attr clients.*.tag` and (if needed) `attr clients.*.floating`), also it could ease listing them in the `cycle_all` order (and this is what I want to use it for)
* restoring *complete* layout of a tag within a session (i.e. when the winids are valid)

t-wissmann:
I'm not sure how to extend the `dump`/`load` command without breaking compatibility. The easiest solution would be to implement a new command
```
list_clients [--floating|--tiling|--all-monitors|--all-tags]
```
But I'm not sure what the output should be. One window id per line? or additional information like the floating state or the window title? (`wmctrl -l` does something very similar, but it can't distinguish between floating and tiled clients).

tiosgz:
> I'm not sure how to extend the `dump`/`load` command without breaking compatibility.

Options (for `dump`, `load` could use autodetection)? Also see below how it could work.
> The easiest solution would be to implement a new command
> 
> ```
> list_clients [--floating|--tiling|--all-monitors|--all-tags]
> ```

Short options `ftAa`? Also, a tag/monitor parameter would be good (empty/missing = focus).
> But I'm not sure what the output should be. One window id per line?

Could be. Sorted by the tag layout (i.e. how they come in `cycle_all`). Other possibility is dump-like `(tag [tiled|floating]:f|t (TILED_DUMP) (floating :N WINID...))` (but this would need to improve a lot) (I'd welcome this option as it's more universal, but it's up to you).
> or additional information like the floating state or the window title?

Only `floating|tiled` if anything. These properties change for every use and are easy to get via `attr` (while the order not).
> (`wmctrl -l` does something very similar, but it can't distinguish between floating and tiled clients).

Unfortunately, this seems to order them by ID and would need a filter so it's unusable for me

t-wissmann:
There also seems to be a need for `list_clients --in-frame`

etosan:
Sorry for not finding this issue sooner, I think this one exactly addresses one of the few remaining pain points I was often hitting with herbstluftwm and that is window collection listing for the purpose of picking and management.

I also agree that proposed dump format of one xid in HEXID form per window per line is sufficient and very scripting friendly. For more detailed output dump is already there.

[...]

PR #1249, opened 2021-03-26
Add new command 'list_frame' that prints out list of x11 clients bound to current frame

etosan:
With tiling window managers users may have very different usage patterns.

herbstlutfwm is unique in that it mixes manual window management (a la ancient musca)
with automatic one (like that in dwm, xmonad, etc.).

Unit of manual management is frame. Unit of automatic management is internal frame
layout mode. Some users like me, prefer explicit manual management style more than
the automatic one.

In my usecase, I often prefer 'max' layout as it allows biggest amount
of window information to be presented and also keeps window placement and ordering
strictly manual. But I also make heavy use of 'max's' "stacking" property. As I go,
I end up quickly opening and closing many windows in a frame (as with terminal like
urxvt, this operation is very cheap), mainly every time I discover I need to do
something related to the task at hand, but I don't want to pollute the current
"primary" window.

This works well for very shortlived windows, but some subtasks may turn out to take
longer then initailly anticipated. Those tend to stack up in the frame in increasing
number.

One can, of course, in those cases, cycle layouts in a frame or try, for example
just 'horizontal' layout, to get very quick overview, but in frames with many "running"
subtasks, this is rather inefficient and can cause heavy redraws of frame clients, just
to find out what they are and what they are doing.

It turns out, that window titles of such subtasks are most indicative information
one needs to manage and cull their numbers (or observe their overall progress),
and this presentation (just as a list of titles) also consumes much less mental
bandwith too.

To present this list, the `rofi`(or `dmenu`) tool is much better suited than anything
wm can provide, as it then allows very agile and configurable operations on this
in-frame windows list, like qucikly slicing, dicing and filtering such window set.

Unfortunately, it is extremely hard and inefficient to get a list of client WindowIDs
in the current frame by current methods. Especially when, with few lines of code,
xid list of current frame can be easily provided by herbstluftwm itself.

This commit adds a command, 'list_frame', which implements exactly that: very simple
iterator that dumps xids of all windows bound to current frame. List is dumped as-is
without any filtering and in it's entirety. This very useful information then can
be processed as user as they see fit, to match their needs.

I believe, that besides me, other users will find other creative uses for this command
as well.

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -215,6 +215,13 @@ list_keybinds::
 
 WARNING: Tabs within command parameters are not escaped!
 
+list_frame::
+    Lists all X11 window ids (as 'HEXID''s) of windows bound to currently
+    focused frame. This is convenience command most useful with max layout,
+    but it works with other layouts as well. Use this if you want to implement
+    custom focused frame (only) window picker using external tools like *rofi*
+    (or *dmenu*).
+
 lock::
     Increases the 'monitors_locked' setting. Use this if you want to do multiple
     window actions at once (i.e. without repainting between the single steps).
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -846,3 +846,14 @@ int frame_move_window_edge(Input input, Output output) {
     return 0;
 }
 
+int frame_list_clients(Output output) {
+    Monitor* current_monitor = get_current_monitor();
+    HSTag* current_tag = current_monitor->tag;
+    auto focusedFrame = current_tag->frame->focusedFrame();
+    focusedFrame->foreachClient([&output](Client* client) {
+        output << WindowID(client->x11Window()).str();
+        output << "\n";
+    });
+    return 0;
+}
+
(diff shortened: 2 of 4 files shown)


t-wissmann:
Does not this command do precisely what you want?
```
herbstclient dump '' '@'|grep -ioE '0x[0-9a-f]*'
```
Here, `@` is the current frame, and the grep just extract all window id's that occur in it.

t-wissmann:
In general, I think it's reasonable to provide the window list natively with a command, but implementing `list_frame` seems a bit too narrow. There is the feature request https://github.com/herbstluftwm/herbstluftwm/issues/831 of listing clients, and having a `--in-frame` seems reasonable.

etosan:
Of course, if there is more generic solution possible I am all for it, this was my way of scratching the itch. 

I too ended up with realization some generic window collection picking and filtering engine would be much better fit, but sadly, it's above my current capabilities. In this patch I was really confused by OOP and "friend" property and methods access handling here. It seems I am not very OOP capable.

However it appears that #831 is really capable of everything I ever wanted and much, much more, so if it can be extended for frame support this patch is completely unnecessary. Sorry for missing it.

t-wissmann:
No worries, sorry for the effort you put in this PR!

PR #1251, opened 2021-03-27
Migrate focus_nth to ArgParse

t-wissmann:
Since focus_nth is just a convenience setting for
tags.focus.tiling.focused_frame.selection it better fits in
GlobalCommands. I've moved the existing test case to the other python
file and added a few more test cases. Also, the completion now
explicitly mentions '0' and '-1' for selecting the first and last window
in a frame.

This requires access to tags.focus which I thus had to make public in
TagManager.


blarz (approved):
Looks good!

PR #1253, opened 2021-03-27
Allow relative values for int and uint attributes

### LINKED ISSUE #1174: allow to increase (or decrease) int attributes
It may be useful to allow increasing/decreasing int attributes.

# Pros:

* I've found two use cases:
    1. [increment a counter](https://github.com/herbstluftwm/herbstluftwm/pull/1140#issuecomment-779226542)
    2. [increment a setting](https://github.com/herbstluftwm/herbstluftwm/issues/42#issue-69303315)
* such a "read then write" feature already exists for booleans (i.e. `toggle`), so it seems natural to allow similar feature for `int`
* as we can use calls like `use_index +1` it would make sense to allow calls like `set_attr tags.focus.index +1`

# Cons:

* perhaps two use cases are not enough?
* perhaps the goal is not to make `herbstclient` turing-complete :-)

t-wissmann:
Similar to the "toggle" value of booleans, this introduces relative
values for (signed and unsigned) integer attributes.

This closes #1174.


blarz (approved):
Looks good to me!

PR #1255, opened 2021-03-27
Migrate 'close' command to ArgParse

t-wissmann:
The usual migration commit: use ArgParse for argument parsing and
command line completions, add tests for it, remove the old
implementation and completion code.

### DIFF
--- a/tests/test_clients.py
+++ b/tests/test_clients.py
@@ -49,9 +49,48 @@ def test_close_without_clients(hlwm):
 
 
 @pytest.mark.parametrize("running_clients_num", [1, 2, 3, 4])
-def test_close(hlwm, running_clients_num):
+def test_close_for_multiple_clients(hlwm, running_clients_num):
     hlwm.create_clients(running_clients_num)
     hlwm.call('close')
+
+
+def test_close_focused_client(hlwm):
+    winid, proc = hlwm.create_client()
+
+    hlwm.call('close')
+
+    proc.wait(10)  # wait for the client to shut down
+
+
+def test_close_unfocused_client(hlwm):
+    focused, _ = hlwm.create_client()
+    unfocused, proc = hlwm.create_client()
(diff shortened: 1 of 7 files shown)


tests/test_clients.py
```diff
@@ -49,11 +49,50 @@ def test_close_without_clients(hlwm):
 
 
 @pytest.mark.parametrize("running_clients_num", [1, 2, 3, 4])
-def test_close(hlwm, running_clients_num):
+def test_close_for_multiple_clients(hlwm, running_clients_num):
```
blarz, line 52: This test and some others related to the "close" command do not assert anything.
Is this inteded? How is it tested, that the clients are actually closed?

blarz (commented):
Looks good, just one question about the tests in general

tests/test_clients.py
t-wissmann: I also can't remember what's the purpose of this test. It's inconvenient to wait for a client to close if we don't have it's process handle. Maybe this test case just checked that hlwm does not crash. Maybe I should just extend `test_close_focused_client` by other unfocused clients to verify that 'close' only closes the focused one.

t-wissmann: I've removed this incomplete test and extended the other test where it is actually verified that the focused client closes.

blarz: Thanks for updating the test!
Should we add checks for the client to be closed in the unfocused and unmanaged case as well?

t-wissmann: In the unfocused and unmanaged case, it is already checked that it is closed (if something is closed, then there is `proc.wait(10)` in the test.)

blarz: Aaah, yes, you're right. I was looking for asserts, but the proc.wait() is the assert in that case

blarz (approved):

Issue #948, opened 2020-07-05
When all clients in non-floating tag are floating, there’s a ghost client


tiosgz:
Switch to an empty tag, open a terminal, per-client float the terminal, cycle-all. No client is focused.

t-wissmann:
This is (partially) intended. If you have a tag in tiling mode, then there is one empty frame. If you open one floating client, then cycle-all cycles between the floating client and the empty frame.
If you split, then you will have two empty frames, leading to two "ghost clients". If one has visible frame borders, one can see that the frame is selected.

Unfortunately, the same behaviour applies if the tag is in floating mode. In that case, it is really confusing, which really should be fixed (so this issue needs to stay open).

tiosgz:
Thanks for the explanation, I didn’t think of that. I’ve changed the title.

bklaase:
Might an additional cycle command offer a solution? I don't know about the others, but cycling with win-tab (`cycle_all`) is only used on non tiling tags. Perhaps there could be a `cycle_all_clients`, or perhaps a `cycle_all --skip-frames` ?

Easily achieved through scripting as well, but perhaps it makes sense?

PR #1258, opened 2021-03-28
Make 'cycle' work for the floating layer

t-wissmann:
If the floating layer is focused, 'cycle' should traverse through the
clients in there. While at it, I've migrated the command to ArgParse and
moved it to HSTag. As usual, there are some new test cases.

Still, this only cycles through the windows explicitly marked as
floating, which is maybe odd when switching the entire tag to floating
(which should be fixed in the long run, see issue #948).


blarz:
Change looks good but I want to play with it a bit more locally and will do so tomorrow

blarz (approved):
Looks good to me!

PR #1262, opened 2021-03-28
Simplify unminimize.sh using integer incrementation

apm256:
Thanks to https://github.com/herbstluftwm/herbstluftwm/commit/bd882ae3504150e6a011b0769f786ac46f6e40f8 , we can simplify the `unminimize.sh` script and use `herbstclient` in place of bash to increment the required counter.
(cc @M1ndo)

### DIFF
--- a/scripts/unminimize.sh
+++ b/scripts/unminimize.sh
@@ -9,65 +9,43 @@
 Mod=${Mod:-Mod4}
 Minimizekey=Shift-m
 Unminimizekey=Ctrl-m
-# get the absolute path of this script, to call it when minimizing
-SCRIPT_PATH=$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/$(basename "${BASH_SOURCE[0]}")
 
 hc() { "${herbstclient_command[@]:-herbstclient}" "$@" ;}
 
 
-# 
-# initialize minimize and unminimize shortcuts
-#
-init() {
+# initialize a global minimization counter
+hc silent new_attr uint my_minimized_counter 1
 
-   # initialize a global minimization counter
-   hc silent new_attr uint my_minimized_counter 1
+# minimize focused window
+hc keybind $Mod-$Minimizekey hc and \
+  . substitute C my_minimized_counter new_attr uint clients.focus.my_minimized_age C \
+  . set_attr my_minimized_counter "+=1" \
+  . set_attr clients.focus.minimized true
 
-   # minimize current window
-   hc keybind $Mod-$Minimizekey spawn "$SCRIPT_PATH" minimize
-
-   # unminimize last window of a tag
-   # if the `my_minimized_age` attribute does not exist (i.e. the window has not been
-   #  minimized with this script), use arbitrary order to unminimize
-   hc keybind $Mod-$Unminimizekey mktemp string LASTCLIENTATT mktemp uint LASTAGEATT chain \
-     . set_attr LASTAGEATT 0 \
-     . foreach CLIENT clients. and \
-       , sprintf MINATT "%c.minimized" CLIENT \
-           compare MINATT "=" "true" \
-       , sprintf TAGATT "%c.tag" CLIENT substitute FOCUS "tags.focus.name" \
-           compare TAGATT "=" FOCUS \
-       , sprintf AGEATT "%c.my_minimized_age" CLIENT or \
-         case: and \
-            : ! get_attr AGEATT \
-            : compare LASTAGEATT "=" 0 \
-         case: and \
-            : substitute LASTAGE LASTAGEATT \
-                compare AGEATT 'gt' LASTAGE \
-            : substitute AGE AGEATT \
-                set_attr LASTAGEATT AGE \
-       , set_attr LASTCLIENTATT CLIENT \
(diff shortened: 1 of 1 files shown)


M1ndo (approved):
Just tested it, it's working just like expected .

scripts/unminimize.sh
```diff
@@ -9,65 +9,43 @@
...
-
-}
-
-
-if [ "$1" = "minimize" ] ; then minimize ; else init ; fi
+# initialize a global minimization counter
+hc silent new_attr uint my_minimized_counter 1
+
+# minimize focused window
+hc keybind $Mod-$Minimizekey hc and \
```
t-wissmann, line 20: what is the second `hc` in line 20? (and why does it even work for you? Maybe an old keybinding doing the same work? `hc` is not a hlwm command, and even if there was `spawn hc`, the `hc` still would not refer to the bash function.)
```
./unminimize.sh 
keybind: the command "hc" does not exist. Did you forget "spawn"?
```

scripts/unminimize.sh
```diff
@@ -9,65 +9,43 @@
...
+  . substitute C my_minimized_counter new_attr uint clients.focus.my_minimized_age C \
```
t-wissmann, line 21: This fails if the client has been minimized before and unminimized through a panel. (Of course, minimizing through a panel also bypasses, but when `try`ing it here, we can at least cover the one direction).

t-wissmann (changes requested):
Thanks for updating. There are two mistakes, and when fixing both, I end up with this command:
```
hc keybind $Mod-$Minimizekey and \
  . try new_attr uint clients.focus.my_minimized_age \
  . substitute C my_minimized_counter set_attr clients.focus.my_minimized_age C \
  . set_attr my_minimized_counter "+=1" \
  . set_attr clients.focus.minimized true
```
Please try it, and if it works for you, too, then update it in in the PR :-)

scripts/unminimize.sh
M1ndo: Thanks for reviewing , i had the same issue last night however i managed to fixed it with following 
``` 
hc keybind $Mod-n chain \
   . substitute C my_minimized_counter new_attr int clients.focus.my_minimized_age C \
   . set_attr my_minimized_counter $(($(herbstclient substitute C my_minimized_counter echo C)+1)) \
   . set_attr clients.focus.minimized true \
```
It also works using no keybinding at all it's only necessary to add a keybind in autostart in order to launch the script when uminimizing or minimizing .

t-wissmann: This shouldn't work correctly, because the `$(( ... ))` is evaluated when the keybind is registered and not when the key is pressed.

apm256:
Thank you for your review.
I'm a little bit embarrassed: obviously I made something wrong when submitting this PR. Here is my local file, which does not exhibit two `hc` calls:

```
#!/bin/bash

# A script allowing to minimize and un-minimize clients in a LIFO way
# (last minimized client will be un-minimized first).
# 
# Run it (`chmod +x unminimize.sh ; ./unminimize.sh`) or add it to autostart.


Mod=${Mod:-Mod4}
Minimizekey=Shift-m
Unminimizekey=Ctrl-m

hc() { "${herbstclient_command[@]:-herbstclient}" "$@" ;}


# initialize a global minimization counter
hc silent new_attr uint my_minimized_counter 1

# minimize focused window
hc keybind $Mod-$Minimizekey and \
  . substitute C my_minimized_counter new_attr uint clients.focus.my_minimized_age C \
  . set_attr my_minimized_counter "+=1" \
  . set_attr clients.focus.minimized true

# unminimize last window of a tag
hc keybind $Mod-$Unminimizekey mktemp string LASTCLIENTATT mktemp uint LASTAGEATT chain \
  . set_attr LASTAGEATT 0 \
  . foreach CLIENT clients. and \
    , sprintf MINATT "%c.minimized" CLIENT \
        compare MINATT "=" "true" \
    , sprintf TAGATT "%c.tag" CLIENT substitute FOCUS "tags.focus.name" \
        compare TAGATT "=" FOCUS \
    , sprintf AGEATT "%c.my_minimized_age" CLIENT and \
      : substitute LASTAGE LASTAGEATT \
          compare AGEATT 'gt' LASTAGE \
      : substitute AGE AGEATT \
          set_attr LASTAGEATT AGE \
    , set_attr LASTCLIENTATT CLIENT \
  . and \
    , compare LASTCLIENTATT "!=" "" \
    , substitute CLIENT LASTCLIENTATT chain \
      : sprintf MINATT "%c.minimized" CLIENT \
          set_attr MINATT false \
      : sprintf AGEATT "%c.my_minimized_age" CLIENT \
          remove_attr AGEATT
```

I will fix the script following your review, and update this PR in a few days. Sorry again!

t-wissmann:
Don't worry :-) You can also post the file in a comment here and I can update the PR accordingly if it's easier for you.

t-wissmann:
Any news on this? I'm by the way implementing this directly in PR #1313

blarz:
This can probably be closed soon, right?

t-wissmann:
Yes, since it is built directly in hlwm with
`longest-minimized` and `last-minimized`, I think this can be closed.

PR #1263, opened 2021-03-29
tests: Relax timeouts for shut down a bit

t-wissmann:
Sometimes the CI fails because the shutdown takes too long.
Thus, a slightly longer timeout might be reasonable. Moreover, the
shutdown of the HlwmBridge is now in parallel, so all processes are
sent the term signal simultaneously, so in sum, this might even speed up
the shut down phase of test cases with clients.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -28,6 +28,7 @@ COPY_ENV_WHITELIST = ['LSAN_OPTIONS']
 # time in seconds to wait for a process to shut down
 PROCESS_SHUTDOWN_TIME = 5
 
+
 def extend_env_with_whitelist(environment):
     """Copy some whitelisted environment variables (if set) into an existing environment"""
     environment.update({e: os.environ[e] for e in os.environ if e in COPY_ENV_WHITELIST})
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -254,12 +254,16 @@ def wait_for_window_of(self, wmclass):
...
             client_proc.terminate()
-            client_proc.wait(2)
-
         self.hc_idle.terminate()
-        self.hc_idle.wait(2)
+
+        # and then wait for each of them to finish:
+        for client_proc in self.client_procs:
+            client_proc.wait(5)
+        self.hc_idle.wait(5)
```
blarz, line 266: You might want to extract the 5 into a constant variable, so in the future we can just increase the value in one place instead of many. HLWM_WAIT_TIMEOUT or something?

blarz (changes requested):

blarz:
Not sure why this CI fix breaks/doesn't trigger the CI

tests/conftest.py
t-wissmann: I've created a new global variable; and now the CI is triggered again :)

blarz: Nice, thanks!

blarz (approved):

PR #1265, opened 2021-03-30
Remove old completion code from command.cpp

t-wissmann:
With the completion code from all commands migrated, most of the code in
command.cpp is now obsolete. There is still a bit of the char**-based code
which needs to be reorganized until we get rid of the old command
interface entirely.

The new tests verify that the exit code of the completion of `spawn` and
`wmexec` still are correct.

### DIFF
--- a/tests/test_completion.py
+++ b/tests/test_completion.py
@@ -207,7 +207,7 @@ def test_inputless_commands(hlwm, name):
 
 
 def test_completionless_commands(hlwm):
-    # commands that accept arguments but don't have compltion.
+    # commands that accept arguments but don't have completion.
     # Their completion must return '0' (instead of NO_PARAMETER_EXPECTED)
     for cmd in ['spawn', 'wmexec']:
         for idx in ['1', '2', '3']:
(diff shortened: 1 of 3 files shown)


tests/test_completion.py
```diff
@@ -206,6 +206,14 @@ def test_inputless_commands(hlwm, name):
         .returncode == 7
 
 
+def test_completionless_commands(hlwm):
+    # commands that accept arguments but don't have compltion.
```
blarz, line 210: typo completion

blarz: Was this even intended to be in this change? The commit message only subscribes removal of unused code.

blarz (changes requested):

tests/test_completion.py
t-wissmann: I've fixed the typo.

Yes this is intended. It tests that the code removal in `complete_against_commands()` does not break anything (`spawn` and `wmexec` are still using the C-style command interface)

blarz: Thanks for the fix and adding the info to the commit message :)

blarz (approved):
Spring cleaning!

t-wissmann:
Thanks :)

PR #1268, opened 2021-04-03
Rename 'geometry_reported' to 'content_geometry'

t-wissmann:
I think the latter name is more self-explanatory and is not specific to
the internals of the X11 protocol.


blarz (approved):
I agree, it sounds nicer

PR #1269, opened 2021-04-03
tox: Switch default Python env to 38

blarz:
Not sure if 3.6 is really used in "current" distributions anymore? At
least the CI is using 3.8. Also adjust the description in HACKING to be
more generic, this is not an Arch specific problem.


t-wissmann (approved):
Seems reasonable :-)

PR #1272, opened 2021-04-04
Add client attribute 'floating_geometry'

t-wissmann:
Add a new attribute to access and write the floating geometry of a
client. In order to keep the diff small, I kept the name of the C++
variable in the source code, even though '...geometry' is more
appropriate.

In the diff, I replaced member write-access with a preparation of the
new Rectangle and a single write. This is inevitable anyway because the
operators * and -> of Attribute_<> only return const references (as they
should). Read-only member access is no problem; here, I only had to
replace '.' with '->'.

There is not much necessity to test the attribute itself, but instead,
it serves as a tool to test other features of hlwm (e.g. the pad and the
monitor geometry).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -20,7 +20,7 @@ Current git version
     'class', 'geometry', 'winid')
   * The setting 'monitors_locked' is now explicitly an unsigned integer.
   * The setting 'default_frame_layout' now holds an algorithm name.
-  * New client attribute:
+  * New client attributes:
     - 'float_geometry' holding the client's floating size (writable).
     - 'content_geometry' holding the geometry of the application's content.
   * The 'shift' command now moves the window to a neighboured monitor if
--- a/NEWS
+++ b/NEWS
@@ -21,7 +21,7 @@ Current git version
   * The setting 'monitors_locked' is now explicitly an unsigned integer.
   * The setting 'default_frame_layout' now holds an algorithm name.
   * New client attributes:
-    - 'float_geometry' holding the client's floating size (writable).
+    - 'floating_geometry' holding the client's floating size (writable).
     - 'content_geometry' holding the geometry of the application's content.
   * The 'shift' command now moves the window to a neighboured monitor if
     the window cannot be moved within a tag in the desired direction.
--- a/NEWS
+++ b/NEWS
@@ -20,8 +20,9 @@ Current git version
     'class', 'geometry', 'winid')
   * The setting 'monitors_locked' is now explicitly an unsigned integer.
   * The setting 'default_frame_layout' now holds an algorithm name.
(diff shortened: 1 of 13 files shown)


t-wissmann:
I'm not sure what's the better name: `float_geometry` or `floating_geometry`? The latter is more consistent I think, but the former is shorter.

t-wissmann:
> In the diff, I replaced member write-access with a preparation of the
> new Rectangle and a single write.

And as I expected, I confused `width` and `height` from time to time. One occurrence was just discovered by the test suite :-).. So I guess I should read the diff again later or tomorrow.

t-wissmann:
> > In the diff, I replaced member write-access with a preparation of the
> > new Rectangle and a single write.
> 
> And as I expected, I confused `width` and `height` from time to time. One occurrence was just discovered by the test suite :-).. So I guess I should read the diff again later or tomorrow.

I've just checked the diff and didn't spot any further swaps of `width`/`height` or `x`/`y`

NEWS
```diff
@@ -20,8 +20,9 @@ Current git version
     'class', 'geometry', 'winid')
   * The setting 'monitors_locked' is now explicitly an unsigned integer.
   * The setting 'default_frame_layout' now holds an algorithm name.
-  * New client attribute 'content_geometry' holding the geometry of the
-    application's content.
+  * New client attribute:
```
blarz, line 23: Rather use the plural attributes here?

blarz:
> I'm not sure what's the better name: `float_geometry` or `floating_geometry`? The latter is more consistent I think, but the former is shorter.

I think I like `floating_geometry` better, consistency is king ;)
Everything else in commands or attributes is referring to something *floating.

NEWS
t-wissmann: [x] fixed.

t-wissmann:
> > I'm not sure what's the better name: `float_geometry` or `floating_geometry`? The latter is more consistent I think, but the former is shorter.
> 
> I think I like `floating_geometry` better, consistency is king ;)
> Everything else in commands or attributes is referring to something *floating.

I've renamed it to `floating_geometry` :)

NEWS
blarz: Thanks!

blarz:
> > > I'm not sure what's the better name: `float_geometry` or `floating_geometry`? The latter is more consistent I think, but the former is shorter.
> > 
> > 
> > I think I like `floating_geometry` better, consistency is king ;)
> > Everything else in commands or attributes is referring to something *floating.
> 
> I've renamed it to `floating_geometry` :)

Thank you :)

blarz (approved):

PR #1273, opened 2021-04-04
New command 'attr_type'

t-wissmann:
The command only prints the type of a given attribute and will be of use
in the python bindings.


tiosgz:
What about `attr_type` or `get_attr_type`?

t-wissmann:
I tried to keep it consistent with `set_attr` and `get_attr`. Often `get` is redundant in command or function names. Let's see what @blarz thinks :)

blarz:
> I tried to keep it consistent with `set_attr` and `get_attr`. Often `get` is redundant in command or function names. Let's see what @blarz thinks :)

I'll have a look tomorrow :)

doc/herbstluftwm.txt
```diff
@@ -850,6 +850,9 @@ set_attr 'ATTRIBUTE' 'NEWVALUE'::
     Assign 'NEWVALUE' to the specified 'ATTRIBUTE' as described in the
     <<OBJECTS,*OBJECTS section*>>.
 
+type_of_attr 'ATTRIBUTE'::
+    Print (one line containing) the type of the specified 'ATTRIBUTE'.
```
blarz, line 854: Is the (one line containing) necessary? It kind of reads "bumpy", if you know what I mean

blarz:
> I tried to keep it consistent with `set_attr` and `get_attr`. Often `get` is redundant in command or function names. Let's see what @blarz thinks :)

What about removing `of_` and reduce it to `type_attr`?
Would be consistent with the other `WORD_attr` (although type is not a verb ;))
Looks OK IMHO. What you think @t-wissmann and @tiosgz?

tiosgz:
> What about removing `of_` and reduce it to `type_attr`?

Then it looks like an adjective to me (that is, an attribute storing a type). Or…
> (although type is not a verb ;))

…unfortunately, it is (`hc type_out_this_attr my_random_attr` into the terminal I’m running if from :P).

doc/herbstluftwm.txt
t-wissmann: Hm, I think I can also remove it. It will be important that the output is precisely one line :)

t-wissmann:
Hm, I think `type_attr` sounds like "typing the attribute" (i.e. assigning a type or typing it on a keyboard).

blarz:
I see both your points :D

Not sure, `type_of_attr` seems.. too long or too verbose?

`attr_type` is snappy, I like it, too. But I also understand that it is not consistent with the other `*_attr` commands.

tiosgz:
I could only think of ‘identify’ & ‘recognise’, which are both too vague or completely off.

Wrt. consistency… ‘get’ & ‘set’ are verbs, but so far we’ve got no verb to put there. When it comes to adjectives/‘of’ in other commands:
- `monitor_rect` not `rect_of_monitor`
- `object_tree` not `tree_of_objects`
- `tag_status` not `status_of_tags`

t-wissmann:
I see your point. I've renamed it to `attr_type`. It's a pitty that this clashes with the tab-completion for `attr`, but at least typing `hc at`<kbd>Tab</kbd> will still complete to `hc attr` and the user only has to type the space manually.

blarz:
> I see your point. I've renamed it to `attr_type`. It's a pitty that this clashes with the tab-completion for `attr`, but at least typing `hc at`Tab will still complete to `hc attr` and the user only has to type the space manually.

attr_type sounds good to me. What do you think @tiosgz ?

doc/herbstluftwm.txt
blarz: Thanks! Reads better now IMHO. If it is important we can add it, maybe the only thing making it "bumpy" were the parantheses

tiosgz:
> attr_type sounds good to me. What do you think @tiosgz ?

Sure it can be.

doc/herbstluftwm.txt
tiosgz: ‘a’ instead of ‘one’ would also make it read better (IMO)

t-wissmann: I've removed the addition in the parentheses already.

blarz: Closing this now :)

blarz (approved):
In it goes :)

PR #1274, opened 2021-04-05
New command 'list_clients'

t-wissmann:
For the new command 'list_clients' the main challenge was to extend
ArgParse by command line flags that take arguments. To make the
completion work, I also had to extend the completion objects by a
'prepend' string that is prepended to every completion result.

This implements the main feature request of #831 and also includes the
functionality of the rejected PR #1249.

### DIFF
--- a/src/argparse.cpp
+++ b/src/argparse.cpp
@@ -54,7 +54,21 @@ bool ArgParse::parsingFails(Input& input, Output& output)
         string valueString;
         input >> valueString;
         // try to parse this token as a flag
-        while (optionalArgumentsRemaining && tryParseFlag(valueString)) {
+        while (optionalArgumentsRemaining) {
+            try {
+                if (!tryParseFlag(valueString)) {
+                    // if it is not a flag, then break, i.e
+                    // continue with positional args
+                    break;
+                }
+            }  catch (std::exception& e) {
+                // if there is a parse error however, then
+                // stop entire parsing process
+                output << input.command() << ": Cannot parse flag \""
+                       << valueString << "\": " << e.what() << "\n";
+                errorCode_ = HERBST_INVALID_ARGUMENT;
+                return true;
+            }
             optionalArgumentsRemaining--;
             // if this token was the last optional argument
             if (optionalArgumentsRemaining == 0 && arg.optional_) {
@@ -85,10 +99,17 @@ bool ArgParse::parsingFails(Input& input, Output& output)
     // try to parse more flags after the positional arguments
     while (!input.empty()) {
         // only consume tokens if they are recognized flags
-        if (tryParseFlag(input.front())) {
-            input.shift();
-        } else {
-            break;
+        try {
+            if (tryParseFlag(input.front())) {
+                input.shift();
+            } else {
+                break;
+            }
+        }  catch (std::exception& e) {
+            output << input.command() << ": Cannot parse flag \""
+                   << input.front() << "\": " << e.what() << "\n";
+            errorCode_ = HERBST_INVALID_ARGUMENT;
+            return true;
         }
     }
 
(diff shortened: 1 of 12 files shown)


t-wissmann:
I'm not sure whether `--in-tag=…` or `--on-tag=…` is better (and same for `--on-monitor=…`); "in" sounds a bit odd, but would be more consistent with `--in-frame`

tiosgz:
> I'm not sure whether `--in-tag=…` or `--on-tag=…` is better (and same for `--on-monitor=…`); "in" sounds a bit odd, but would be more consistent with `--in-frame`

(or remove the prepositions… though @etosan said it sounds more natural to him with them)

tests/test_global_commands.py
```diff
@@ -264,3 +264,115 @@ def test_focus_nth_completion(hlwm):
...
+    def list_clients(flags):
+        return sorted(hlwm.call(['list_clients'] + flags).stdout.splitlines())
+
+    assert list_clients(['--in-frame=']) == sorted(clients)
+    assert list_clients(['--in-frame=@']) == sorted(clients[0:2])
+    assert list_clients(['--in-frame=0']) == sorted(clients[0:2])
+    assert list_clients(['--in-frame=1']) == sorted(clients[2:4])
+
+
+def test_list_clients_invalid_arg(hlwm):
```
blarz, line 331: I tested ```list_clients --on-monitor=``` and it works, is this intended?

src/argparse.cpp
```diff
@@ -213,17 +234,55 @@ void ArgParse::completion(Completion& complete)
 }
 
 /**
- * @brief try parse a flag
+ * @brief try parse a flag, possibly throwing an exception
```
blarz, line 237: try *to* parse a flag?

blarz:
> > I'm not sure whether `--in-tag=…` or `--on-tag=…` is better (and same for `--on-monitor=…`); "in" sounds a bit odd, but would be more consistent with `--in-frame`
> 
> (or remove the prepositions… though @etosan said it sounds more natural to him with them)

I wouldn't oppose `--tag` or `--monitor` either to be honest.

tests/test_global_commands.py
t-wissmann: Yes, this is a corner case because the empty string refers to the focused monitor.

src/argparse.cpp
t-wissmann: [x] fixed, thanks!

t-wissmann:
Yes, simply `--tag=` and `--monitor=` make sense. I've renamed them. So far, no hlwm commands follow the `--in-…`/`--or-…` pattern (the only command line flags with parameters so far were the ones for the `rule` command)

src/argparse.cpp
```diff
@@ -54,7 +54,21 @@ bool ArgParse::parsingFails(Input& input, Output& output)
...
         // try to parse this token as a flag
-        while (optionalArgumentsRemaining && tryParseFlag(valueString)) {
+        while (optionalArgumentsRemaining) {
+            try {
+                if (!tryParseFlag(valueString)) {
+                    // if it is not a flag, then break, i.e
+                    // continue with positional args
+                    break;
+                }
+            }  catch (std::exception& e) {
```
blarz, line 64: This branch is not covered according to the coverage report. Does it make sense to add a test for this?

src/argparse.cpp
blarz: Thanks!

tests/test_global_commands.py
blarz: I see!

src/argparse.cpp
t-wissmann: This code is not reachable at the moment because it is only executed if there is a command line flag with parameters in a command with position arguments. `list_clients` however is the onle command with parametrized flags, but it does not have positional args. Once this is merged, I plan to extend `foreach` by a `--name-match=…` flag, and then we can test this branch.

blarz: Ok thanks! Just wanted to check :)

blarz (approved):
Very nice, thanks!

PR #1275, opened 2021-04-05
Migrate completion commands

t-wissmann:
This migrates the entry points of the completion system: the commands
'complete' and 'complete_shell'. Now, the Completion objects are used
throughout the completion system.

This also fixes a bug in `mergeResultsFrom()` and in the test case
for the completion of `! true` along the way.


src/command.cpp
```diff
@@ -134,24 +127,27 @@ shared_ptr<const CommandTable> Commands::get() {
...
+        }
+    } else {
+        auto it = commandTable->find(completion[0]);
+        if (it == commandTable->end()) {
+            completion.invalidArguments();
+        } else if (it->second.hasCompletion()) {
+            // get new completion context with command name omitted.
+            Completion shifted = completion.shifted(1);
+            it->second.complete(shifted);
+            // TODO: call mergeResultsFrom(), once #1274 is merged.
```
blarz, line 143: Now that #1274  is merged, do you want to update this PR or fix the TODO in a separate one?

t-wissmann: [x] done :-)

blarz: Thanks :)

blarz:
`test_close_completion` seems to fail

t-wissmann:
There was a bug in `mergeResultsFrom()` (even before #1274), lets see whether fixing it makes some other test case fail.

blarz (approved):
Nice!

PR #1277, opened 2021-04-09
Auto-convert attributes to/from python types

t-wissmann:
In the python bindings, automatically convert the attribute values from
(resp. to) their corresponding python types when writing (resp. reading)
attributes.

This requires touching (but fortunately simplifying) a lot of assertions
in the test cases. Now, working with the contents of hlwm.attr should
feel as if one had the hlwm attributes available as python variables.

A few new test cases at the end of test_python_binds.py illustrate the
usage for all types that are available at the moment.

### DIFF
--- a/python/herbstluftwm/types.py
+++ b/python/herbstluftwm/types.py
@@ -48,3 +48,81 @@ class Rectangle:
         are adjusted by the provided deltas.
         """
         return Rectangle(self.x + dx, self.y + dy, self.width + dw, self.height + dh)
+
+
+class HlwmType:
+    """
+    Wrapper functions for converting between python types and types
+    in herbstluftwm.
+    """
+    def __init__(self, name, from_user_str, to_user_str, is_instance):
+        # a hlwm type should define the following
+        self.name = name  # type: str
+        # a callback for parsing
+        self.from_user_str = from_user_str  # of type: str -> T
+        # a callback for printing
+        self.to_user_str = to_user_str  # of type: T -> str
+        # a predicate whether a python variable has this type:
+        self.is_instance = is_instance  # of type: Anything -> bool
+
+    @staticmethod
+    def by_name(type_name):
+        """Given the full name of a hlwm type, return
+        the metadata
+        python type"""
+        for t in hlwm_types():
+            if t.name == type_name:
+                return t
+        return None
+
+    @staticmethod
+    def by_type_of_variable(python_variable):
+        """Given a variable, detect its type
+        """
+        for t in hlwm_types():
+            if t.is_instance(python_variable):
+                return t
+        return None
+
+
+def hlwm_types():
+    """
+    Return a list of HlwmType objects.
+
+    Unfortunately, the order matters for the is_instance() predicate: Here, the
+    first matching type in the list must be used. (This is because
+    `isinstance(True, int)` is true)
+    """
+    types = [
+        HlwmType('bool',
+                 bool_from_user_str,
+                 lambda b: 'true' if b else 'false',
+                 lambda x: isinstance(x, bool)),
+
+        HlwmType('int', int, str, lambda x: isinstance(x, int)),
+
+        # there is no uint in python, so we just convert it to 'int'
+        HlwmType('uint', int, str, lambda x: False),
+
+        HlwmType('rectangle',
+                 Rectangle.from_user_str,
+                 Rectangle.to_user_str,
+                 lambda x: isinstance(x, Rectangle)),
+
+        HlwmType('string',
+                 lambda x: x,
+                 lambda x: x,
+                 lambda x: isinstance(x, str)),
+    ]
+
+    return types
+
+
+def bool_from_user_str(bool_string):
+    """Parse a the string description of a hlwm boolean to
+    a python boolean"""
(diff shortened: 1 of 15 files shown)


python/herbstluftwm/types.py
```diff
@@ -48,3 +48,81 @@ def adjusted(self, dx=0, dy=0, dw=0, dh=0):
...
+
+        HlwmType('rectangle',
+                 Rectangle.from_user_str,
+                 Rectangle.to_user_str,
+                 lambda x: isinstance(x, Rectangle)),
+
+        HlwmType('string',
+                 lambda x: x,
+                 lambda x: x,
+                 lambda x: isinstance(x, str)),
```
The-Compiler, line 115: Personally I find this a bit hard to follow, because the positional lambdas are different "methods", and then there's things scattered in different places in this file (e.g. the `bool_from_user_str` below).

Have you considered making those subclasses of `HlwmType` (with the name being a constant class attribute, and the lambdas being methods instead)?

tests/test_layout.py
```diff
@@ -369,23 +369,23 @@ def test_cycle_stays_in_floating_layer(hlwm):
     for winid in floating:
         assert hlwm.attr.clients.focus.winid() == winid
         # the floating layer always stays focused
-        assert hlwm.attr.tags.focus.floating_focused() == hlwm.bool(True)
+        assert hlwm.attr.tags.focus.floating_focused() is True
```
The-Compiler, line 372:
```suggestion
        assert hlwm.attr.tags.focus.floating_focused()
```

maybe? We're not really testing the `floating_focused` getter type here, and I assume in C(++) you'd also write `if (floating_focused)` rather than `if (floating_focused == true)`.

(Same for all other instances of `is True` and `is False`, except maybe `test_implicit_type_conversion_bool` because we're really interested in the type there)

tests/test_layout.py
```diff
@@ -369,23 +369,23 @@ def test_cycle_stays_in_floating_layer(hlwm):
...
         hlwm.call(['cycle', '+1'])
 
     # we're back a the first floating client:
     assert hlwm.attr.clients.focus.winid() == floating[0]
-    assert hlwm.attr.tags.focus.floating_focused() == hlwm.bool(True)
+    assert hlwm.attr.tags.focus.floating_focused() is True
```
The-Compiler, line 377: ditto

tests/test_monitor.py
```diff
@@ -158,7 +158,7 @@ def test_move_monitor(hlwm):
 
 
 def test_move_monitor_completion(hlwm):
-    assert hlwm.attr.monitors.focus.geometry() in hlwm.complete('move_monitor 0')
+    assert hlwm.get_attr('monitors.focus.geometry') in hlwm.complete('move_monitor 0')
```
The-Compiler, line 161: What's the difference?

The-Compiler (changes requested):
Lovely! I added a couple of comments.

python/herbstluftwm/types.py
t-wissmann: I considered this, but it felt odd to have subclasses with `@staticmethod`s only (and undefined `@staticmethod`s in `HlwmType`). Also I had the impression that it's more write effort.

I agree  that the positional arguments are hard to read. I have just replaced them with named arguments; is this more readable?

tests/test_layout.py
t-wissmann: Hm you're probably right, but it feels fragile because the type is detected at run-time and returning the string `"false"` here would make the `assert` hold. (Also, I think `assert … is True` reads quite well)

tests/test_monitor.py
t-wissmann: the completions is always a list of strings, but `hlwm.attr.monitors.focus.geometry()` is now of type `Rectangle`. I've just replaced it with `str(hlwm.attr.monitors.focus.geometry)`, maybe this emphasizes more what's going on.

t-wissmann:
Thanks for reviewing the PR :-) It's good to have you python expertise :)

python/herbstluftwm/types.py
The-Compiler: Definitely better! I'd still prefer classes (then you could also e.g. have proper type annotations instead of describing the types in the comments), but no strong opinions either way.

blarz (changes requested):
Small typo, other than that, I'd leave it to @The-Compiler  to approve this change

python/herbstluftwm/types.py
```diff
@@ -48,3 +48,87 @@ def adjusted(self, dx=0, dy=0, dw=0, dh=0):
...
+                 from_user_str=lambda x: x,
+                 to_user_str=lambda x: x,
+                 is_instance=lambda x: isinstance(x, str)),
+    ]
+
+    return types
+
+
+def bool_from_user_str(bool_string):
+    """Parse a the string description of a hlwm boolean to
```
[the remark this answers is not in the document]
blarz: Thanks!

blarz (commented):
Removing my request for changes :)

The-Compiler (approved):
Looks good, sorry for the delay!

PR #1278, opened 2021-04-10
Support error output channel internally

t-wissmann:
In the command system, the 'Output' parameter now also carries an error
channel and the command name (in addition to the standard output
channel). This is accomplished with a new class OutputChannels in
commandio.h (and Output = OutputChannels&).

So far, this only affects the command system, but not the IPC interface,
which thus simply concatenates the two output channels in the reply to
'herbstclient'.

Thus, we can now print error messages to this dedicated channel and have
the error message prefixed with the command name automatically. Most of
the diff consists of migrating the error message printing to the new
perror(), which does the prefixing.

As a bonus, CallOrComplete does not need the command name anymore.

Also, HlwmCommon::callCommand() was only used in XMainLoop, so I have
used it there.

### DIFF
--- a/src/commandio.h
+++ b/src/commandio.h
@@ -12,7 +12,58 @@ class Completion;
 using Path = ArgList;
 
 /* Types for I/O with the user */
-using Output = std::ostream&;
+/**
+ * @brief A bundle of the standard output and the
+ * error output channel. Writing to it (via the << operator) just passes
+ * the data to the output channel.
+ *
+ * It is recommended to pass references to OutputChannels (see the Output type)
+ * around to avoid that the command name is copied over and over again.
+ */
+class OutputChannels {
+public:
+    OutputChannels(const std::string& commandName, std::ostream& output, std::ostream& error)
+        : commandName_(commandName)
+        , output_(output)
+        , error_(error)
+    {}
+    static OutputChannels stdio();
+
+    const std::string& command() {
+        return commandName_;
+    }
+
+    std::ostream& error() {
+        return error_;
+    }
+
+    std::ostream& output() {
+        return output_;
+    }
+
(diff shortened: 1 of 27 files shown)


src/commandio.h
```diff
@@ -69,7 +120,6 @@ class Input : public ArgList {
 class CallOrComplete {
 public:
     //! the name of the command that is called.
```
blarz, line 122: Can be removed now?

src/metacommands.cpp
```diff
@@ -801,8 +801,12 @@ int MetaCommands::tryCommand(Input input, Output output) {
 
 int MetaCommands::silentCommand(Input input, Output output) {
     stringstream dummyOutput;
+    // discard output and error channel
+    // TODO: pass through the error channel as soon as
+    // the ipc-protocoll supports it.
```
blarz, line 806: protocol

blarz (changes requested):
Two minor findings, other than that, this looks good and grep shows me that you spotted all .command() occurrences

src/commandio.h
t-wissmann: Indeed, thanks!

src/metacommands.cpp
t-wissmann: Fixed, thanks!

t-wissmann:
> Two minor findings, other than that, this looks good and grep shows me that you spotted all .command() occurrences

Thanks, I've fixed them. Regarding `command()`, I had temporarily renamed `Input::command()` and `CallOrCommand::command()` and went through all compile errors :)

src/commandio.h
blarz: Thanks!

blarz (approved):
Looks good to me!

PR #1283, opened 2021-04-17
Remove alpha-value from colors in panel.sh

### LINKED ISSUE #1282: dzen: error, cannot allocate color '#101010cc'
Hi,
just getting started with hlwm (thank you for it).
I've installed the -git version from arch, and have the autostart copied and working okay.  I cannot get the panel to work though.  on my laptop with no external monitor:

```
% sh .config/herbstluftwm/panel.sh 0
dzen: error, cannot allocate color '#101010cc'
% 
```

any ideas how to start?  I've not changed any color configs so I think that color is 

```
% hc get frame_border_normal_color 
#101010cc
```

t-wissmann:
In panel.sh, we extract the current colors from the hlwm settings.
However, the latter possibly have alpha-values (some even have in the
default autostart). So, we need to remove these alpha-values, turning the
rgba-format into rgb.

This fixes #1282

### DIFF
--- a/share/panel.sh
+++ b/share/panel.sh
@@ -19,8 +19,9 @@ y=${geometry[1]}
 panel_width=${geometry[2]}
 panel_height=16
 font="-*-fixed-medium-*-*-*-12-*-*-*-*-*-*-*"
-bgcolor=$(hc get frame_border_normal_color)
-selbg=$(hc get window_border_active_color)
+# extract colors from hlwm and omit alpha-value
+bgcolor=$(hc get frame_border_normal_color|sed 's,^\(\#[0-9a-f]\{6\}\)[0-9a-f]\{2\}$,\1,')
+selbg=$(hc get window_border_active_color|sed 's,^\(\#[0-9a-f]\{6\}\)[0-9a-f]\{2\}$,\1,')
 selfg='#101010'
 
 ####


tiosgz:
Do you plan to use an output format different from `#rrggbbaa` in the future? `var="${var:0:7}"` seems simpler to me (but requires two lines unless it can be done on strings as well).

t-wissmann:
I first thought about that too. If someone sets the color to a named color like `orange`, I'm not sure what the value should be when reading it. Currently it reading it will be `#ffa500`, but maybe `orange` would be more helpful.

tiosgz:
Hmm, that might make quite a great problem for scripts that can’t do this name-hex conversion. Anyway, if you are considering changes in that direction, then I’m for keeping the panel more future-proof.

t-wissmann:
That's indeed a strong argument for keeping the output in the `#rrggbb(aa)` format, and you convinced me to not change it. Anyway, it feels more robust to use the regular expression here instead of just dropping everything from index 7 on. (There is no performance impact, `sed` is only called when the panel starts up)

blarz (approved):
Fix looks good!

Could we have spotted this issue earlier? Do we need a test for the "default setup", i.e. default autostart and dzen2, so we know that it always works out of the box?

t-wissmann:
I think it's very cumbersome to verify the panel. And in most cases, the main issue with the panel not starting  up is some incombatibility between dzen2, textwidth, and xft font names. So I'm not sure whether it's worth the effort. Of course, if you want to write a test, please go ahead (but I probably will not)

PR #1285, opened 2021-04-18
Extend IPC protocol by error channel

t-wissmann:
This makes the ipc protocol and herbstclient support the error channel.
In order to make herbstclient compatible with old herbstluftwm ipc
server implementations, the error channel support is indicated by an
atom on hlwm's ipc server window (which is so far only used for the hook
system). Conversely, an old herbstclient version should also work more
or less with the new herbstluftwm ipc server: it simply ignores the
error channel.

Due to this change, a few more missing `output.perror()` calls surfaced,
affecting implementation and test cases.

In order to test the new herbstclient, the test suite features a
minimalist ipc server implementation which can be used to simulate wrong
behaviour and thus to stress the error handling in the client.

### DIFF
--- a/ipc-client/ipc-client.h
+++ b/ipc-client/ipc-client.h
@@ -32,7 +32,7 @@ bool hc_create_client_window(HCConnection* con);
  * @param the returned error channel (needs to be free'd by the caller on success)
  * @param the returned exit status
  * @return whether the reply was received successfully.
- * if false is returned, then none of the ret_⎄-pointers is set.
+ * if false is returned, then none of the ret_...-pointers is set.
  */
 bool hc_send_command(HCConnection* con, int argc, char* argv[],
                      char** ret_out, char** ret_err, int* ret_status);
--- a/ipc-client/ipc-client.h
+++ b/ipc-client/ipc-client.h
@@ -23,8 +23,19 @@ void hc_disconnect(HCConnection* con);
 /* ensure there is a client window for sending commands */
 bool hc_create_client_window(HCConnection* con);
 
+/**
+ * @brief send a command and wait for its reply to arrive
+ * @param the connection to hlwm
+ * @param number of arguments of the ipc call
+ * @param the arguments of the ipc call
+ * @param the returned output channel (needs to be free'd by the caller on success)
+ * @param the returned error channel (needs to be free'd by the caller on success)
+ * @param the returned exit status
+ * @return whether the reply was received successfully.
+ * if false is returned, then none of the ret_⎄-pointers is set.
+ */
 bool hc_send_command(HCConnection* con, int argc, char* argv[],
-                     char** ret_out, int* ret_status);
+                     char** ret_out, char** ret_err, int* ret_status);
 
 bool hc_hook_window_connect(HCConnection* con);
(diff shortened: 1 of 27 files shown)


t-wissmann:
I think, now it is ready :)

tests/test_rules.py
```diff
@@ -140,10 +142,11 @@ def test_rule_parse_error_printed_at_client(hlwm, command, rulearg, errormsg):
     else:
         full_cmd += [rulearg]
 
-    proc = hlwm.call(full_cmd)
-    # since the command does not fail, the error
+    proc = hlwm.unchecked_call(full_cmd)
+
+    # the command does not fail, but the error
     # message appears on stdout
```
blarz, line 148: stderr?

tests/test_herbstclient.py
```diff
@@ -220,3 +225,175 @@ def test_ensure_newline(hlwm):
                            universal_newlines=True)
     assert proc1.stdout == "test\n"
     assert proc2.stdout == "test\n"
+
+
+class IpcServer:
+    """A simply re-implementation of the ipc server
```
blarz, line 231: simple

src/ipc-protocol.h
```diff
@@ -5,9 +5,17 @@
 //#define HERBST_IPC_READY "HERBST_IPC_READY"
 //#define HERBST_IPC_ATOM  "_HERBST_IPC"
 #define HERBST_IPC_ARGS_ATOM "_HERBST_IPC_ARGS"
+//! the (utf8 text) atom containing the output channel
 #define HERBST_IPC_OUTPUT_ATOM "_HERBST_IPC_OUTPUT"
+//! the (utf8 text) atom containing the error channel
+#define HERBST_IPC_ERROR_ATOM "_HERBST_IPC_ERROR"
 #define HERBST_IPC_STATUS_ATOM "_HERBST_IPC_EXIT_STATUS"
 
+/** if this (integer) atom present and non-empty on the hook window,
```
blarz, line 14: "is" missing?

ipc-client/ipc-client.h
```diff
@@ -23,8 +23,19 @@ void hc_disconnect(HCConnection* con);
...
+/**
+ * @brief send a command and wait for its reply to arrive
+ * @param the connection to hlwm
+ * @param number of arguments of the ipc call
+ * @param the arguments of the ipc call
+ * @param the returned output channel (needs to be free'd by the caller on success)
+ * @param the returned error channel (needs to be free'd by the caller on success)
+ * @param the returned exit status
+ * @return whether the reply was received successfully.
+ * if false is returned, then none of the ret_⎄-pointers is set.
```
blarz, line 35: ret_⎄ strange unicode garbage?

blarz (changes requested):
Minor findings

blarz:
Not related to this change but maybe for another PR:
```
out << "error: Command \"" << args.command() << "\" not found" << endl;
```
in command.cpp could also be moved to stderr somehow?

t-wissmann:
Thanks for looking through the long diff :-)
I've fixed all the points you mentioned and created a new PR for the "command not found" error: #1286

blarz (approved):
Looks good to me, played around with it locally, too

PR #1286, opened 2021-04-21
Write 'command not found' to the error channel

t-wissmann:
This is yet another error message that belongs to the new error channel.


blarz (approved):
Thanks! <3

blarz (approved):

PR #1287, opened 2021-04-21
Print error message on too many arguments

t-wissmann:
If ArgParse::parsingAllFails() notices too many arguments, print a
message to the error channel. This affects e.g. 'hc help'.


blarz (approved):
Makes sense, use stderr everywhere :)

blarz (approved):

PR #1290, opened 2021-04-21
Test for commands: object_tree, new_attr, sprintf

t-wissmann:
This adds a couple of more tests for the meta commands object_tree,
new_attr, and sprintf. This mainly concerns error handling and
completion. This should improve the test coverage.

### DIFF
--- a/src/metacommands.cpp
+++ b/src/metacommands.cpp
@@ -393,7 +393,7 @@ static std::map<string, function<Attribute*(string)>> name2constructor {
 Attribute* MetaCommands::newAttributeWithType(string typestr, string attr_name, Output output) {
     auto it = name2constructor.find(typestr);
     if (it == name2constructor.end()) {
-        output << "error: unknown type \"" << typestr << "\"";
+        output.perror() << "unknown type \"" << typestr << "\"";
         return nullptr;
     }
     auto attr = it->second(attr_name);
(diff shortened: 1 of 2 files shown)


blarz (dismissed):
It improves coverage indeed!

t-wissmann:
It turned out a couple of error messages had to be moved to the error channel.

blarz (approved):

src/metacommands.cpp
```diff
@@ -393,7 +393,7 @@ static std::map<string, function<Attribute*(string)>> name2constructor {
 Attribute* MetaCommands::newAttributeWithType(string typestr, string attr_name, Output output) {
     auto it = name2constructor.find(typestr);
     if (it == name2constructor.end()) {
-        output << "error: unknown type \"" << typestr << "\"";
+        output.perror() << "error: unknown type \"" << typestr << "\"";
```
blarz, line 396: "error:" can probably be removed?

blarz (changes requested):

src/metacommands.cpp
blarz: (Which probably is the reason for the merge conflict)

t-wissmann: Indeed, thanks! (it's resolved now)

blarz: Thanks!

blarz (approved):

PR #1291, opened 2021-04-23
client: Remove unused methods

blarz:
Found via the coverage report


t-wissmann (approved):
Nice! It's good to get rid of these! (I guess these became obsolete when commands like `raise` and `close` were migrated to ArgParse/Converter)

PR #1292, opened 2021-04-25
Test apply_tmp_rules invalid client error

blarz:
This was tested for apply_rules already, but not for apply_tmp_rule,
which was easily added.


t-wissmann (approved):
Thanks!

PR #1293, opened 2021-04-25
utils: Remove unused method

blarz:
posix_sh_compress_inplace seems unused, posix_sh_escape on the other
hand is not.


t-wissmann (approved):
Thanks!

PR #1294, opened 2021-04-26
Test herbstclient -0 --idle

t-wissmann:
Test the -0 command line option of herbstclient. Unfortunately, this
still uses tab characters to separate the arguments.

To avoid the race condition when starting herbstclient --idle, I'm
making the HcIdle implementation in conftest.py capable of -0 and simply
use this in the new test cases.

### DIFF
--- a/tests/test_herbstclient.py
+++ b/tests/test_herbstclient.py
@@ -148,7 +148,7 @@ def test_lastarg_only(hlwm, repeat):
     assert proc.returncode == 0
 
 
-@pytest.mark.parametrize('zero_separated', [True, False])  # use more repetitions to test the race-condtion
+@pytest.mark.parametrize('zero_separated', [True, False])
 def test_zero_byte_separator(hlwm, zero_separated):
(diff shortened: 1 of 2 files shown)


tests/test_herbstclient.py
```diff
@@ -147,6 +148,22 @@ def test_lastarg_only(hlwm, repeat):
     assert proc.returncode == 0
 
 
+@pytest.mark.parametrize('zero_separated', [True, False])  # use more repetitions to test the race-condtion
```
blarz, line 151: Remove the "use more repetitions" comment as it doesn't apply?

blarz (changes requested):
Just one minor thing

tests/test_herbstclient.py
t-wissmann: Right, thanks! It's fixed now.

blarz: Thanks!

blarz (approved):

PR #1303, opened 2021-05-02
Handle text properties more robustly

t-wissmann:
This improves reading and writing of utf8 string list X11-properties,
and makes hlwm work even if the locale is partly broken, e.g. on some
fresh Manjaro installations (#1276).

When writing string lists in XConnection::setPropertyString(), we first
try the utf8-function, and if it fails, fall back to `XStdICCTextStyle`,
which seems to work even in the above mentioned scenario. Unfortunately,
I don't know how to reproduce this scenario in the CI, so this is not
covered code in the CI. (Here, we would need a test case for which
XSupportsLocale() in main.cpp fails).

Reading string lists is now centralized in
XConnection::getWindowPropertyTextList(). Formerly, we used
Xutf8TextPropertyToTextList() here, which unfortunately failed in the
setup of #1276. Interestingly, tools like `xprop` don't use it at all,
and instead call the plain `XGetWindowProperty()` and parse the string
list from there, and indeed this seems to work with or without working
locales (and it's tested thoroughly by the CI). For the sake of avoiding
redundancy, we can use this implementation in the IpcServer (it formerly
duplicated the source code) and when reading ordinary utf8 string
properties (XConnection::getWindowProperty()).

### DIFF
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -398,11 +398,12 @@ std::experimental::optional<vector<string>>
         return {};
     }
     if (format != 8) {
-        fprintf(stderr, "herbstluftwm error: can not parse the "
+        fprintf(stderr, "herbstluftwm: error: can not parse the "
                         " atom \'%s\' of window 0x%lx: expected format=8 but got"
                         " format=%d\n",
                         XGetAtomName(m_display, property), window, format);
         XFree(items_return);
+        return {};
     }
     unsigned long offset = 0;
     vector<string> arguments;
@@ -435,7 +436,12 @@ std::experimental::optional<vector<string>>
                 arguments.push_back(*list);
                 XFreeStringList(list);
             } else {
-                fprintf(stderr, "herbstluftwm warning: can not read text property\n");
+                fprintf(stderr, "herbstluftwm: error: can not parse the "
+                                " atom \'%s\' of window 0x%lx: unknown text format \'%s\'\n",
+                                XGetAtomName(m_display, property),
+                                window,
+                                XGetAtomName(m_display, prop_type));
+                return {};
             }
         }
         // skip the string, and skip the null-byte
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -393,7 +393,7 @@ std::experimental::optional<vector<string>>
     int status = XGetWindowProperty(m_display, window,
             property, 0, ULONG_MAX, False, AnyPropertyType,
             &prop_type, &format, &count, &bytes_left,
-            (unsigned char**)&items_return);
+            &items_return);
     if (Success != status || prop_type == None || format == 0) {
         return {};
     }
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -406,7 +406,10 @@ std::experimental::optional<vector<string>>
     }
     unsigned long offset = 0;
     vector<string> arguments;
-    while (offset < count) {
+    // the trailing 0 at items_return[count] might be crucial:
+    // if the string list ends with the empty string, then we
+    // need to access items_return[count].
+    while (offset <= count) {
         char* textChunk = items_return + offset;
         // let us hope that items_return is properly null-byte terminated.
         unsigned long textChunkLen = strlen(textChunk);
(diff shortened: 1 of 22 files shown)


src/xconnection.cpp
```diff
@@ -389,22 +385,63 @@ std::experimental::optional<vector<Window>>
...
+            textprop.value = reinterpret_cast<unsigned char*>(textChunk);
+            int n = 0;
+            char** list = nullptr;
+            if (XmbTextPropertyToTextList(m_display, &textprop, &list, &n) == Success
+                && n > 0 && *list)
+            {
+                arguments.push_back(*list);
+                XFreeStringList(list);
+            } else {
+                fprintf(stderr, "herbstluftwm warning: can not read text property\n");
```
blarz, line 438: Is the mix of fprintf and std::cerr and HSDebug in this file intended?
Also "herbstluftwm warning" deviates from the rest of fprintf's patterns,
usually you see "herbstluftm: warning" or something?

src/xconnection.cpp
```diff
@@ -389,22 +385,63 @@ std::experimental::optional<vector<Window>>
...
+    unsigned long offset = 0;
     vector<string> arguments;
-    for (int i = 0; i < count; i++) {
-        arguments.push_back(list_return[i]);
+    // the trailing 0 at items_return[count] might be crucial:
+    // if the string list ends with the empty string, then we
+    // need to access items_return[count].
+    while (offset <= count) {
+        char* textChunk = items_return + offset;
+        // let us hope that items_return is properly null-byte terminated.
```
blarz, line 414: This sounds unsettling :D

blarz (changes requested):

tests/test_herbstclient.py
```diff
@@ -414,3 +414,18 @@ def reply_output():
     assert re.search(f'could not get window property "{faulty_name}"', hc.reply.stderr)
     assert not re.search('EEE', hc.reply.stderr)
     assert hc.reply.returncode == 1
+
+
+def test_command_tokenization(hlwm):
```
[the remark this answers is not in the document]
t-wissmann: The test is related, because it ensures that it's `<=` in `while (offset <= count)`. Maybe `tokenization` isn't the best word for it, because it's not about tokenization on the command line but about tokenization when reading the x11-property (there's no command line tokenization at all by the way)

src/xconnection.cpp
```diff
@@ -389,22 +385,63 @@ std::experimental::optional<vector<Window>>
 std::experimental::optional<vector<string>>
     XConnection::getWindowPropertyTextList(Window window, Atom property)
 {
-    XTextProperty text_prop;
-    if (!XGetTextProperty(m_display, window, &text_prop, property)) {
+    Atom prop_type;
+    int format;
+    unsigned long bytes_left;
+    char* items_return; // TODO: char* or unsigned char*?
```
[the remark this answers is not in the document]
t-wissmann: `XGetWindowProperty` actually expects a `unsigned char**`, so I'll change it to `unsigned` and added multiple `reinterpret_cast`s where `items_return` is used later.

src/xconnection.cpp
t-wissmann: With these long format strings, I find fprintf more readable, what do you think?

src/xconnection.cpp
t-wissmann: I think I once read in the docs that `items_return` always returns one element more than `count` and that it's always null, but I couldn't find it anymore when writing this PR.

t-wissmann:
I've also added test cases for the two error messages.

src/xconnection.cpp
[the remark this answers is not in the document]
blarz: Thanks!

src/xconnection.cpp
blarz: Fingers crossed...

src/xconnection.cpp
blarz: I'm fine with either, just wanted to check if the "inconsistency" was on purpse.

blarz (approved):
Tested locally, works for me, let's hope it fixes those Manjaro issues

PR #1306, opened 2021-05-04
Wait passively for hc in test suite ipc server

t-wissmann:
This tries to fix a race condition in the herbstclient tests with the
test suite IpcServer class that happened quite regurarly in the CI:
according to the log the herbstclient did not boot up (in time).

Instead of waiting for a certain amount of seconds, the IpcServer class
now waits for a certain number of X11 events for the herbstclient to
show up. The downside is that on failure, the CI hangs until the CI
itself times out, but hopefully this does not happen at all.

### DIFF
--- a/tests/test_herbstclient.py
+++ b/tests/test_herbstclient.py
@@ -4,7 +4,6 @@ import os
 import re
 import pytest
 import sys
-import time
 import contextlib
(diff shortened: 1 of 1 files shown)


t-wissmann:
I'd say: let us see how it improves the CI robustness :)

PR #1308, opened 2021-05-05
Test shift/resize up/left and focus cycling

t-wissmann:
This adds some more tests to the floating commands that have not been
tested yet according to the coverage report: So far, shift and
resize for 'left' and 'up' have not been tested. Also, directional focus
also switches between windows with identical geometry.

### DIFF
--- a/tests/test_floating.py
+++ b/tests/test_floating.py
@@ -1,4 +1,5 @@
 import pytest
+from herbstluftwm.types import Rectangle
 
 
 @pytest.mark.parametrize('clientFocused', [0, 1, 2, 3])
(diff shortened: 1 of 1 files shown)


tests/test_floating.py
```diff
@@ -84,6 +85,62 @@ def test_directional_shift_right(hlwm, x11, start_y_rel, hits_obstacle):
...
+        # expected position after shift up: same x, different y
+        expected_pos = (160, obstacle_pos + bw + gap)
+    elif direction == 'left':
+        # expected position after shift left: different x, same y
+        expected_pos = (obstacle_pos + bw + gap, 170)
+
+    hlwm.call(['jumpto', client])
+    hlwm.call([command, direction])  # the actual shift/resize
+
+    # check that we top left corner of the window is adjusted as expected
```
blarz, line 131: s/we/the/?

blarz (changes requested):
Just a tiny typo in a comment

tests/test_floating.py
t-wissmann: [x] fixed, thanks!

blarz: Thanks!

blarz (dismissed):

blarz (approved):

PR #1312, opened 2021-05-13
Doc on rule floatplacement vs floating_geometry

t-wissmann:
This documents and tests the interaction of the floatplacement and
floating_geometry rule consequences. Since 'floatplacement' is not
applied by 'apply_rules' / 'apply_tmp_rules', the test case only
mentions the 'rule' command (also I've fixed a typo along the way).

For the readability in the tests, I've introduced a Point datastructure
to herbstluftwm's python bindings.


blarz (approved):
Perfect, thanks! That should clear things up :)

Issue #1014, opened 2020-10-14
How to solve systray + minimization problem


etosan:
Due to work and stuffs I started using some software packages that have global hotkeys support and are supposed to go to systray. I run no systray bar, so there is no place for software to go, thus if one sends it close event, instead of "backgrounding" the software terminates and thus stops handling hotkeys.

If one wants the global hotkeys to be handled, software must be running. This is fine, usage wise, because it allows given window to be "brought" into focused monitor/frame, but then this window constantly occupies one frame, still.

What is proper solution for tilers to handle such cases? Would running systray capable bar app solve this?

Would it be possible for herbstlufwm "minimize" (is minimize equal to "unmap"?) window instead, so that it can be brought (with autoraise on urgent or manually somehow) later on without need for systray?

t-wissmann:
> What is proper solution for tilers to handle such cases? Would running systray capable bar app solve this?

Yes, running a bar with systray solves it (e.g. xfce4-panel or lxpanel).

> Would it be possible for herbstlufwm "minimize" (is minimize equal to "unmap"?) window instead, so that it can be brought (with autoraise on urgent or manually somehow) later on without need for systray?

I'd like to implement minimization, but it isn't in hlwm yet.

t-wissmann:
In the current git version (since #1032 / #1033), hlwm supports window minimization! You can minimize through a panel (like lxpanel or xfce4-panel or tint2, ...) or through a keybinding like
```
hc keybind $Mod-Shift-m set_attr clients.focus.minimized true
```
To un-minimize a window, use the bar again or select the window with any other kind of menu (e.g. rofi) or pass its window ID to the `jumpto` command.

etosan:
I will try git master, hopefully, this evenning.

federicotorrielli:
> In the current git version (since #1032 / #1033), hlwm supports window minimization! You can minimize through a panel (like lxpanel or xfce4-panel or tint2, ...) or through a keybinding like
> 
> ```
> hc keybind $Mod-Shift-m set_attr clients.focus.minimized true
> ```
> 
> To un-minimize a window, use the bar again or select the window with any other kind of menu (e.g. rofi) or pass its window ID to the `jumpto` command.

Hi, you should really put this into the documentation! It's very useful!

t-wissmann:
[...]

PR #1313, opened 2021-05-13
Client alias 'longest-minimized' and 'last-minimized'

t-wissmann:
This adds client aliases for accessing the longest and the most recently
minimized client, and thus provides a native way to un-minimize clients
with key bindings (either in a lifo or fifo way). Thus, this also works
when minimizing clients via the panel and thus is more universal than
restricting to the unminimize.sh script (which was written for the lifo
unminimize).

With the new feature, we do not need the complicated binding for
unminimizing all clients anymore and we can move it to the faq (and
simplify it a bit).

This solves another feature request raised in the discussion of #1014.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -33,6 +33,8 @@ Current git version
     settings for compatibility).
   * New object 'types' containing documentation on (attribute-) types.
   * New command 'attr_type' printing the type of a given attribute.
+  * New client alias 'last-minimized' and 'longest-minimized' for focusing
+    and unminimizing minimized clients.
   * Relative values for integer attributes ('+=N' and '-=N')
   * The 'cycle' command now also cycles through floating windows.
   * The 'rule' command now reports errors already during rule creation.
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1416,6 +1416,10 @@ follows:
   - +urgent+ references some window that is urgent.
   - +0x+'HEXID' -- where 'HEXID' is some hexadecimal number -- references the
     window with hexadecimal X11 window id 'HEXID'.
+  - +longest-minimized+ references the minimized window on the focused tag
+    that has been minimized for the longest time.
+  - +latest-minimized+ references the minimized window on the focused tag
+    that has been minimized most recently.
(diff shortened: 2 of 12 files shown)


t-wissmann:
Should I remove the other un-minimize shortcut (Mod-Ctrl-m) that unminimizes all windows on a tag?

blarz:
> Should I remove the other un-minimize shortcut (Mod-Ctrl-m) that unminimizes all windows on a tag?

I actually like that shortcut, at least the code that is behind it. But maybe we can remove/replace it as soon as #1262 is merged?

Also we could switch the new jumto last-minimized and the unminimize all shortcut?
Mod-Ctrl-Shift-m is quite a fingerbreaker (well, maybe not for emacs users :sweat_smile:)

t-wissmann:
> > Should I remove the other un-minimize shortcut (Mod-Ctrl-m) that unminimizes all windows on a tag?
> 
> I actually like that shortcut, at least the code that is behind it. But maybe we can remove/replace it as soon as #1262 is merged?

I don't see the connection between #1262 and the keybinding for unminmizing all clients. #1262 manually implements `jumpto last-minimized`.

blarz:
> > > Should I remove the other un-minimize shortcut (Mod-Ctrl-m) that unminimizes all windows on a tag?
> > 
> > 
> > I actually like that shortcut, at least the code that is behind it. But maybe we can remove/replace it as soon as #1262 is merged?
> 
> I don't see the connection between #1262 and the keybinding for unminmizing all clients. #1262 manually implements `jumpto last-minimized`.

Err, you're absolutely right, nevermind then!

t-wissmann:
But we could bind `jumpto last-minimized` to `Mod-Ctrl-m` and move the `foreach` loop to an faq entry.

blarz:
> But we could bind `jumpto last-minimized` to `Mod-Ctrl-m` and move the `foreach` loop to an faq entry.

I like it! That way the foreach is not lost and the FAQ is a good place to "store" it :)

t-wissmann:
I've moved it.

blarz (approved):
Looks good to me, tested locally, test cases are also nice IMHO!

t-wissmann:
Thanks!

PR #1321, opened 2021-05-15
Execute test suites in random order

blarz:
Just want to see if something crashes


blarz:
Just wanted to try it out. Not sure if we actually want that in the CI...

PR #1324, opened 2021-05-22
Update floating geometry if size hints change

### LINKED ISSUE #1322: Window border thicker on sides for floating window with floating_geometry rule or other specific pixel size method
I've noticed that if I spawn a floating window with a specific pixel size either with the floating_geometry rule or through xdotool, the window will have one side of it's border be incorrect i.e. thicker than what it is suppose to be. I tried `hc attr clients.focus.sizehints_floating false` and that made no difference. Here's a screenshot to illustrate:

This is with urxvt windows, so my guess is that urxvt expects to be a specific incremental size rather than pixel size. If there is a way already, I'm not sure how to make herbstluftwm go ahead and respect that size despite some other method requesting an exact pixel size

t-wissmann:
If the size hints of a client change, then its floating geometry must be
updated. Also, this must happen even if the client is not floated at the
moment. To accomplish the latter, the applysizehints() function is
extended by an optional `force` parameter. The new test case verifies
both the floating and non-floating case.

This fixes #1322.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -343,10 +343,10 @@ void Client::resize_tiling(Rectangle rect, bool isFocused, bool minimalDecoratio
 
 /**
  * @brief Update the given window size according to the client's size hints
- * @param width
- * @param height
- * @param If set, always apply the size hints. If not set, only
- * apply the size hints if the according sizehints_/floating_sizehints_tiling_
+ * @param w width
+ * @param h height
+ * @param force If set, always apply the size hints. If not set, only
+ * apply the size hints if the according sizehints_floating_ / sizehints_tiling_
  * attribute is set
  * @return whether the size changed
  */
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -334,8 +334,16 @@ void Client::resize_tiling(Rectangle rect, bool isFocused, bool minimalDecoratio
     dec->resize_outline(rect, scheme);
 }
 
-// from dwm.c
-bool Client::applysizehints(int *w, int *h) {
+/**
+ * @brief Update the given window size according to the client's size hints
+ * @param width
+ * @param height
+ * @param If set, always apply the size hints. If not set, only
+ * apply the size hints if the according sizehints_/floating_sizehints_tiling_
+ * attribute is set
+ * @return whether the size changed
+ */
+bool Client::applysizehints(int* w, int* h, bool force) {
     bool baseismin;
 
     /* set minimum possible */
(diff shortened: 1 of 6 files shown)


src/client.cpp
```diff
@@ -334,8 +334,16 @@ void Client::resize_tiling(Rectangle rect, bool isFocused, bool minimalDecoratio
     dec->resize_outline(rect, scheme);
 }
 
-// from dwm.c
-bool Client::applysizehints(int *w, int *h) {
+/**
+ * @brief Update the given window size according to the client's size hints
+ * @param width
```
blarz, line 339: w

src/client.cpp
```diff
@@ -334,8 +334,16 @@ void Client::resize_tiling(Rectangle rect, bool isFocused, bool minimalDecoratio
...
+ * @param height
```
blarz, line 340: h

src/client.cpp
```diff
@@ -334,8 +334,16 @@ void Client::resize_tiling(Rectangle rect, bool isFocused, bool minimalDecoratio
...
+ * @param If set, always apply the size hints. If not set, only
```
blarz, line 341: @param force

blarz (changes requested):
Just some remarks on docstrings, other than that, it looks good to me.
(NEWS might need a merge conflict resolve)

src/client.cpp
blarz: Oh, we don't actually state the name of the parameter here again, so "width" is the description of the parameter 'w'. Resolving this.

src/client.cpp
blarz: Nevermind

src/client.cpp
blarz: Nevermind

t-wissmann:
Regarding restating the parameter name after `@param`: I'm not entirely sure how doxygen defines it. But I'd guess that for avoiding redundancy, one does not need to restate it because the doxygen generator can mention the parameter name next its description automatically.

t-wissmann:
Oh no, it seems I am wrong and one must restate the parameter name. Running `doxygen` yields e.g.:
image

t-wissmann:
I've fixed it for the `applysizehints()` function:
image

blarz:
> I've fixed it for the `applysizehints()` function:
> image

Thanks for looking into that and fixing it right away! :)

blarz (approved):

PR #1325, opened 2021-05-23
Set X11 property for client's floating state

### LINKED ISSUE #1172: [REQUEST] Create a new atom to detect whether window is floating
I'd like to request an atom for herbstluftwm to report whether a window is floating or not, similar to i3's `I3_FLOATING_WINDOW`. This would allow specifying a picom rule to exclude shadows, fade, blur, etc for either floating or non-floating windows

t-wissmann:
The purpose of this change is to indicate a clients effective floating
state via X11 properties to make use of it in rules of X11 compositors.

The change happens in two steps: The first step is a new read-only
attribute `floating_effectively` which combines the client's single
window floating state with its tag's floating state. This attribute is
tested semantically. The second step is simply a signal connection that
sets or removes the X11 properties HLWM_FLOATING_WINDOW and
HLWM_TILING_WINDOW, depending on the value of `floating_effectively`.

This closes #1172.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,12 @@ herbstluftwm NEWS -- History of user-visible changes
 See the link:MIGRATION[] file for changes that introduced incompatibility to
 older versions.
 
+Current git version
+-------------------
+
+  * New client attribute 'floating_effectively' and associated X11 properties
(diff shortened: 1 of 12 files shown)


blarz (approved):
Looks good to me! :)

t-wissmann:
Thanks!

PR #1328, opened 2021-05-28
Move text drawing to own function

t-wissmann:
Encapsulate the text drawing in a separate function, the existing test
cases should cover most of it.


blarz (approved):
Looks good to me, redrawPixmap is already quite long.
Did some minor rewording of the commit message

PR #1329, opened 2021-05-31
Shorten window title to decoration width

t-wissmann:
Shorten a client's title such that it is at most as wide as the window
content, i.e. such that it does not reach into window border.

The new tests verify this by looking for the leftmost and rightmost
pixels belonging to the title. Unfortunately, proper utf8 characters
work in the test suite only with xft fonts. But in a manual xephyr
session, the plain X fonts also work with utf8 characters.

### DIFF
--- a/src/fontdata.cpp
+++ b/src/fontdata.cpp
@@ -81,7 +81,7 @@ void FontData::initFromStr(const string& source)
  * @param glyphs Only consider that many glyphs from the given text
  * @return The width in pixels
  */
-int FontData::textwidth(const std::string& text, size_t len) const
+int FontData::textwidth(const string& text, size_t len) const
 {
     if (!s_xconnection) {
         return 0;
--- a/src/fontdata.cpp
+++ b/src/fontdata.cpp
@@ -78,7 +78,8 @@ void FontData::initFromStr(const string& source)
 /**
  * @brief compute the with of the given text
  * @param text The text
- * @param glyphs Only consider that many glyphs from the given text
+ * @param len Only consider that many glyphs or bytes from the given text,
+ *            depending on the concrete font type
  * @return The width in pixels
  */
 int FontData::textwidth(const string& text, size_t len) const
@@ -103,5 +104,5 @@ int FontData::textwidth(const string& text, size_t len) const
     if (xFontStruct_) {
         return XTextWidth(xFontStruct_, text.c_str(), len);
     }
-    return -1;
+    return 0;
 }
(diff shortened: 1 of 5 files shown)


src/fontdata.cpp
```diff
@@ -74,3 +74,34 @@ void FontData::initFromStr(const string& source)
                 string("cannot allocate font \'") + source + "\'");
     }
 }
+
+/**
+ * @brief compute the with of the given text
+ * @param text The text
+ * @param glyphs Only consider that many glyphs from the given text
```
blarz, line 81: the second parameter of this method is "len"

src/fontdata.cpp
```diff
@@ -74,3 +74,34 @@ void FontData::initFromStr(const string& source)
...
+    }
+    if (xFontSet_) {
+        XRectangle logical;
+        Xutf8TextExtents(xFontSet_, text.c_str(), static_cast<int>(len), nullptr, &logical);
+        return logical.width;
+    }
+    if (xFontStruct_) {
+        return XTextWidth(xFontStruct_, text.c_str(), len);
+    }
+    return -1;
```
blarz, line 106: -1 or 0? If there is no x connection you chose 0, in the else case it's -1.

tests/test_decorations.py
```diff
@@ -117,6 +123,59 @@ def test_title_different_letters_are_drawn(hlwm, x11, font):
...
+    if font[0] != '-':
+        # for xft fonts, also test utf8 window titles
+        utf8titles = [w * '♥']
+    else:
+        # for plain X fonts, it does not seem to work in tox/pytest
+        # (but strangely, it works in a manual Xephyr session)
+        utf8titles = []
+
+    for title in [w * '=', w * '|'] + utf8titles:
+        print("title = " + title)
```
blarz, line 152: debug leftover?

blarz (changes requested):
Tested locally works fine, just minor issues

src/fontdata.cpp
t-wissmann: I've fixed it, thanks! It seems that depending on the font used, this could be either the number of glyphs or the number of bytes, so I've also updated the description accordingly.

src/fontdata.cpp
t-wissmann: Thanks! this should be '0'. I've only changed it to -1 for debugging. :-)

tests/test_decorations.py
t-wissmann: Yes, thanks; it's gone :)

src/fontdata.cpp
blarz: Thanks!

src/fontdata.cpp
blarz: Thanks :)

blarz (approved):

PR #1330, opened 2021-05-31
Migrate spawn and wmexec to Input and add tests

t-wissmann:
The commands spawn and wmexec were the last that used the old char**
argv interface. For 'spawn', test cases existed already, for 'wmexec',
I've added some more test cases.

### DIFF
--- a/tests/test_herbstluftwm.py
+++ b/tests/test_herbstluftwm.py
@@ -3,8 +3,9 @@ import os
 import pytest
 import subprocess
 import textwrap
-from conftest import BINDIR, HlwmProcess
+from conftest import BINDIR, HlwmProcess, HlwmBridge
 import conftest
+import os.path
 from Xlib import X, Xatom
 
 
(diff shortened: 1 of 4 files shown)


tests/test_herbstluftwm.py
```diff
@@ -21,6 +22,46 @@ def test_reload(hlwm_process, hlwm):
...
+                            read_hlwm_output=False)
+    assert p.returncode == 0
+    hlwm_process.read_and_echo_output(until_stdout='hlwm started')
+
+    assert hlwm.attr.settings.snap_gap() != 13
+    if with_client:
+        assert winid in hlwm.list_children('clients')
+
+
+@pytest.mark.parametrize("with_client", [True, False])
```
blarz, line 45: Do we actually need this parameter? We have to create a client anyway as you said in line 50

blarz (commented):
Just one question, otherwise looks good! I like the wmexec touch test :D

tests/test_herbstluftwm.py
t-wissmann: I've removed the comment about the necessity for a client (and the client itself), because the client isn't necessary here (the client was only necessary in the already existing wmexec test case `test_ewmh::test_tags_restored_after_wmexec` where the preservation of the tag names is tested: there, xvfb clears all X properties if there is no client)

blarz: Cool, thanks for removingit!

blarz (approved):

PR #1331, opened 2021-05-31
Update cursor shape and resize accordingly

### LINKED ISSUE #1138: Option to change cursor when mouse focuses window-border
Changing the cursor when the mouse focuses a window-border as an visual indicator could be helpful in the following scenarios:

1. Grabbing thin-borders
2. Preventing accidentally grabbing borders. For example I sometimes try to grab the scrollbar of a window but accidentally grab the window border and resize it instead of scrolling.

The only issue I see is that if you would drag the mouse over the screen each time the mouse crosses a border it would change the cursor. Some people could find it annoying. It probably should be optional. A timer could be a solution such that the cursor doesnt change immediately.

An alternative solution would be instead of changing the cursor an option could be introduced to disable resizing a window by dragging the window-border.

t-wissmann:
When the mouse cursor is at a client's window border, the cursor shape
now changes to indicate the direction in which the window resizes.
Also, I've updated the resize code itself to allow resizing where the
width or height is fixed and the client is resized only into one
dimension.

Unfortunately, the cursor preview isn't entirely correct for tiled
windows. Also does not seem to be possible to query the cursor shape in
test cases.

This closes #1138.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,7 @@ Current git version
   * New client attribute 'floating_effectively' and associated X11 properties
     'HLWM_FLOATING_WINDOW' and 'HLWM_TILING_WINDOW'
   * New read-only client attribute 'decoration_geometry'.
+  * The cursor shape now indicates resize options.
   * Bug fixes:
     - Update floating geometry if a client's size hints change
 
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -25,6 +25,9 @@
 #include "utils.h"
 #include "xconnection.h"
 
+using std::pair;
+using std::vector;
+using std::shared_ptr;
 using std::string;
 using std::stringstream;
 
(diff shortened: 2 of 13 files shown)


blarz:
Played around with it for a bit and it works very well IMHO!

You're right, in tiling mode it is bit strange (especially when it shows the diagonal arrow but you can obviously only move into horizontal/vertical directions). In floating mode, everything looks good!

Yes, testing this change is hard if you cannot query the cursor style :(

The only thing code-wise I'd suggest to improve is the magic number 12 of the resize area ;)
You added a comment for 4*3 but I stumbled over the for(i=0; i<12;i++) and asked myself what the 12 was about before I saw that comment. Maybe you can add a const expr/define/whatever with a more "descriptive" name?

t-wissmann:
Thanks for testing :) I've made the 12 a static constexpr.

blarz:
> Thanks for testing :) I've made the 12 a static constexpr.

Thanks! :)

blarz (approved):
Looks good to me and tested locally

PR #1332, opened 2021-06-03
Add an autostart object

### LINKED ISSUE #762: Implement an autostart object
This `autostart` object should be responsible for starting the autostart and it should also:

* allow to see and change the autostart path (`autostart.path`)
* keep track whether the actual autostart process is still running
* show this via an attribute (`autostart.running`, maybe `autostart.pid`)
* if `autostart.auto_lock` is set (which I'd set per default), the autostart object takes care of `lock`ing/`unlock`ing the monitor state depending on whether the autostart is running
* deny executing `reload` while  the autostart is still running

t-wissmann:
This adds an autostart object which allows modifying the custom
autostart path at run time and also helps querying information about the
last invoked autostart run (pid, status, running).

This closes #762.

### DIFF
--- a/src/autostart.cpp
+++ b/src/autostart.cpp
@@ -1,0 +1,89 @@
+#include "autostart.h"
+#include "globals.h"
+
+#include <unistd.h>
+
+using std::pair;
+using std::string;
+
+Autostart::Autostart(const string& autostartFromCmdLine, const string& globalAutostart)
+    : autostart_path_(this, "path", autostartFromCmdLine)
+    , global_autostart_path_(this, "global_path", globalAutostart)
+    , pid_(this, "pid", 0)
+    , running_(this, "running", false)
+    , last_status_(this, "last_status", 0)
+{
+    autostart_path_.setWritable();
+    autostart_path_.setDoc(
+                "Custom path to the user\'s autostart path. "
+                "If it is empty, then the autostart in $XDG_CONFIG_HOME "
+                "or $HOME is used.");
+    global_autostart_path_.setDoc("Path of the system-wide autostart, used as a fallback.");
+    pid_.setDoc(
+                "the process id of the last autostart invokation. "
+                "Even if the autostart is not running anymore, its pid "
+                "is still present here.");
+    running_.setDoc("whether the autostart process (with \'pid\') is still running.");
+    last_status_.setDoc(
+                "the exit status of the last autostart run. "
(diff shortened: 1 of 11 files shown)


t-wissmann:
Maybe `last_status` can be simplified to `status` or `exit_status`.

src/autostart.cpp
```diff
@@ -0,0 +1,90 @@
...
+    , last_status_(this, "last_status", 0)
+{
+    autostart_path_.setWritable();
+    autostart_path_.setDoc(
+                "Custom path to the user\'s autostart path. "
+                "If it is empty, then the autostart in $XDG_CONFIG_HOME "
+                "or $HOME is used.");
+    global_autostart_path_.setDoc("Path of the system-wide autostart, used as a fallback.");
+    pid_.setDoc(
+                "the process id of the last autostart invokation. "
```
blarz, line 24: invocation

src/main.cpp
```diff
@@ -464,8 +420,9 @@ int main(int argc, char* argv[]) {
     if (g.trueTransparency) {
         X->tryInitTransparency();
     }
-    // remove zombies on SIGCHLD
-    sigaction_signal(SIGCHLD, remove_zombies);
+    // ignore sigchld, but set it anyway such that we don't
+    // receive anything on on stopped children (SA_NOCLDSTOP)
```
blarz, line 424: on on

blarz (changes requested):
Some small typos, otherwise looks good!

Also: there are some autostart related tests in test_herbstluftwm.py, maybe they can be removed or update?

blarz:
> Some small typos, otherwise looks good!
> 
> Also: there are some autostart related tests in test_herbstluftwm.py, maybe they can be removed or update?

At least the `# TODO: verify the path as soon as we have an autostart object` comment seems like something that could be done (maybe in a different PR?)

t-wissmann:
I've fixed the typos and moved the test cases to `test_autostart.py`. The comment in `test_autostart_path` however requires another attribute that does not exist yet, so I think this should be done in separate PR.

blarz:
> I've fixed the typos and moved the test cases to `test_autostart.py`. The comment in `test_autostart_path` however requires another attribute that does not exist yet, so I think this should be done in separate PR.

Thanks for the fixes. And makes sense :)

blarz (approved):

PR #1342, opened 2021-06-23
Makefile: add phony 'check' target

Flowdalic:


Makefile
```diff
@@ -16,3 +16,7 @@ $(BUILDDIR):
 
 clean:
 	rm -r $(BUILDDIR)/
+
+.PHONY: check
+check: all
+	cd $(BUILDDIR); tox -c ..  -- -v --maxfail=1
```
t-wissmann, line 22: Can you add `-m 'not exclude_from_coverage'` to the pytest arguments? (i.e. after the `--`). The `exclude_from_coverage` take extra long. Before running tox, a couple of other commands are:
```
flake8
ci/check-using-std.sh
```

t-wissmann:
Also, you could mention `make check` in the `HACKING` file.

blarz:
Looks good to me! :)

t-wissmann:
Right, I didn't yet approve this PR because it doesn't yet contain `-n auto`: Should we add it here or is there a pytest option that we can put in `tox.ini` that always run pytest with `-n auto`?

The-Compiler:
There's `addopts`, see [Basic patterns and examples — pytest documentation](https://docs.pytest.org/en/6.2.x/example/simple.html#how-to-change-command-line-options-defaults)

t-wissmann (approved):
thanks! I think I need to tune it a bit further before I'm able to use it, but if it already helps you, I'm fine with it!

PR #1343, opened 2021-07-12
tests: Close stdout/stderr after subprocess.Popen

t-wissmann:
When using subprocess.Popen() and piping stdout or stderr, the pipe
needs to be closed on shutdown.

For the present change, I've grep'ed for all uses of subprocess.Popen()
and added the necessary close() invocations. In one occasion, I could
simply replace subprocess.Popen() / communicate() with subprocess.run()
which already takes care of this.

Unclosed file descriptors and similar issues are reported by passing
`-Wdefault` to pytest.

Reported-by: Florian Schmaus <flo@geekplace.eu>

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -267,6 +267,7 @@ class HlwmBridge(herbstluftwm.Herbstluftwm):
         for client_proc in self.client_procs:
             client_proc.wait(PROCESS_SHUTDOWN_TIME)
         self.hc_idle.wait(PROCESS_SHUTDOWN_TIME)
+        self.hc_idle.stdout.close()
 
     def bool(self, python_bool_var):
         """convert a boolean variable into hlwm's string representation"""
@@ -448,6 +449,8 @@ class HlwmProcess:
 
         # Make sure to read and echo all remaining output (esp. ASAN messages):
         self.read_and_echo_output(until_eof=True)
+        self.proc.stdout.close()
+        self.proc.stderr.close()
 
         if self.proc.returncode is None:
             # only wait the process if it hasn't been cleaned up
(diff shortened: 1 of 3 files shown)


blarz:
>  In one occasion, I could simply replace subprocess.Popen() by subprocess.communicate()

You replaced a Popen/communicate with run()

blarz:
> Unclosed file descriptors and similar issues are reported by passing -Wdefault to pytest.

Should we add this to our tox.ini or something?

t-wissmann:
> > In one occasion, I could simply replace subprocess.Popen() by subprocess.communicate()
> 
> You replaced a Popen/communicate with run()

Thanks! It took me bit until I saw where my mistake in the phrasing was :-)

> > Unclosed file descriptors and similar issues are reported by passing -Wdefault to pytest.
> 
> Should we add this to our tox.ini or something?

Hm not sure, I don't want to spend too much time on it, because it's only the test suite, not the actual software. Other people probably disagree...

blarz:
> > > In one occasion, I could simply replace subprocess.Popen() by subprocess.communicate()
> > 
> > 
> > You replaced a Popen/communicate with run()
> 
> Thanks! It took me bit until I saw where my mistake in the phrasing was :-)

Thanks for fixing it!

> > > Unclosed file descriptors and similar issues are reported by passing -Wdefault to pytest.
> > 
> > 
> > Should we add this to our tox.ini or something?
> 
> Hm not sure, I don't want to spend too much time on it, because it's only the test suite, not the actual software. Other people probably disagree...

Alright, just wanted to ask if it would be useful.

The change as such looks good btw!

blarz (approved):

t-wissmann:
Thanks!

Issue #1340, opened 2021-06-22
Link user contributions on website


t-wissmann:
There are multiple scripts in users' repositories wich deserve being linked from the herbstluftwm homepage. I propose we collect them here and then put a list on the homepage.

https://github.com/AnomalousBit/tag_spaces

juacq97:
Something like awesome repos but in the webpage. Maybe the ability to send PR to the webpage if someone wants to add their own scripts?

t-wissmann:
The wegpage is part of hlwm's git repository, so sending PRs for the wegpage is already possible :-)

PR #1344, opened 2021-07-16
Add an "Users contributions" section on the webpage

juacq97:
This adds a `contrib.txt` file to the `www` directory and solves #1340

I decided to add the new section as a new tab. I included instructions on how to add new contributions and I added some contributions I found at r/herbstluftwm. I have to make it clear: I didn't ask  permission from the authors to include their contributions on the webpage, I just assumed that I could do it since they had already shared it on reddit and they're using free licences. Of course, each contribution has the username of whoever shared it and (if apply) a link to their repo.

I also changed the `compose.py` files so the new section can be compiled by asciidoc.

### DIFF
--- a/www/Makefile
+++ b/www/Makefile
@@ -50,6 +50,8 @@ faq-toc.html: faq-content.html ./gentoc.sh
 
 faq.html: faq-toc.html
 
+contrib.html: contrib.txt
+
 favicon.ico: ../share/herbstluftwm.svg
 	convert -density 256x256 -background transparent $< \
         -define icon:auto-resize -colors 256 $@
--- a/www/Makefile
+++ b/www/Makefile
@@ -50,8 +50,6 @@ faq-toc.html: faq-content.html ./gentoc.sh
 
 faq.html: faq-toc.html
 
-contrib.html: contrib.txt
-
 favicon.ico: ../share/herbstluftwm.svg
 	convert -density 256x256 -background transparent $< \
         -define icon:auto-resize -colors 256 $@
--- a/www/compose.py
+++ b/www/compose.py
@@ -23,7 +23,7 @@ tabs = OrderedDict([
         ("download", "Download"),
     ])),
     ("Users contributions", OrderedDict([
-		("contrib", "Users contributions"),
+        ("contrib", "Users contributions"),
     ])),
     # ("Wiki", "http://wiki.herbstluftwm.org"),
 ])
--- a/www/compose.py
+++ b/www/compose.py
@@ -22,6 +22,9 @@ tabs = OrderedDict([
     ("Download", OrderedDict([
         ("download", "Download"),
     ])),
+    ("Users contributions", OrderedDict([
+		("contrib", "Users contributions"),
+    ])),
     # ("Wiki", "http://wiki.herbstluftwm.org"),
 ])
 
--- a/www/contrib.txt
+++ b/www/contrib.txt
@@ -33,9 +33,7 @@ Get it from https://github.com/pjones/hlwmrc/blob/9f38a3023660c8c1b617be7689823f
 Layout save and recall fuction by u/ToPow1
 ------------------------------------------
 
-This function allow you to save your current layout and give it a name. You can save as many layouts as you like using the +zentity+ packages.
-
-The saved layouts can then called back on any TAG you like.
+This function allow you to save your current layout and give it a name. You can save as many layouts as you like using the +zentity+ packages. The saved layouts can then called back on any TAG you like.
 
 This function is done with two short bash scripts. The first to save a layout and name it, the second to list the saved layouts and print it to the current TAG.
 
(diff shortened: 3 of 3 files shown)


www/Makefile
```diff
@@ -50,6 +50,8 @@ faq-toc.html: faq-content.html ./gentoc.sh
 
 faq.html: faq-toc.html
 
+contrib.html: contrib.txt
+
```
t-wissmann, line 54: This should be taken care of automatically by `TXT = $(shell git ls-files '*.txt')` in line 8, so actually you don't need to (and thus should not) modify the Makefile. (Probably it didn't work for you originally because `contrib.txt` wasn't tracked by git)

www/compose.py
```diff
@@ -22,6 +22,9 @@
     ("Download", OrderedDict([
         ("download", "Download"),
     ])),
+    ("Users contributions", OrderedDict([
+		("contrib", "Users contributions"),
+    ])),
     # ("Wiki", "http://wiki.herbstluftwm.org"),
```
t-wissmann, line 28: Can you please indent with spaces instead of tabs? then, flake8 in the CI will also be happy:
```
./www/compose.py:26:1: W191 indentation contains tabs
./www/compose.py:26:1: E101 indentation contains mixed spaces and tabs
./www/compose.py:26:3: E126 continuation line over-indented for hanging indent
ERROR: InvocationError for command /__w/herbstluftwm/herbstluftwm/.tox/flake8/bin/flake8 . (exited with code 1)
```

www/contrib.txt
```diff
@@ -0,0 +1,73 @@
+Users contributions
+===================
+
+How to contribute?
+------------------
+
+* Clone the repo with +git clone https://github.com/herbstlufwm/herbstluftwm+
+* Open the +herbstluftwm/www/contrib.txt+ file.
+* Edit the file and add your contribution! It uses the https://powerman.name/doc/asciidoc[asciidoc format].
+* Send your pull request! If you want to test the site run the +make+ command (it needs asciidoc installed). Just remember to run +make clean+ before sending the pull request.
```
t-wissmann, line 10: Actually, I don't see why `make clean` is necessary before sending the pull request :-)

www/contrib.txt
```diff
@@ -0,0 +1,73 @@
...
+
+Automatically disable focus_follows_mouse for floating windows by u/pmade
+-------------------------------------------------------------------------
+
+This script automatically turns off +focus_follows_mouse+ for floating windows, and turns it back on for tiled windows. Now you can use +focus_follows_mouse+ like normal but when a pesky dialog opens I have an extra step of clicking to lose focus.
+
+Get it from https://github.com/pjones/hlwmrc/blob/9f38a3023660c8c1b617be7689823f09115ecf75/bin/hlwm-ffm-skip-floats.sh[github!]
+
+Layout save and recall fuction by u/ToPow1
+------------------------------------------
```
t-wissmann, line 34: I like the first two sections, because they consist of a short description and then provide a link. The third section is a bit too long for my taste. Can you link some repository or reddit post here?

I want like to avoid code here, because then I would become the sort of maintainer.

t-wissmann (changes requested):
Thanks for your work! I've made a few comments. The biggest being the point that I don't want the contrib page to become a wiki but more a list of references.

www/Makefile
juacq97: You're right! I remove that line from there.

www/contrib.txt
juacq97: In the repo, there's only text files and no html file. In order to mantain it like that I added that part, but is not necessary

juacq97:
> Thanks for your work! I've made a few comments. The biggest being the point that I don't want the contrib page to become a wiki but more a list of references.

Regarding this, I just copied the description provided by the original author on their repos or reddit post. An alternative could be something like this:

* tag_spaces by AnomalousBit: groups of tags, similar to KDE Plasma's activities [link to the repo]
* Disble focus_follows_mouse for floating windows by u/pmade: [link to the repo]

t-wissmann:
> > Thanks for your work! I've made a few comments. The biggest being the point that I don't want the contrib page to become a wiki but more a list of references.
> 
> Regarding this, I just copied the description provided by the original author on their repos or reddit post. An alternative could be something like this:
> 
> * tag_spaces by AnomalousBit: groups of tags, similar to KDE Plasma's activities [link to the repo]
> * Disble focus_follows_mouse for floating windows by u/pmade: [link to the repo]
> 
> I like the idea to add the username of the author, specially if the original post was made on a reddit post or the script is saved as a gist

I also like this compact format and especially that you mention the usernames for giving them credit. I also suggest to remove the longer description text that you have copied over. A user will find this long description anyway when clicking on the link.

t-wissmann:
By the way, I have also updated this PR's title and description slightly, because they will be the commit message of the commit created by this PR.

www/contrib.txt
```diff
@@ -0,0 +1,14 @@
+Users contributions
+===================
+
+* tag_spaces by AnomalousBit: groups of tags similar to KDE Plasma activities. Get it from https://github.com/AnomalousBit/tag_spaces[github!]
+* Automatically disable focus_follows_mouse for floating windows by u/pmade. Get it from https://github.com/pjones/hlwmrc/blob/9f38a3023660c8c1b617be7689823f09115ecf75/bin/hlwm-ffm-skip-floats.sh[github!]
+* Layout save and recall fuction by u/ToPow1. Get it from https://www.reddit.com/r/herbstluftwm/comments/o13vny/layout_save_and_recall_function_for_herbstluftwm[this post on reddit!]
```
blarz, line 6: Small typo: "fuction"

blarz: Thanks!

t-wissmann (approved):
Thanks!

PR #1348, opened 2021-07-28
Resize frames in both dimensions

t-wissmann:
This implements a feature which was already indicated by the cursor shape:
resizing frames in both x and y direction. For this, it was necessary to
entirely change the drag handler for resizing frames.

As a little bonus, the special cursor shape persists during the drag.

The new test case essentially checks that resizing in both dimensions
works as expected. The existing testcase for 1-dimensional resize
could be simplified a lot using new attributes and the convenience
functions in the python bindings.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -249,6 +249,29 @@ ResizeAction Decoration::positionTriggersResize(Point2D p)
     return act;
 }
 
+/**
+ * @brief Find the most apropriate ResizeAction given the current
+ * cursor position. This is a very fuzzy version of positionTriggersResize()
+ * @param the cursor position
+ * @return
+ */
+ResizeAction Decoration::resizeFromRoughCursorPosition(Point2D cursor)
+{
+    if (!last_scheme) {
+        // this should never happen, so we just randomly pick:
+        // never resize if there is no decoration scheme
+        return ResizeAction();
+    }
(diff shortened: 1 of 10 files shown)


src/decoration.cpp
```diff
@@ -249,6 +249,29 @@ ResizeAction Decoration::positionTriggersResize(Point2D p)
     return act;
 }
 
+/**
+ * @brief Find the most apropriate ResizeAction given the current
+ * cursor position. This is a very fuzzy version of positionTriggersResize()
+ * @param the cursor position
+ * @return
```
blarz, line 256: Add a description of the return value?

blarz (changes requested):
Just two minor things, other than that it looks good me!

t-wissmann:
Thanks! I've fixed the two instances.

src/decoration.cpp
blarz: Thanks!

src/decoration.cpp
```diff
@@ -249,6 +249,29 @@ ResizeAction Decoration::positionTriggersResize(Point2D p)
     return act;
 }
 
+/**
+ * @brief Find the most apropriate ResizeAction given the current
```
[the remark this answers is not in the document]
blarz: Thanks!

blarz (approved):

PR #1350, opened 2021-08-07
FAQ: update frame layout for existing tags

t-wissmann:
The following snippet updates the layout algorithm of all root frames to
the current default_frame_layout setting.


t-wissmann:
The CI failed, but only because of the shutdown timeout of xvfb, tackled in #1351.

blarz (approved):

Issue #1347, opened 2021-07-22
FreeSync not working with managed windows


mserajnik:
I'm experiencing the issue that FreeSync isn't working under herbstluftwm (using no compositor) using a Radeon RX 6900 XT GPU, while it works just fine under KDE Plasma. See my [initial issue](https://gitlab.freedesktop.org/drm/amd/-/issues/1659) in the AMD DRM project that I made before finding out it's herbstluftwm-specific for more details.

After figuring that out, I did some more testing and noticed that it does work with unmanaged windows, e.g.:

```bash
hc rule class='love' manage=off
```

With this rule, the [VRRTest](https://github.com/Nixola/VRRTest) window is unmanaged, causing FreeSync to work.
However, that is not really ideal since now the window apparently doesn't get focus (and therefore doesn't get keyboard inputs) as well as it also not being possible to access the rest of the desktop anymore (I'm effectively stuck in that one fullscreen window); I would still need to be able to switch workspaces and use the rest of the desktop as usual while running a game in fullscreen with FreeSync.

Given that KDE Plasma is able to have managed fullscreen windows with FreeSync working (at least that's what I assume since they integrate seamlessly), I am not sure if this is a bug or limitation in herbstluftwm, or if I am simply not aware how to configure this properly.

On a sidenote: when I was still using a GeForce GTX 1080 Ti about half a year ago, G-SYNC (or rather NVIDIA's implementation of FreeSync, since it's just a "G-SYNC compatible" FreeSync monitor) was working just fine under herbstluftwm, but maybe that was due to a implementation difference compared to AMD's FreeSync.

t-wissmann:
Hi, thanks for the report and sorry for the trouble! This will be hard to reproduce because I don't have any radeon GPUs here. I don't think that the fullscreen state is a real issue here, because the fullscreen isn't anything special (it is essentially maximized, without decoration, and hiding panels).

The main difference between normal windows and unmanaged windows in this regard is that unmanaged windows are not reparented. Window reparenting means that the actual client window is swallowed by another window which shows the window decorations.

My first guess is that the issue here is the [EWMH `_NET_WM_SYNC_REQUEST` protocol](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46841634376672) which is not implemented by hlwm.

mserajnik:
@t-wissmann 

Thank you for the quick response!

[...]

PR #1374, opened 2021-10-17
Attribute to disable window decorations

t-wissmann:
This adds a client attribute 'decorated' that adjusts whether a client
has a window decoration frame. Changing the option reparents the window
either to the root or to the decoration window.

This should make it easier to work with shaped windows and to solve
issues with freesync (#1347).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,7 @@ Current git version
     'HLWM_FLOATING_WINDOW' and 'HLWM_TILING_WINDOW'
   * The 'spawn' command now prints an error message on exec failure
   * New read-only client attribute 'decoration_geometry'.
+  * New attribute 'decorated' to disable window decorations
   * The cursor shape now indicates resize options.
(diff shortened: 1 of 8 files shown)


t-wissmann:
This is still in WIP state: there are no test cases yet and maybe the attribute should be split into multiple attributes like `decorated_floating`/`decorated_tiling`

tiosgz:
> maybe the attribute should be split into multiple attributes like `decorated_floating`/`decorated_tiling`

Just an idea: how much additional work would it require to make it work like most attributes in the `theme.` subtree so that setting `decorated` would change both `decorated.tiling` and `decorated.floating`?

t-wissmann:
Now, it's ready :-)

blarz:
> Now, it's ready :-)

I'll have a look on the weekend!

blarz (approved):
Looks good to me and works nicely :)

Thanks!

t-wissmann:
Thanks!

Issue #33, opened 2014-10-09
Focus stealing bug


The-Compiler:
Sometimes, dwb or mpv steal the focus, and switching keyboard focus does not work (immediately focuses back to dwb/mpv again).

Maybe related (?): Some clients (steam) create windows which have focus without having the focus borde

fdietze:
I can reproduce this using entr (http://entrproject.org/) and their reload-browser script with Chromium and Firefox.

The-Compiler:
I just saw the same thing happening with two urxvt instances, so it doesn't seem to be connected to those applications.

ypnos:
I found an easy method to reproduce X/hlwm focus mismatch. Use `xdo-tool` to hand focus to a specific window. Hlwm fails to reflect the change and redraw borders.

ypnos:
I recently experienced another focus problem with a Java application. Lets say I have a layout A | B. In B the java application window resides. In A a terminal window resides.
Now I have the terminal window focused and move it over to B.

While the terminal is drawn as expected, all keyboard input actually goes to the hidden java application. I have to cycle between them to get focus back to the terminal.

t-wissmann:
I think this now has been fixed (at least in the instances that I could reproduce via xdotool and emacsclient). If it still happens with the current git version or with the next release (0.9.4 or higher), please re-open.

PR #1386, opened 2021-11-27
React to focus stealing via XSetInputFocus

t-wissmann:
Usually, clients should request a focus change via an EWMH request,
which the WM can then respect or ignore. Some clients however just
change the input focus themselves (i.e. they change the window where
keyboard events are sent to), e.g. via XSetInputFocus().
Examples are `xdotool windowfocus` or `emacsclient -n $filename`.

The only way to handle these input focus changes is to update hlwm's
state to match the new focus. I am not entirely confident in the code
handling it, but it seems to update the clients.focus object accordingly
(see the new test case) and it does not trigger any endless loop (which
would happen without the XCheckMaskEvent()).

I assume that this fixes the issue #33.

### DIFF
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -622,3 +622,29 @@ const char* XConnection::requestCodeToString(int requestCode)
     }
     return nullptr;
 }
+
+#define DetailCodeAndString(C)  { C, #C }
+/**
+ * @brief print the name of the 'detail' of a XFocusedChangedEvent
+ * @param the 'detail' of a XFocusedChangedEvent
+ * @return
+ */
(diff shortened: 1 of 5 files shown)


blarz (approved):
Looks ok to me, let's see if it breaks anything :D

t-wissmann:
I just had a horrible loop where the focus quickly togged between two windows (but I'm not able to reproduce it). I think this PR needs to be revised.

blarz:
Oh noes, ok.. Revised or reverted? ;)

t-wissmann:
Hopefully only 'revised', but maybe I should first check the icccm specification how focusing should work correctly.

Issue #1097, opened 2020-12-29
Ideas for window title appearance


The-Compiler:
Some rough ideas about how to show window titles:

- Show the text centered rather than left-aligned (cf. [Reddit comment](https://www.reddit.com/r/herbstluftwm/comments/km0rkz/initial_window_title_support_just_got_merged/ghdwc87?utm_source=share&utm_medium=web2x&context=3)
- Only draw the background color behind the title, if this is possible - i.e. so it shows like a "tab" rather than a wide colored bar, and is transparent otherwise
- Perhaps have some indicator for max layout, and number of windows or so? Or combined with the above idea, show multiple windows as tabs?

dnnr:
Centering the window titles might be problematic with terminals that quickly change their titles. Lots of jumping around compared to left-aligned. This should be configurable, imho.

The-Compiler:
Yeah, I don't think it should be the default - but might be useful as an option.

t-wissmann:
I've been brainstorming a lot on this. For me, the optimal solution would be a format string which allows inserting arbitrary information (from the object/attributes).
And when it comes to alignment, I'm wondering wheter something like "centered alignment" should be one of the formating directives (as it is for lemonbar) or whether ther should be three such title_formats: for the left, the center, and the right part of the decoration.

apm256:
> I'm wondering wheter something like "centered alignment" should be one of the formating directives (as it is for lemonbar) 

I already use the panel (via `barpyrus`) to display information like: focused window's title, the number of clients stacked in the current frame when using the `max` layout, the current layout type, the "maximized.sh" status, the presence of minimized windows in the current tag, and so on.
So I was wondering: could we see window title as a frame-dedicated panel?
(Sorry if this comment is somewhat naive!)

t-wissmann:
> > I'm wondering whether something like "centered alignment" should be one of the formating directives (as it is for lemonbar)
> 
> I already use the panel (via `barpyrus`) to display information like: focused window's title, the number of clients stacked in the current frame when using the `max` layout, the current layout type, the "maximized.sh" status, the presence of minimized windows in the current tag, and so on.
> So I was wondering: could we see window title as a frame-dedicated panel?
> (Sorry if this comment is somewhat naive!)

In principle, it's possible because the frames already exist as x11-windows. But like my dream of a format string for client titles, it's quite an effort to implement.

The-Compiler:
[...]

PR #1390, opened 2021-12-04
Make title alignment configurable

t-wissmann:
This adds a new attribute title_align that configures the horizontal
alignment of the window title within the title bar, respectively tab.

This was requested in #1097.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,7 @@ Current git version
   * Tabbed window titles in the 'max' layout algorithm (controllable via the
     'tabbed_max' setting)
   * New autostart object with attributes 'path', 'running', 'pid', 'last_status'
+  * New decoration attribute 'title_align'
   * New client attribute 'floating_effectively' and associated X11 properties
     'HLWM_FLOATING_WINDOW' and 'HLWM_TILING_WINDOW'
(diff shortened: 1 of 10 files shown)


blarz:
Left aligned text has some padding, right aligned doesn't and it looks a bit a bit odd IMHO.

Should the same padding be applied to the right, too?

t-wissmann:
> Left aligned text has some padding, right aligned doesn't and it looks a bit a bit odd IMHO.
> 
> Should the same padding be applied to the right, too?

Well spotted, Thanks! I've applied the pad also to the 'width', but of course I have to apply it twice. I've fixed it :-)

blarz:
> > Left aligned text has some padding, right aligned doesn't and it looks a bit a bit odd IMHO.
> > Should the same padding be applied to the right, too?
> 
> Well spotted, Thanks! I've applied the pad also to the 'width', but of course I have to apply it twice. I've fixed it :-)

Great, thank you, right align looks good now :)

blarz (approved):
Very cool, thank you!

t-wissmann:
There is an issue with mergify, I am taking a look at it.

PR #1393, opened 2021-12-05
Draw title of tabbed windows before horizontal border

### LINKED ISSUE #1388: Titles of non-focused windows in max layout bleeds into focused window's border
See the `p` in `desktop` in this screenshot:

Or the much more noticeable `g`s in this screenshot:

Relevant settings:
```
$ herbstclient attr theme.
    inner_color/inner_width
          ╻        outer_color/outer_width
          │                  ╻
          │                  │
    ┌────╴│╶─────────────────┷─────┐ ⎫ border_width
    │     │      color             │ ⎬ + title_height
    │  ┌──┷─────────────────────┐  │ ⎭ + padding_top
    │  │====================....│  │
    │  │== window content ==....│  │
    │  │====================..╾──────── background_color
    │  │........................│  │
    │  └────────────────────────┘  │ ⎱ border_width +
    └──────────────────────────────┘ ⎰ padding_bottom

Setting an attribute of the theme object just propagates the value to the respective attribute of the +tiling+ and the +floating+ object.

7 children:
  active.
  floating.
  fullscreen.
  minimal.
  normal.
  tiling.
  urgent.
16 attributes:
 .---- type
 | .-- writable
 | | .-- hookable
 V V V
 c w h background_color = #1a1b1b
 u w h border_width = 3
 c w h color = #000000
 c w h inner_color = #000000
 u w h inner_width = 1
 c w h outer_color = #000000
 u w h outer_width = 1
 u w h padding_bottom = 0
 u w h padding_left = 0
 u w h padding_right = 0
 u w h padding_top = 0
 s w - reset = "Writing this resets all attributes to a default value"
 b w h tight_decoration = false
 c w h title_color = #000000
 f w h title_font = Source Code Pro:size=7:semibold
 u w h title_height = 12
$ herbstclient attr theme.active.color
#a89984
$ herbstclient attr theme.normal.color
#504945
```

A workaround is setting `padding_top` to some fitting number in relation to `(border|inner|outer)_width`: `2` in this case. Though I'd rather have some (slightly) ugly titles than unnecessary imbalanced padding.

gravndal:
Draw the tab title earlier, such that letters reaching below the baseline get
covered by the tab border if there is not enough space.

This fixes #1388.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -617,6 +617,9 @@ void Decoration::redrawPixmap() {
                 }
                 XSetForeground(display, gc, get_client_color(tabScheme.outer_color));
                 XFillRectangles(display, pix, gc, &borderRects.front(), borderRects.size());
+                drawText(pix, gc, tabScheme.title_font->data(), tabScheme.title_color(),
+                         tabGeo.tl() + Point2D { tabPadLeft, (int)s.title_height()},
+                         tabClient->title_(), titleWidth - tabPadLeft);
                 if (client_ != tabClient) {
                     // horizontal border connecting the focused tab with the outer border
                     Point2D westEnd = tabGeo.bl();
@@ -641,9 +644,6 @@ void Decoration::redrawPixmap() {
                                    fillWidth, remainingBorderColorHeight
                                    );
                 }
-                drawText(pix, gc, tabScheme.title_font->data(), tabScheme.title_color(),
-                         tabGeo.tl() + Point2D { tabPadLeft, (int)s.title_height()},
-                         tabClient->title_(), titleWidth - tabPadLeft);
                 tabIndex++;
             }
         }


src/decoration.cpp
```diff
@@ -615,6 +615,9 @@ void Decoration::redrawPixmap() {
                     );
                     titleWidth -= tabScheme.outer_width;
                 }
+                drawText(pix, gc, tabScheme.title_font->data(), tabScheme.title_color(),
+                         tabGeo.tl() + Point2D { tabPadLeft, (int)s.title_height()},
+                         tabClient->title_(), titleWidth - tabPadLeft);
                 XSetForeground(display, gc, get_client_color(tabScheme.outer_color));
                 XFillRectangles(display, pix, gc, &borderRects.front(), borderRects.size());
```
gravndal, line 622: It might be more appropriate to put the drawText call after these two lines?

t-wissmann:
Thanks, makes perfect sense! I've moved your screenshots out of the commit message and extended your commit message a bit.

Screenshots:

src/decoration.cpp
t-wissmann: Yes, the draw text should be after that, otherwise the title might be covered for plain X11 fonts; hopefully the CI will discover this issue :-). (There is something special about the drawing of Xft fonts, and this is why the issue is probably not visible in your screenshot)

gravndal: Updated the pr.

t-wissmann:
I've just realized that I misread your question about the order with `XFillRectangles`. I thought that it fills the background, but it actually just draws the 'outer border'. So I think that the order does not really matter and it can stay as it is :-)

t-wissmann (approved):
Works well, thanks a lot!

PR #1394, opened 2021-12-05
Extend foreach command by --unique and --recursive

t-wissmann:
This adds two flags to foreach: --unique removes duplicates (or more
precisely, aliases) of objects and --recursive recurses into subobjects.
To ensure termination --recursive implicitly also uses --unique. The
recursion is done in a breadth-first way such that deeply nested objects
are visited later.

The main application for --unique is to avoid the duplicate
clients.focus and the main application of --recursive is to traverse all
frames in the tree layout.

### DIFF
--- a/tests/test_meta_commands.py
+++ b/tests/test_meta_commands.py
@@ -746,12 +746,108 @@ def test_foreach_object_completion(hlwm):
 
 
 def test_foreach_flags_completion(hlwm):
-    flag_name = '--filter-name='
-    assert flag_name in hlwm.complete(['foreach'], partial=True)
-    assert flag_name in hlwm.complete(['foreach', 'X'], partial=True)
-    assert flag_name in hlwm.complete(['foreach', 'X', 'tags.'], partial=True)
-    # do not complete flags when the first command token appeared
-    assert flag_name not in hlwm.complete(['foreach', 'X', 'tags.', 'echo'], partial=True)
+    for flag_name in ['--filter-name=', '--unique ', '--recursive ']:
+        assert flag_name in hlwm.complete(['foreach'], partial=True)
+        assert flag_name in hlwm.complete(['foreach', 'X'], partial=True)
+        assert flag_name in hlwm.complete(['foreach', 'X', 'tags.'], partial=True)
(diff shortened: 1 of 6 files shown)


tests/test_meta_commands.py
```diff
@@ -746,12 +746,108 @@ def test_foreach_object_completion(hlwm):
...
+    hlwm.call('split bottom')
+    hlwm.call('split left')
+
+    def object_paths():
+        return hlwm.call(['foreach', 'X', '', '--recursive', 'echo', 'X']) \
+                   .stdout.splitlines()
+
+    without_client = object_paths()
+
+    # and put the a client in the frame leaf:
```
blarz, line 782: typo the a?

tests/test_meta_commands.py
```diff
@@ -746,12 +746,108 @@ def test_foreach_object_completion(hlwm):
...
+
+    found_in_without_client = False
+    for tag_id in ['0', 'focus']:
+        frame_object = f'tags.{tag_id}.tiling.focused_frame'
+        assert frame_object not in with_client
+        if frame_object in without_client:
+            found_in_without_client = True
+
+    assert found_in_without_client, \
+        "The frame must be listed under tags. before client creation"
```
blarz, line 799: replace . with ,?

blarz (commented):
Minor issues, otherwise looks good to me!

tests/test_meta_commands.py
t-wissmann: I've removed 'a'

tests/test_meta_commands.py
t-wissmann: actually, I meant the object `tags.`, I've simply replaced it with `'tags'`

t-wissmann:
Thanks for checking, I've fixed both! :-)

tests/test_meta_commands.py
blarz: Aaah, thanks for clearing that up! :)

blarz:
> Thanks for checking, I've fixed both! :-)

Thanks :)

blarz (approved):

PR #1396, opened 2021-12-05
Make fixed font visible in man page

t-wissmann:
This patches the man page creation such that text formatted in fixed
font is recognizable in the man page. This is makes examples stand out
more in the man page.

On the technical side, the cmake file now does the two steps which are
normally done by a2x: first creating a docbook xml, and then
transforming the xml to a man page. By doing these steps manually, we
can now patch the intermediate xml file to our needs.

Maybe, this also helps us to see more error messages which were formerly
buried in a2x. In fact, the present change helped to discover two
broken links within the herbstluftwm man page, which I've fixed along
the way. Strictly speaking, we don't need the a2x binary anymore, but
since it comes with asciidoc anyway, I've left it as a dependency.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -249,6 +249,7 @@ jobs:
           cat > CMakeLists.txt <<EOF
           project(Herbstluftwm)
           file(STRINGS VERSION VERSION) # read file 'VERSION'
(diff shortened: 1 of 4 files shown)


t-wissmann:
Old version is on the left, new one on the right:
image

blarz (approved):
Tested locally, works fine :)

PR #1397, opened 2021-12-06
Remove remaining os.path usage in setup.py

ju-sh:
Similar to what was done in the [sampleproject repo](https://github.com/pypa/sampleproject/blob/main/setup.py).

Also capitalized starting of description.


t-wissmann (approved):
Thanks! looks good to me.

PR #1398, opened 2021-12-06
Update mergify configuration

t-wissmann:
This now uses the merge queue of mergify. Unfortunately,
the mergify documtation also will remove the `commit_message` option
from the queue action. It is unclear to me how we can keep the current
behaviour of squashing the entire PR into a single commit with the
commit message coming from the PR title and description.

The mergify config checker tells:

> The configuration uses the deprecated commit_message mode of the queue
> action. A brownout is planned for the whole March 21th, 2022 day. This
> option will be removed on April 25th, 2022. For more information:
> https://docs.mergify.com/actions/queue/

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,24 +1,33 @@
+queue_rules:
+  - name: default
+    conditions:
+      # Conditions to get out of the queue (= merged)
+      - check-success=Build and test on current ubuntu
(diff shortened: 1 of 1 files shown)


t-wissmann:
image

Strangely enough, `method: squash` is not deprecated:
image

But `commit_message` is:
image

PR #1399, opened 2021-12-10
Add attributes 'title_when' and 'title_depth'

t-wissmann:
This makes it configurable when to show the title bar. So far, there are
four options that essentially depend on the number of tabs.
So far, I've used padding_top to add space for characters in the title
that reach below the baseline. Since the space below the baseline should
be present only when the title is shown, I've added a new setting
title_depth.

This new feature was requested multiple times on the IRC and in the
discussion of #1097 and #1388.

### DIFF
--- a/tests/test_theme.py
+++ b/tests/test_theme.py
@@ -186,7 +186,7 @@ def test_title_when_behaviour(hlwm):
 def test_title_when_for_absence_of_tabs(hlwm, floating):
     """
     for floating clients or frames with an algorithm other than 'max',
-    there should no be tabs at all and thus the titlebar is
+    there should be no tabs at all and thus the titlebar is
     only shown if 'title_when' is 'always'.
     """
     hlwm.attr.theme.title_height = 10
--- a/tests/test_theme.py
+++ b/tests/test_theme.py
@@ -1,6 +1,8 @@
 import pytest
 from herbstluftwm.types import Point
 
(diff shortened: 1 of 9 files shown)


t-wissmann:
I can fix the failing test case by extending `showTitle()` with:
```C++
bool DecorationScheme::showTitle(size_t tabCount) const
{
    if (title_height() == 0) {
        return false;
    }
    switch ...
}
```
but first I'd like to understand where X11 errors come from that make hlwm quit.

tests/test_theme.py
```diff
@@ -134,3 +136,79 @@ def test_decoration_geometry_vs_x11_data(hlwm, x11, floating):
...
+                assert title_size == expected_title_size
+            else:
+                assert title_size == 0
+
+
+@pytest.mark.parametrize("floating", [True, False])
+def test_title_when_for_absence_of_tabs(hlwm, floating):
+    """
+    for floating clients or frames with an algorithm other than 'max',
+    there should no be tabs at all and thus the titlebar is
```
blarz, line 189: be no

blarz (changes requested):
Just a small grammar error, otherwise looks very good!

tests/test_theme.py
blarz: or not be ;)

t-wissmann: oups. yes, "there should be no tabs at all" :-)

t-wissmann:
Thanks :)

blarz (approved):
Very cool, thank you! :)

PR #1400, opened 2021-12-10
Add settings.ellipsis

### LINKED ISSUE #1395: Appending ellipsis or user defined string when truncating titles
As discussed on IRC, an option for users to either choose from unicode or ascii ellipsis, or define an arbitrary string to append to the end of a truncated title would be fantastic.

t-wissmann:
As requested in #1395, this adds an ellipsis, which is appended to the
window title if its too long for the available space. This requires that
we erase characters in the middle of a utf-8 string, and so we need to
take care of characters that amount to multiple bytes in the string.

This closes #1395.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,7 @@ Current git version
   * New read-only client attribute 'decoration_geometry'.
   * New attribute 'decorated' to disable window decorations
   * The cursor shape now indicates resize options.
+  * New setting 'ellipsis'
   * Frames can be simultaneously resized in x and y direction with the mouse.
   * Bug fixes:
     - Update floating geometry if a client's size hints change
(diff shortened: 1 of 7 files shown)


blarz (approved):
Nice!

blarz:
Ah, just noticed: should this new settings also be added to doc/herbstluftwm.txt?

t-wissmann:
Thanks for merging :-)

On the long run, I'd like to move the documentation of all settings to the auto-generated part for attributes in the man page. So I hope that it is well-documented enough if the new settings are only documented there:
image
Maybe I should move the documentation of the other settings there as well.

blarz:
Oh yes I totally forgot about the auto-generated docs. That's perfect then :)

PR #1401, opened 2021-12-10
Let active and inactive tabs have same text width

t-wissmann:
If the title was centered or right-aligned, the title sometimes moved
slightly. This was caused by the text area having slightly different
widths for active and inactive tabs. The present change fixes this and
adds a test case for it.

Unfortunately, the test case is quite time consuming (8 seconds here),
because it takes 3 * 2 = 6 screenshots, but I think it's worth the
effort.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -568,7 +568,7 @@ void Decoration::redrawPixmap() {
                     tabButton.tabClient_ = tabClient;
                     buttons_.push_back(tabButton);
                 }
-                int titleWidth = tabGeo.width;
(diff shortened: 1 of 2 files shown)


blarz (approved):
Wow, if you wouldn't have mentioned it, I'd never noticed.

Thanks for fixing this!

t-wissmann:
> Wow, if you wouldn't have mentioned it, I'd never noticed.
> 
> Thanks for fixing this!

Thanks :) Sometimes this issue is just hidden (e.g. it is not visible at all for align=left, and even for align=center it sometimes disappears because of integer division)

PR #1402, opened 2021-12-11
Simplify urgency handling by using signals

t-wissmann:
This simplifies the handling of urgency handling and as a bonus, it
makes the urgency attribute writable.

The main idea behind this change is to separate the logic from the
communication with X11. One function (readWmHints) reads the urgency
hint from X11 to the attributes, and a signal propagates changes of the
attribute's value back to X11. The entire hlwm 'logic' only operates on
the urgency attribute (and so we can also grant the user write access to
it). This is similar to the focus change handling in #1403.

Also, this avoids a few X11 errors if a new client is already urgent.
Before this change, an urgent client triggered an early relayout, which
even led to the CI failure in #1399.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -84,6 +84,7 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
     sizehints_floating_.setWritable();
     sizehints_tiling_.setWritable();
     minimized_.setWritable();
+    urgent_.setWritable();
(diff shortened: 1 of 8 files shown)


blarz (approved):
Fixed a typo and now looks good to me :)

t-wissmann:
Thanks :)

PR #1403, opened 2021-12-12
Move input focus handling to XMainLoop

t-wissmann:
In order to make the Client class a bit more independent of X11, we move
the input focus control to XMainLoop, which is invoked by the signal
whenever clients.focus is changed -- very similar to what is done for
the urgent attribute in #1402.


t-wissmann:
This will conflict with the diff in #1402, but it should be easily resolvable.

blarz (approved):

PR #1405, opened 2021-12-12
Move settings doc to in-source documentation

t-wissmann:
This removes the section on settings from the man page and replaces it
with the automatically generated documentation on the attributes of the
settings object.

Unfortunately, the asciidoc syntax is a bit of a mess, and so it is very
complicated to correctly put longer doc strings containing code blocks
together. And in addition to that, escaping special characters is also a
mess, and so I'm now using html entities in the doc string of the
tree_style setting (for the time being, the help command just literally
prints the html entities, but the html and man page documentation
resolves them correctly).


t-wissmann:
The messy asciidoc syntax really consumes too much time. I wish there was something better that could be used here...

bew:
> The messy asciidoc syntax really consumes too much time. I wish there was something better that could be used here...

There might be something possible, by generating AST of asciidoc and printing its string version?
Would be in ruby, some notes about sth close in https://github.com/asciidoctor/asciidoctor/issues/2378

t-wissmann:
Yes, but I doubt it is easy to generate the string representation from an AST because paragraphs and code blocks in itemizes are very nasty.

t-wissmann:
I've now fixed the test case to also support multi-line doc strings in the output of the `help` command

blarz (approved):
Checked generated html output and that all settings were migrated correctly. Looks good to me! :)

PR #1408, opened 2021-12-21
Check early on click if client resize is possible

### LINKED ISSUE #1407: Usability regression in #1404
Before https://github.com/herbstluftwm/herbstluftwm/pull/1404 it was possible to click the top edge of the monitor to focus clients when frame_gap was set to 0. This means you don't need to aim.

With https://github.com/herbstluftwm/herbstluftwm/pull/1404 you now have to click frame_width(?) pixels down from the edge.

As https://github.com/herbstluftwm/herbstluftwm/pull/1404 solves a real issue I would suggest adding a check if the resize action is for the top edge of the monitor, and skipping it. This shouldn't have any downsides as resizing on the top edge isn't something that makes all that much sense to do.

t-wissmann:
When clicking a client's border, check it early whether the client can
be resized at all. If it can not be resized, then focus the client
instead, because this is what the (normal) cursor shape indicates.

This closes #1407, a bug introduced by PR #1404.

### DIFF
--- a/src/xmainloop.cpp
+++ b/src/xmainloop.cpp
@@ -246,6 +246,9 @@ void XMainLoop::buttonpress(XButtonEvent* be) {
             bool decorationClicked = be->window == client->decorationWindow();
             if (decorationClicked) {
                 resize = client->dec->positionTriggersResize({be->x, be->y});
+                if (resize) {
+                    resize = resize * client->possibleResizeActions();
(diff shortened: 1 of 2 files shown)


blarz (approved):
Thanks for fixing this!

t-wissmann:
Thanks for merging :)

PR #1410, opened 2021-12-30
Add theme options for unselected tabs

### LINKED ISSUE #1389: Separate "normal" color settings for unfocused tabs
Loving the new tab support!

Right now, if you have several windows stacked up with tabs across the top, the currently active window's tab gets the "active" coloring. But all of the inactive tabs have the same "normal" coloring that unfocused windows have. While this makes some sense, in practice it can be a little jarring.

For example, on my current theme, the active window it red, while all unfocused windows are white. This used to be fine, the white made the title seem invisible against the white window contents. However, I now have unfocused tabs in white as well, which makes them stand out next to the red.

They are technically unfocused, but they are still part of the active frame, so a separate color (in my case, something dark rather than white, so the red continues to stand out) would be nice.

t-wissmann:
In order to distinguish unselected from selected tabs, this adds new
attributes for controlling tab appearance. Those attributes can also be
empty, which will cause the normal attributes to be used (as before);
this makes the change 100% backwards compatible.

This implements what has been requested during the discussion of #1388
and closes #1389.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,9 +11,10 @@ Current git version
     'tabbed_max' setting)
   * New decoration setting 'title_when' to control, when window titles and tabs
     are shown.
-  * New decoration setting 'title_depth'.
+  * New decoration settings 'title_depth' and 'title_align'.
+  * New decoration settings for configuring unselected tabs: 'tab_color',
+    'tab_outer_width', 'tab_outer_color', 'tab_title_color'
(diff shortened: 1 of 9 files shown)


t-wissmann:
tab-colors

blarz (approved):
Looks good to me! :)

t-wissmann:
Thanks!

PR #1411, opened 2022-01-03
Do not map already visible unmanaged windows

t-wissmann:
In the MapNotify for an unmanaged window, manage_client() is called,
which should not call XMapWindow(), because the window is already
visible. When XMapWindow() was still called, this could trigger wrong
behaviour if a x11 client quickly maps and then unmaps the window,
because hlwm would then map the window again.

Concretely, this bug triggered an endless loop when pressing Alt in
pcmanfm (one should only test this in Xephyr, otherwise the session
hangs).

The new test case quickly maps and unmaps a window, and verifies that
hlwm does not try to map the window again.

### DIFF
--- a/src/clientmanager.cpp
+++ b/src/clientmanager.cpp
@@ -208,7 +208,8 @@ Client* ClientManager::manage_client(Window win, bool visible_already, bool forc
     auto stdio = OutputChannels::stdio();
     changes = Root::get()->rules()->evaluateRules(client, stdio, changes);
     if (!changes.manage || force_unmanage) {
-        // map it... just to be sure
+        // if the window becomes unmanaged and wasn't visible before,
+        // then map it.
         if (!visible_already) {
             XMapWindow(X_->display(), win);
         }
--- a/src/clientmanager.cpp
+++ b/src/clientmanager.cpp
@@ -209,7 +209,9 @@ Client* ClientManager::manage_client(Window win, bool visible_already, bool forc
     changes = Root::get()->rules()->evaluateRules(client, stdio, changes);
     if (!changes.manage || force_unmanage) {
         // map it... just to be sure
-        XMapWindow(X_->display(), win);
+        if (!visible_already) {
+            XMapWindow(X_->display(), win);
+        }
         delete client;
         return {};
     }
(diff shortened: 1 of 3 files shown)


tests/test_basic_events.py
```diff
@@ -143,3 +144,31 @@ def test_enternotify_do_not_drop_events(hlwm, mouse, client_count):
...
+    handle, winid = x11.create_client(force_unmanage=True, pre_map=pre_map, pre_sync=pre_sync)
+    map_notify_events = []
+    events = x11.pop_pending_events()
+    for e in events:
+        if e.type == Xlib.X.MapNotify:
+            map_notify_events.append(e)
+
+    # the rules are evaluated correctly:
+    assert ['rule', 'tempmanage', winid] in hc_idle.hooks()
+    # and thlwm didn't try to call XMapWindow:
```
blarz, line 173: Typo thlwm

src/clientmanager.cpp
```diff
@@ -209,7 +209,9 @@ Client* ClientManager::manage_client(Window win, bool visible_already, bool forc
     changes = Root::get()->rules()->evaluateRules(client, stdio, changes);
     if (!changes.manage || force_unmanage) {
         // map it... just to be sure
```
blarz, line 211: Should we update the comment?

blarz (commented):
Just two minor suggestions

tests/test_basic_events.py
t-wissmann: [x] fixed

src/clientmanager.cpp
t-wissmann: Yes, I've tried to say something more meaningful.

tests/test_basic_events.py
blarz: Thanks!

src/clientmanager.cpp
blarz: Awesome, thank you!

blarz (approved):
Looks good, verified the fixed endless loop with Xephyr!

t-wissmann:
Thanks for testing!

PR #1412, opened 2022-01-03
Fix input focus for preexisting clients

t-wissmann:
This correctly puts the input focus to clients that exist before hlwm.
This is a bug that was brought to the surface by PR #1403, because now,
hlwm only calls XSetInputFocus() if the focus actually changes.

The solution now is: the decoration window is 'unmapped' if and only if
the client window itself is 'unmapped'. This is true already for all
clients that show up during the hlwm execution, but wasn't true for
clients existing before. With this property, making the client visible
(in Client::set_visible()) will actually set the client to the 'mapped'
state, triggering a MapNotify event, whose handler then calls
XSetInputFocus() if necessary.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -204,6 +204,9 @@ void Client::make_full_client() {
         // point, we did not call XSelectInput() yet, so the above
         // XReparentWindow does not trigger an event for the window.
         visible_ = false;
+        // we also unmap the window such that the decoration window is
+        // in 'mapped' state f and only if the window itself is so.
+        XUnmapWindow(X_.display(), window_);
     }
     // get events from window
     XSelectInput(X_.display(), dec->decorationWindow(), (EnterWindowMask | LeaveWindowMask |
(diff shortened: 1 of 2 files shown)


src/client.cpp
```diff
@@ -204,6 +204,9 @@ void Client::make_full_client() {
         // point, we did not call XSelectInput() yet, so the above
         // XReparentWindow does not trigger an event for the window.
         visible_ = false;
+        // we also unmap the window such that the decoration window is
+        // in 'mapped' state f and only if the window itself is so.
```
blarz, line 208: "i" missing

blarz (commented):
Just a typo

src/client.cpp
t-wissmann: [x] fixed

blarz: Thanks!

blarz (approved):

PR #1413, opened 2022-01-04
Prevent endless focus stealing loop

t-wissmann:
I think that this now solves the issue of the endless FocusIn event loop
triggered by #1386. However, I still do not know how to reproduce the
endless loop. In the present solution, I avoid that the handler of
FocusIn triggers further FocusIn or similar events (via the member
duringFocusIn_).

While at it, I've moved all calls to XSetInputFocus() to a custom
setInputFocus() function that possibly falls back to the WM_TAKE_FOCUS
method if the client requests so.


blarz (approved):
Looks good to me!

Issue #1414, opened 2022-01-08
Testsuite fails on big-endian systems


glaubitz:
The testsuite fails during `test_focus_on_click` test with the following backtrace:

```
============================= test session starts ==============================
platform linux -- Python 3.9.7, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
rootdir: /<<PKGBUILDDIR>>, configfile: tox.ini
plugins: forked-1.3.0, xvfb-1.2.0, xdist-2.4.0
collected 2129 items

../tests/test_basic_events.py E

==================================== ERRORS ====================================
_______________ ERROR at setup of test_focus_on_click[True-True] _______________

hlwm_process = <conftest.HlwmProcess object at 0x7fffb4c23040>
hlwm = <conftest.HlwmBridge object at 0x7fffb4c23df0>

    @pytest.fixture()
    def mouse(hlwm_process, hlwm):
        class Mouse:
            def __init__(self):
                self.move_to(0, 0, wait=True)
    
            def move_into(self, win_id, x=1, y=1, wait=True):
                if wait:
                    with hlwm_process.wait_stderr_match('EnterNotify'):
                        # no --sync here, because we're waiting for the EnterNotify anyways
                        self.call_cmd(f'xdotool mousemove --window {win_id} {x} {y}', shell=True)
                    # reaching this line only means that hlwm started processing
                    # the EnterNotify. So we need to wait until the event is fully processed:
                    hlwm.call('true')
                else:
                    self.call_cmd(f'xdotool mousemove --sync --window {win_id} {x} {y}', shell=True)
    
            def click(self, button, into_win_id=None, wait=True):
                if into_win_id:
                    # no need to wait for herbstluftwm to process, we're just positioning the mouse to click
                    self.move_into(into_win_id, wait=False)
                if wait:
                    with hlwm_process.wait_stderr_match('ButtonPress'):
                        subprocess.check_call(['xdotool', 'click', button])
                    # reaching this line only means that hlwm started processing
                    # the ButtonPress. So we need to wait until the event is fully processed:
                    hlwm.call('true')
                else:
                    subprocess.check_call(['xdotool', 'click', button])
    
            def move_to(self, abs_x, abs_y, wait=True):
                abs_x = str(int(abs_x))
                abs_y = str(int(abs_y))
                self.call_cmd(f'xdotool mousemove --sync {abs_x} {abs_y}', shell=True)
                if wait:
[...]

PR #1417, opened 2022-01-11
tests: Register herbstluftwm module for assertion rewriting

The-Compiler:
By default, pytest has its assertion rewriting disabled for code outside of the
tests and conftest modules, to avoid testing different code than what runs in
production.

However, we have assertions in the herbstluftwm module, so it would be good to
benefit from rewriting there, too.

Docs: https://docs.pytest.org/en/latest/how-to/writing_plugins.html#assertion-rewriting

For a test such as:

    def test_failing(hlwm):
        hlwm.call("false")

This turns the output:

    >       assert proc.returncode == 0
    E       AssertionError

into output with additional assertion information:

    >       assert proc.returncode == 0
    E       AssertionError: assert 1 == 0
    E        +  where 1 = CompletedProcess(args=['.../build/herbstclient', '-n', 'false'], returncode=1, stdout='', stderr='').returncode

Which will hopefully help making diagnosing issues such as #1414 easier.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,9 +14,11 @@ import sys
 import textwrap
 import time
 import types
-import herbstluftwm
 
 import pytest
+
+pytest.register_assert_rewrite("herbstluftwm")
+import herbstluftwm  # noqa: E402
 
 
 BINDIR = os.path.join(os.path.abspath(os.environ['PWD']))


t-wissmann:
I see, makes sense, thanks!

only flake8 is a bit picky about the change:
```
flake8 run-test: commands[0] | flake8 .
./tests/conftest.py:21:1: E402 module level import not at top of file
```

The-Compiler:
Well, it's correct in principle, but this is a common exception, as we need to do this *before* the import. I thus silenced the warning on that line.

t-wissmann (approved):
Thanks! Let's see if it helps

PR #1422, opened 2022-01-24
panel.sh: Remove 2 forks/second per panel

johnLate:
I noticed the pids on my system getting quite big somewhat fast. With dual monitors, I had ≥ 4 forks/sec or ≥ 14400 forks/hour just from panel.sh.

While this is not a big issue, I hope you find these changes useful.

### DIFF
--- a/share/panel.sh
+++ b/share/panel.sh
@@ -5,6 +5,12 @@ quote() {
 	printf '%s' "${q% }"
 }
 
+if [[ -f /usr/lib/bash/sleep ]]; then
+    # load and enable 'sleep' builtin (does not support unit suffixes: h, m, s!)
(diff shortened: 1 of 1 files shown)


johnLate:
> How did you measure this?

In bash run this:

```bash
enable -f /usr/lib/bash/sleep sleep
prev=0; while true; do read last < /proc/sys/kernel/ns_last_pid ; echo $((last-prev)) ; prev=$last ; sleep 1 ; done
```

With the old panel.sh I observed 4 as lowest value, after my patches it's 0.

t-wissmann (approved):
Thanks! I've tested it locally and it works fine.

Of course, I can strongly recommend [barpyrus](https://github.com/t-wissmann/barpyrus) which does not require forks at all.

PR #1423, opened 2022-01-25
Clamp to XRectangle dimensions

### LINKED ISSUE #1418: Inconsistent border colour with equal border_width-, inner_width-, and outer_widths in max layout.
With `border_width == inner_width == outer_width` in the `max` layout you get this somewhat odd behaviour:

hlwm_bw=iw=ow

Contrast this with other layouts or floating windows where the order of precedence is outer > inner > basic.

Having a chequered border might be neat, and this doesn't look too bad with two or three windows, but I don't think it's intentional?

Edit: settings used in the gif:
```
INNER_COLOR='#ebdbb2'
OUTER_COLOR=$INNER_COLOR
ACTIVE_TITLE_COLOR='#f2e5bc'
ACTIVE_COLOR='#282828'
NORMAL_TITLE_COLOR='#282828'
NORMAL_COLOR='#ebdbb2'
```

t-wissmann:
If inner_width, outer_width, and border_width are identical, then this
makes one of the rectangle heights negative. Since heights of XRectangle
are of type unsigned short, this caused a very large rectangle. To fix
this, the helper function toUnsigned() now correctly casts widths and
heights to unsigned short.

This fixes #1418

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -480,13 +480,23 @@ void Decoration::redrawPixmap() {
     auto inner = last_inner_rect;
     inner.x -= last_outer_rect.x;
     inner.y -= last_outer_rect.y;
+    // convert signed and possibly negative integers to
+    // unsigned short, which is used by XRectangle.
+    auto toUnsigned =
+            [](int length) -> unsigned short {
+        if (length < 0) {
+            return 0;
+        } else {
+            return static_cast<unsigned short>(length);
+        }
+    };
     if (iw > 0) {
         /* fill rectangles because drawing does not work */
         vector<XRectangle> rects{
-            { (short)(inner.x - iw), (short)(inner.y - iw), (unsigned short)(inner.width + 2*iw), iw }, /* top */
-            { (short)(inner.x - iw), (short)(inner.y), iw, (unsigned short)(inner.height) },  /* left */
-            { (short)(inner.x + inner.width), (short)(inner.y), iw, (unsigned short)(inner.height) }, /* right */
-            { (short)(inner.x - iw), (short)(inner.y + inner.height), (unsigned short)(inner.width + 2*iw), iw }, /* bottom */
+            { (short)(inner.x - iw), (short)(inner.y - iw), toUnsigned(inner.width + 2*iw), toUnsigned(iw) }, /* top */
+            { (short)(inner.x - iw), (short)(inner.y), toUnsigned(iw), toUnsigned(inner.height) },  /* left */
+            { (short)(inner.x + inner.width), (short)(inner.y), toUnsigned(iw), toUnsigned(inner.height) }, /* right */
+            { (short)(inner.x - iw), (short)(inner.y + inner.height), toUnsigned(inner.width + 2*iw), toUnsigned(iw) }, /* bottom */
         };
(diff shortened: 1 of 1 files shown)


blarz:
Looks good to me!

Would it make sense to add a regression test for the scenario described above? (inner_width = outer_width = border_width)

t-wissmann:
> Looks good to me!
> 
> Would it make sense to add a regression test for the scenario described above? (inner_width = outer_width = border_width)

There was no crash or error message, so one would need to investigate the screenshot of the decoration (we have a couple of such tests already), but right now I would say that it is not worth the effort.

blarz:
> > Looks good to me!
> > Would it make sense to add a regression test for the scenario described above? (inner_width = outer_width = border_width)
> 
> There was no crash or error message, so one would need to investigate the screenshot of the decoration (we have a couple of such tests already), but right now I would say that it is not worth the effort.

Alright!

blarz (approved):

t-wissmann:
By the way, the current CI failure is unrelated (neglectable): in the big completion test case (where 4000 commands are run), hlwm did not shut down in time:

```
[gw0] [ 92%] ERROR tests/test_completion.py::test_completable_commands[False] 

==================================== ERRORS ====================================
____________ ERROR at teardown of test_completable_commands[False] _____________
[gw0] linux -- Python 3.8.10 /__w/herbstluftwm/herbstluftwm/.tox/py38/bin/python

hlwm_spawner = <function hlwm_spawner.<locals>.spawn at 0x7fc65f852ca0>
request = <SubRequest 'hlwm_process' for <Function test_completable_commands[False]>>
xvfb = <conftest.MultiscreenDisplay object at 0x7fc65edc1e20>

    @pytest.fixture()
    def hlwm_process(hlwm_spawner, request, xvfb):
        parameters = {
            'transparency': True,
        }
        if hasattr(request, 'param'):
            # the param must contain a dictionary possibly
            # updating the default parameters
            parameters.update(request.param)
    
        """Set up hlwm and also check that it shuts down gently afterwards"""
        additional_commandline = ['--no-tag-import']
        if not parameters['transparency']:
            additional_commandline += ['--no-transparency']
        hlwm_proc = hlwm_spawner(additional_commandline, display=xvfb.display)
    
        yield hlwm_proc
    
>       hlwm_proc.shutdown()

tests/conftest.py:614: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <conftest.HlwmProcess object at 0x7fc65edc1df0>

    def shutdown(self):
        self.proc.terminate()
    
        # Make sure to read and echo all remaining output (esp. ASAN messages):
        self.read_and_echo_output(until_eof=True)
        self.proc.stdout.close()
        self.proc.stderr.close()
    
        if self.proc.returncode is None:
            # only wait the process if it hasn't been cleaned up
            # this also avoids the second exception if hlwm crashed
            try:
>               assert self.proc.wait(PROCESS_SHUTDOWN_TIME) == 0
E               assert -13 == 0
E                 +-13
E                 -0

tests/conftest.py:461: AssertionError
```

PR #1428, opened 2022-01-27
Use xmllint step of a2x for patching manpages

t-wissmann:
This essentially reverts the cmake changes of f61211ad0f (#1396) and
instead injects the patch-manpage-xml.py via asciidoc's xmllint step.

The positive effect of this is that we do not need to know the path used
in ASCIIDOC_MANPAGE_XSL. Recently on arch linux, it was moved from

    /etc/asciidoc/docbook-xsl/manpage.xsl

to

    /usr/lib/python3.10/site-packages/asciidoc/resources/docbook-xsl/manpage.xsl

which of course depends on a lot of parameters (python version, etc).
Since I couldn't find a way to let asciidoc tell me the path, so having
the path a cmake parameter will presumably cause a lot of packaging
headache.

Instead, I noticed, that we could abuse the 'xmllint' step of the
asciidoc compilation to inject a tool that patches the man page xml.
As a side-effect, this changes make the documentation build again
successfully out of the box on arch linux.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -78,7 +78,7 @@ function(gen_manpage sourcefile man_nr)
         # our script patching the man page xml instead of running xmllint
         set(custom_asciidoc_cfg "${CMAKE_CURRENT_BINARY_DIR}/a2x-patched-dont-install.conf")
         file(WRITE ${custom_asciidoc_cfg}
-        "XMLLINT = \'${CMAKE_CURRENT_SOURCE_DIR}/patch-manpage-xml.py\'")
+        "XMLLINT = '${CMAKE_CURRENT_SOURCE_DIR}/patch-manpage-xml.py'")
         add_custom_command(
             OUTPUT ${dst}
             COMMAND ${ASCIIDOC_A2X}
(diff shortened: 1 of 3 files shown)


doc/CMakeLists.txt
```diff
@@ -75,34 +74,22 @@ function(gen_manpage sourcefile man_nr)
...
+        "XMLLINT = '${CMAKE_CURRENT_SOURCE_DIR}/patch-manpage-xml.py'")
         add_custom_command(
-            OUTPUT ${dst} ${docbookxml}
-            # first create the docbook xml:
-            COMMAND ${ASCIIDOC}
-                    -o ${docbookxml}
-                    -b docbook45
+            OUTPUT ${dst}
+            COMMAND ${ASCIIDOC_A2X}
+                    --verbose
```
blarz, line 85: Do we still want this or was this just for debugging purposes?

t-wissmann: I'd like to keep it because otherwise, a2x shows too little information.

blarz: Alright! Just wanted to make sure this is intended :)

blarz (approved):
Nice workaround :)

t-wissmann:
Thanks for reviewing (this and the other PR) :-)

PR #1429, opened 2022-01-28
Premultiply colors with alpha if compositing

### LINKED ISSUE #1382: Alpha of colors is still interpreted wrong
image

The old issue with semi-transparent windows still exists in the color management of the `FrameDecoration` class.

t-wissmann:
This fixes the color handling of rgba colors. The compositing of X11
expects the true transparency colors to be 'premultiplied' with the
alpha value. That is, a color r=g=b=80 with 50% alpha should have
r=g=b=40 (see also [1,2]).

This bug was still present in colors for both frame decorations and
normal decorations.

Due to this premultiplication, colors would appear darker if the
premultiplication was applied also in absence of a compositing manager
(aka compositor). Thus, this change also includes new code that detects
the presence and absence of a compositing manager at run time, using the
xfixes X11 extension library (which should be available on all systems
that have X11).

The new test cases essentially check that hlwm toggles this
premultiplication depending on whether a compositor claims to exist.

On the implementation side, this color conversion was moved to
XConnection, and so a few Color member functions are not needed anymore.

This fixes #1382.

[1] https://stackoverflow.com/q/39505100/4400896
[2] https://stackoverflow.com/a/39927205/4400896

### DIFF
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -63,6 +63,47 @@ XConnection* XConnection::connect(string display_name) {
     return s_connection;
 }
 
+/**
+ * @brief convert the given color via the given color map
+ * or via the default colormap is none is given
+ * @param maybeColormap is a colormap or 0
+ * @param color
+ * @return
+ */
+unsigned long XConnection::allocColor(Colormap maybeColormap, const Color& color)
+{
+    XColor xcol = color.toXColor();
+    if (maybeColormap) {
+        /* get pixel value back appropriate for client */
+        /* this possibly adjusts xcol */
+        XAllocColor(display(), maybeColormap, &xcol);
+    }
+    // explicitly set the alpha-byte to the one from the color
+    if (usesTransparency() && compositorRunning_ && color.alpha_ != 0xffu) {
+        auto scaleValueByAlpha =
+                [](unsigned short colorValue, unsigned short alpha) -> unsigned long  {
+            // color value is between 0 and 0xffff
+            // alpha value is between 0 and 0xff
+            // return the multiplied color value between 0 and 0xff
+            return (static_cast<unsigned long>(colorValue)
+                    * static_cast<unsigned long>(alpha))
+                    / 0xffffu;
+        };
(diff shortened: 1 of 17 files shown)


t-wissmann:
This is how the premultiplied alpha looks:

premultiplied-alpha

Here is the run-time detection of compositors in action:
compositor-detection

t-wissmann:
The builds fail because  the docker images lack the `xfixes` library.

tiosgz:
> composting manager

laughing hard at this typo :D

a few other minor things: system`[s]`; test cases … check`s`; comp`i`sitor

t-wissmann:
> > composting manager
> 
> laughing hard at this typo :D
> 
> a few other minor things: all system`[s]`; test cases … check`s`; comp`i`sitor

Thanks a lot! Did I fix all you mentioned? (I became impatient while writing the message...)

tiosgz:
> Did I fix all you mentioned?

yep, thanks!

t-wissmann:
I've rebuilt the docker images and pushed them, so now, it should compile :-)

src/xconnection.cpp
```diff
@@ -63,6 +63,47 @@ XConnection* XConnection::connect(string display_name) {
     return s_connection;
 }
 
+/**
+ * @brief convert the given color via the given color map
+ * or via the default colormap is none is given
```
blarz, line 68: if none is given?

src/xconnection.cpp
```diff
@@ -176,6 +217,17 @@ void XConnection::tryInitTransparency()
     XFree(infos);
 }
 
+/**
+ * @brief tell xconnection whether a compositor (aka compositing manager)
+ * is running at the momenet. this affects the color computation.
```
blarz, line 222: typo moment

blarz (commented):
Just two typos.

Cool that you worked around the Dockerhub issue!

t-wissmann:
Thanks for reviewing! I've fixed both typos

blarz (approved):
Thanks!

PR #1430, opened 2022-01-28
Use same visual for all windows in the decoration

t-wissmann:
Even though I'm not aware of any bad consequences of different visuals,
it might be cleaner to have the same 'Visual' structure in all windows
that are nested in the decoration window.

### DIFF
--- a/src/decoration.cpp
+++ b/src/decoration.cpp
@@ -101,7 +101,7 @@ void Decoration::createWindow() {
     dec->bgwin = XCreateWindow(display, dec->decwin, 0,0, 30, 30, 0,
                         dec->depth,
                         InputOutput,
-                        xcon.visual(),
+                        visual ? visual : xcon.visual(),
                         mask, &at);
     XMapWindow(display, dec->bgwin);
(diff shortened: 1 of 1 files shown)


tiosgz:
> Even though I don't know of any bad consequences of this, it might be
> cleaner to have the same 'Visual' structure in all windows that are
> nested in the decoration window.

Sorry for being picky about commit messages today… but i’m having trouble making sense of this. Is it meant as follows?

> Even though I don't know of any bad consequences of not doing this, …

t-wissmann:
> > Even though I don't know of any bad consequences of this, it might be
> > cleaner to have the same 'Visual' structure in all windows that are
> > nested in the decoration window.
> 
> Sorry for being picky about commit messages today… but i’m having trouble making sense of this. Is it meant as follows?
> 
> > Even though I don't know of any bad consequences of not doing this, …

Thanks for reading my commit messages, this is much appreciated! I tried to clarify the 'Even though...' part.

tiosgz:
> I tried to clarify the 'Even though...' part.

thanks!

blarz (approved):

PR #1432, opened 2022-01-30
Extend 'smart_frame_surroundings' for more granular control

gravndal:
It now accepts 'off', 'hide_gaps', and 'hide_all'.

In addition to what was previously possible, this change enables only hiding frame gaps when the frame tree consists solely of the one root frame.

This change is not fully backwards compatible, see MIGRATION.


gravndal:
Seems I forgot to specifically ask: Of the two ways to do this, which would be preferred?
1. Add a new boolean setting like `smart_frame_surroundings_gap_only`.
2. Convert the current setting to a string with possible values `0, 1, false, true, gap`, and maybe even `frame`.

tiosgz:
IIUC, the answer may possibly be #1433 – but take that with a pinch of salt.

gravndal:
Yeah, I just saw that. A general overhaul of the theming capabilities would certainly be a welcome, if more long term, solution.

t-wissmann:
Sorry for the confusion. I wrote an answer in this PR but it seems I forgot to send it and so the answer was lost...

I really appreciate the effort here, especially that you also updated the doc string :-)
In principle, I understand the usecase, but I'm very reluctant regarding regarding incompatible changes (see the Migration file), but maybe they are ok in this case. In the long run, I'd like to go for the more general theming config (#1433 as mentioned by @tiosgz), and your PR was the reason why I finally put my ideas into #1433 (after being in my head for the past couple of weeks).

Regarding the implementation, I'd go for option 2, using the template type `Finite<>`, and then one can call the options `hide_all` (with alias `on`, `true`), `hide_gaps`, `off` (with alias `false`).

Since templates are always a bit hard to get started with, I think the type definitions should be the following:
```c++
// in settings.h:

enum class SmartFrameSurroundings {
    hide_all,
    hide_gaps,
    off,
};

template <>
struct is_finite<SmartFrameSurroundings> : std::true_type {};
template<> Finite<SmartFrameSurroundings>::ValueList Finite<SmartFrameSurroundings>::values;
template<> inline Type Attribute_<SmartFrameSurroundings>::staticType() { return Type::NAMES; }




// in settings.cpp:

template<>
Finite<SmartFrameSurroundings>::ValueList Finite<SmartFrameSurroundings>::values = ValueListPlain {
    { SmartFrameSurroundings::hide_all, "hide_all" },
    { SmartFrameSurroundings::hide_all, "on" },
    { SmartFrameSurroundings::hide_all, "true" },
    { SmartFrameSurroundings::hide_all, "1" },
    { SmartFrameSurroundings::hide_gaps, "hide_gaps" },
    { SmartFrameSurroundings::off, "off" },
    { SmartFrameSurroundings::off, "false" },
    { SmartFrameSurroundings::off, "0" },
};

```

Then, you can make the setting have type `Attribute_<SmartFrameSurroundings>` and `Finite<>` will take care of parsing/printing/tab-completion (a similar use-case is the `title_when` attribute).

Also, the first message in the PR will be used as the commit message, so there, you should describe the changes by your change. Also, the change should be documented in `NEWS` and `MIGRATION`.

Additionally, there should be some test cases (see e.g. `title_when`), you can try, but if it is too much I can also take care of it.

If you want to implement it, I think I will merge it. (I don't want to block features again like I did during the winterbreeze rewrite. Also the aliases hopefully make it quite backwards-compatible).

gravndal:
Updated the pr.

> but if it is too much I can also take care of it.

As you've already basically done all the actual work here, I think I'll leave that to your much more capable hands.

gravndal:
Forgot to rebase.

t-wissmann:
> Updated the pr.
> 
> > but if it is too much I can also take care of it.
> 
> As you've already basically done all the actual work here, I think I'll leave that to your much more capable hands.

Can't you at least try? I don't know when I want to invest time into it.

gravndal:
> Can't you at least try?

Are you trying to guilt me into learning your test suite? That might actually be a working strategy ;)

> I don't know when I want to invest time into it.

Same, I might give it a go in a week or two.

t-wissmann:
Hi, I accidentally pushed and now can't re-open this PR anymore.
Can you please open a new PR after fetching my changes from
the `smarter_frame_surroundings` branch in this repository?

gravndal:
@t-wissmann https://github.com/herbstluftwm/herbstluftwm/pull/1464

PR #1436, opened 2022-02-02
eliminate autostart bashism, tag reconfiguration

madroach:
Actually two changes:
- I just don't like depending on bash when plain borne shell suffices. Also I believe the code is more readable without array index acrobatics, but instead using the `read` ideom.
- the tag index is impossible to guess when tags are _renamed_ during autostart reload, because the old names are not removed. (The old names cannot be guessed either). Therefore use `use` and `move` instead of `use_index` and `move_index`.

### DIFF
--- a/share/autostart
+++ b/share/autostart
@@ -63,18 +63,31 @@ hc keybind $Mod-Control-Up      resize up +$resizestep
 hc keybind $Mod-Control-Right   resize right +$resizestep
 
 # tags
-tag_names=( {1..9} )
-tag_keys=( {1..9} 0 )
-
-hc rename default "${tag_names[0]}" || true
-for i in "${!tag_names[@]}" ; do
-    hc add "${tag_names[$i]}"
-    key="${tag_keys[$i]}"
+i=0
+while read name key ; do
+    [ -n "$name" ] || continue
+    if [ $i -eq 0 ] ; then
+        [ -n "$name" ] && hc rename default "$name"
+    else
+        [ -n "$name" ] && hc add "$name"
+    fi
     if [ -n "$key" ] ; then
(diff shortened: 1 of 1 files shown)


t-wissmann:
Hi, to be honest, the old version of the autostart is much more readable. Also I'm not sure whether you are really referring to plain [_bourne_ shell](https://en.wikipedia.org/wiki/Bourne_shell) or to posix shell.
I think it's a general consensus that `for each i in …`-style loops are much more intuitive than manual `while` loops involving I/O. If you want the indices of the tags to be precisely as specified in the array, then after `hc add "${tag_names[$i]}"`, you can add
```bash
hc attr tags.by-name."${tag_names[$i]}".index = "$i"
```
(We probably should add this to the default autostart)

madroach:
To me its mostly about removing bash from dependencies, and I'm thinking about posix shell.

But still the for loop is indeed not readable in this case. It loops over indices extracted from one array. Then it uses the index to extract strings from two different arrays. In the end you try to simulate a numbered list of pairs here in a very non idiomatic way. Using read in a while loop is [idiomatic shell code](https://bash.cyberciti.biz/guide/While_loop) that gets closest to a numbered list of records. The here document can easily be edited and you can instantly see which name maps to which number. That is not the case for the two arrays. To what name is the trailing `0` in `tag_keys` supposed to map anyway?

I won't press this any further. Its just a config file after all…

your suggestion of adding
```
hc attr tags.by-name."${tag_names[$i]}".index = "$i"
```
might indeed be helpful. But won't indices collide?

madroach:
I see bashisms are used more heavily in other scripts, so no use in trying to remove them.

PR #1437, opened 2022-02-02
Move focus horizontally through tabbed_max frame

madroach:
This allows moving the focus through windows in a tabbed max frame using the directional `focus left` and `focus right`. This is intuitive, because the tabs visually suggest a horizontal layout.


t-wissmann:
In principle, this seems to be a reasonable change. However, I don't think that everybody wants `focus` to navigate within a tabbed frame. There is still the `cycle` command. So I think this should be configurable by a setting like `directional_focus_within_tabbed_max` (I don't have a better name). Also there should be test cases for both the new and the old behaviour of `focus` and `shift` (note that this also affects how `shift` behaves)

tiosgz:
> I think this should be configurable by a setting like `directional_focus_within_tabbed_max` (I don't have a better name)

what about `focus -t` or something alike? (note that it is basically a variant of `-i`, & thus implies it.)

i wonder if hlwm’s eventually going to support tabs & stack, like i3/sway 😜

madroach:
I would prefer changing `focus` (and `shift`) by adding proposed `-t`. Is this ok with you, @t-wissmann ?

t-wissmann:
Yes, adding `-t` is ok. (it's only a bit more complicated to implement because you have to propagate this flag through the function calls, but usability-wise it's fine!

t-wissmann:
Maybe, the implementation is easier if you replace `bool external_only` in all the function calls with a class
```c++
class DirectionFlags {
public
   bool externalOnly;
   bool withinTabs;
};
```

madroach:
Ok. I implemented additional flags for `focus` / `shift`. I call them `-ii`, `-iii`.
Instead of the class you suggested I use an ordered `enum class` of `Frame = 0, Visible = 1, Tabs = 2, All = 3`.
Larger means more clients may be selected.

t-wissmann:
Thanks for implementing it. This flag repetition `-i` is a very non-standard thing (it is only used to increase verbosity level in some applications, and even there it is hacky). I suggest you instead use a named parameter like `--level=frame|visible|tabs|all`, because those are more self-explanatory (note that you used these words in the explanation, so these words should also be used in the interface). For example the `listClientsCommand()` uses such parametrized flags.

t-wissmann:
Oh, of course you can also call the flags explicitly `"--level=frame"`, `"--level=visible"`, `"--level=tabs"`...

madroach:
I didn't manage to get the _explicit_ variant with `--level=tabs` to compile, so I went through the hassle of adding `Converter` functions for the `enum class DirectionLevel`. I'm new to C++, so if something does not fit your style, feel free to change it.

doc/herbstluftwm.txt
```diff
@@ -425,7 +425,7 @@ split 'ALIGN' ['FRACTION' ['FRAMEINDEX']]::
         * split h
         * split top 0.2 ''   (splits the root frame)
 
-focus ['-i'|'-e'] 'DIRECTION'::
+focus [*--level=*'LEVEL'] 'DIRECTION'::
```
tiosgz, line 428: Oh wait, isn’t this a breaking change?

(Sorry if you’re already planning to handle it properly; didn’t want it to land in `master` unnoticed.)

t-wissmann: You're right! But the flags -i/-e are still handled, so I just added it to the documentation again.

tiosgz: Ah, alrighty then. Thanks!

t-wissmann:
The version with --level=... looked very good. I've only applied the following changes:

  - instead of a Converter-instance, I'm now using Finite<> because it gives tab-completion for free
  - further test cases now entirely cover the PR's diff.
  - the implementation in `getInnerNeighbourIndex()` now does not use case fallthrough. Also, I've rearranged the if-statements such that it is easier to read.

From my point of view, the PR can now be merged as is.

madroach:
Your changes are very fine. Many thanks for your effort.

t-wissmann (approved):
Great, thanks! Then let us merge it!

PR #1438, opened 2022-02-03
Use hlwm colors and font when running dmenu

madroach:
The code to fetch colors is taken from `panel.sh`


tiosgz (changes requested):
could you please fix the quoting, so that e.g. a font name with spaces (which is quite common to see) doesn’t mess it up? (i would use an array for `$dmenu`.)

also, if i’m not mistaken, the default foreground colour is preserved? but that’s actually the same problem with `panel.sh` and there (in both scripts) are some assumptions made anyway, so i think this isn’t worth bothering with.

share/dmenu_run_hlwm
```diff
@@ -5,12 +5,18 @@ if ! command -v dmenu > /dev/null 2>/dev/null ; then
...
+selfg='#101010'
+FLAGS=( -fn "$font" -nb "$bgcolor" -sb "$selbg" -sf "$selfg" )
+
 # Get the currently active tag
 tag=$(herbstclient attr tags.focus.name)
 
 # Redirect to dmenu_path if available
 if command -v dmenu_path > /dev/null 2>/dev/null ; then
-        selectedPath=$(dmenu_path | dmenu "$@")
+        selectedPath=$(dmenu_path | dmenu ${FLAGS[@]} "$@")
```
t-wissmann, line 19: With correct quoting, this would be 
```
selectedPath=$(dmenu_path | dmenu "${FLAGS[@]}" "$@")
```

share/dmenu_run_hlwm
```diff
@@ -22,9 +28,9 @@ elif command -v stest > /dev/null 2>/dev/null ; then
 	fi
 	IFS=:
 	if stest -dqr -n "$cache" $PATH; then
-		selectedPath=$(stest -flx $PATH | sort -u | tee "$cache" | dmenu "$@")
+		selectedPath=$(stest -flx $PATH | sort -u | tee "$cache" | dmenu ${FLAGS[@]} "$@")
 	else
-		selectedPath=$(dmenu "$@" < "$cache")
+		selectedPath=$(dmenu ${FLAGS[@]} "$@" < "$cache")
```
t-wissmann, line 33: Also `"${FLAGS[@]}"` in both lines

share/dmenu_run_hlwm
```diff
@@ -5,12 +5,18 @@ if ! command -v dmenu > /dev/null 2>/dev/null ; then
 	exit 1
 fi
 
+font=$(herbstclient attr theme.title_font)
```
t-wissmann, line 8: I don't think it is a very stable thing to take the font from hlwm's theme, because the font description might be incompatible easily. E.g. hlwm supports xft fonts, whereas dmenu does not support xft out of the box iirc.

t-wissmann (changes requested):

share/dmenu_run_hlwm
madroach: Actually it supports _only_ xft fonts.
Would it be sensible to use the font name from hlwm only if it does not look like a XLFD font description?

madroach: ok. I now did just that. The font is ignored if it matches the pattern of a x font description.

madroach: One can still do `spawn dmenu_run_hlwm -fn $font` in such a case.

t-wissmann: Sorry for not reacting!
You're right, I wasn't aware that dmenu now only supports xft. Thanks for incorporating the change!

t-wissmann (approved):
Thanks  and sorry for the delay! Looks good to me

PR #1443, opened 2022-02-03
Allow binding to key release events

madroach:
This is supposed to work like xbindkeys.
A virtual modifier "Release" will indicate key release instead of key press.

showing the panel only when the modifier is pressed can be done like this:
```
herbstclient keybind Super_L emit_hook togglehidepanel
herbstclient keybind Mod4+Release+Super_L emit_hook togglehidepanel
```

### DIFF
--- a/src/keycombo.cpp
+++ b/src/keycombo.cpp
@@ -150,6 +150,12 @@ KeyCombo KeyCombo::fromString(const string& str) {
     return combo;
 }
 
+bool KeyCombo::match(const KeyCombo& other) const {
+    bool sameMods = modifiers_ & ReleaseMask || modifiers_ == other.modifiers_;
+    bool sameKeySym = keysym == other.keysym;
+    return sameMods && sameKeySym;
+}
+
 bool KeyCombo::operator==(const KeyCombo& other) const {
     bool sameMods = modifiers_ == other.modifiers_;
     bool sameKeySym = keysym == other.keysym;
--- a/src/keycombo.cpp
+++ b/src/keycombo.cpp
@@ -24,6 +24,7 @@ const vector<KeyCombo::ModifierNameAndMask> ModifierCombo::modifierMasks = {
     { "Shift",      ShiftMask },
     { "Control",    ControlMask },
     { "Ctrl",       ControlMask },
+    { "Release",    ReleaseMask },
 };
 
 ModifiersWithString::ModifiersWithString()
--- a/src/keycombo.h
+++ b/src/keycombo.h
@@ -15,6 +15,8 @@
  * Handles the parsing and creation of string representations of itself.
  */
 
+#define ReleaseMask		((unsigned int)1<<31)
+
 class ModifierCombo {
 public:
     unsigned int modifiers_ = 0;
--- a/src/keycombo.h
+++ b/src/keycombo.h
@@ -69,6 +69,7 @@ public:
     KeyCombo() = default;
 
     std::string str() const;
+    bool match(const KeyCombo& other) const;
     bool operator==(const KeyCombo& other) const;
     static KeySym keySymFromString(const std::string& str);
     static KeyCombo fromString(const std::string& str);
--- a/src/keymanager.cpp
+++ b/src/keymanager.cpp
@@ -133,7 +133,7 @@ void KeyManager::handleKeyPress(XKeyEvent* ev) const {
 
     auto found = std::find_if(binds.begin(), binds.end(),
             [=](const unique_ptr<KeyBinding> &other) {
-                return pressed == other->keyCombo;
+                return other->keyCombo.match(pressed);
             });
     if (found != binds.end()) {
         // execute the bound command
--- a/src/xkeygrabber.cpp
+++ b/src/xkeygrabber.cpp
@@ -70,7 +70,8 @@ void XKeyGrabber::ungrabAll() {
 void XKeyGrabber::changeGrabbedState(const KeyCombo& keyCombo, bool grabbed) {
     // List of ignored modifiers (key combo will be grabbed for each of them):
     const unsigned int ignModifiers[] = { 0, LockMask, numlockMask_, numlockMask_ | LockMask };
-    const unsigned int modifiers = keyCombo.modifiers_ & ~ReleaseMask;
+
+    if (keyCombo.modifiers_ & ReleaseMask) return;
 
     KeyCode keycode = XKeysymToKeycode(g_display, keyCombo.keysym);
     if (!keycode) {
@@ -81,10 +82,10 @@ void XKeyGrabber::changeGrabbedState(const KeyCombo& keyCombo, bool grabbed) {
     // Grab/ungrab key for each modifier that is ignored (capslock, numlock)
     for (auto& ignModifier : ignModifiers) {
         if (grabbed) {
-            XGrabKey(g_display, keycode, ignModifier | modifiers, g_root,
+            XGrabKey(g_display, keycode, ignModifier | keyCombo.modifiers_, g_root,
                     True, GrabModeAsync, GrabModeAsync);
         } else {
-            XUngrabKey(g_display, keycode, ignModifier | modifiers, g_root);
+            XUngrabKey(g_display, keycode, ignModifier | keyCombo.modifiers_, g_root);
         }
     }
 }
--- a/src/xkeygrabber.cpp
+++ b/src/xkeygrabber.cpp
@@ -43,6 +43,9 @@ KeyCombo XKeyGrabber::xEventToKeyCombo(XKeyEvent* ev) const {
     combo.modifiers_ = ev->state;
 
     // Normalize
+    if (ev->type == KeyRelease) {
+	combo.modifiers_ |= ReleaseMask;
+    }
     combo.modifiers_ &= ~(numlockMask_ | LockMask);
 
     return combo;
@@ -67,6 +70,7 @@ void XKeyGrabber::ungrabAll() {
 void XKeyGrabber::changeGrabbedState(const KeyCombo& keyCombo, bool grabbed) {
     // List of ignored modifiers (key combo will be grabbed for each of them):
     const unsigned int ignModifiers[] = { 0, LockMask, numlockMask_, numlockMask_ | LockMask };
+    const unsigned int modifiers = keyCombo.modifiers_ & ~ReleaseMask;
 
     KeyCode keycode = XKeysymToKeycode(g_display, keyCombo.keysym);
     if (!keycode) {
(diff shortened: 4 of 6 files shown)


t-wissmann:
Thanks that you put work into implementing it! I think there is still some work to do before this can be merged. There are some test cases one needs to add:
  * When binding both the press and release of a key combo, are the correct commands executed when pressing and releasing?
  * When binding `x`, then binding `Release-x`, and then unbinding `x`, does the keybinding for `Release-x` still work? (and also the other way round). I think this test case would fail at the moment because the `changeGrabbedState` for the two key combos interfere with each other.
  
I know that writing test cases takes a lot of time (for me, it is usually more time than the actual implementation). But they are necessary to keep hlwm maintainable. Also, this change needs to be documented in the man page and in `NEWS`.

gravndal:
To make this more useful it probably shouldn't trigger for modifier keys if they are used in a chorded keybinding before release.

E.g. say use use `Mod4` as your primary modifier when interacting with hlwm, and bind `Mod4-Release-Super_L` to `spawn rofi -show run`. You probably then don't want to actually spawn rofi when doing `Mod4-h` to switch to another window.

t-wissmann:
> To make this (even) more useful it probably shouldn't trigger for modifier keys if they are used in a chorded keybinding before release.
> 
> E.g. say you use `Mod4` as your primary modifier when interacting with hlwm, and bind `Mod4-Release-Super_L` to `spawn rofi -show run`. You probably then don't want to actually spawn rofi when doing `Mod4-h` to switch to another window.

There is no event triggered if only modifiers are pressed, and if Super_L is part of the Super modifier, then pressing it will not trigger any key event.

t-wissmann:
Oh, you are right. I never tried to actively bind a sole modifier key. But in general I still don't see the issue you're describing: it seems that binds act precisely like they were configured.
 (I.e. if you bind `Mod4-Release-Super_L` to `spawn rofi`, then rofi is spawned whenever you release `Super_L` while `Mod4` is active)

gravndal:
> it seems that binds act precisely like they were configured.

Yes, and that technically correct behaviour isn't always going to the what's desired. Simply put: being able to bind a single modifier key to do something without interfering with the rest of the wm is actually useful (see every start menu).

Anyway, this isn't truly that important as workarounds exist as long as you have left over keys that wouldn't be otherwise usable.

madroach:
> * When binding `x`, then binding `Release-x`, and then unbinding `x`, does the keybinding for `Release-x` still work? (and also the other way round). I think this test case would fail at the moment because the `changeGrabbedState` for the two key combos interfere with each other.

To me it seems very difficult to get this right, because XKeyGrab does not differentiate between release and press events. Also I assume you usually want to ignore modifiers when binding a release event further complicating the bind.

It seems that binding to the KeyPress (without modifiers) will also catch any release (ignoring modifiers). Which is what I want and seems sensible to me.

Therefore I pushed two further commits:
* 7635cbf: Do not establish grabs on `Release` key bindings
* 6ea770f: Ignore modifiers when matching `KeyRelease` Events

The downside I see is that doing only a `keybind Release…` won't work without also doing the same `keybind` without `Release`.

I would like to adapt the `keybind` syntax to reflect this and use something like `keybind KEY PRESSCOMMAND RELEASECOMMAND`. Sadly this doesn't work because of the optional command arguments.

Any thoughts on this?

t-wissmann:
> Yes. That's what one would expect.
If this is what everybody would expect, then hlwm should behave this

>But this hassle is not necessary.

What hassle? Do you mean the extra data structure? Or the implementation effort?

> Why would one bind only to key release, but not press?

For binding the screenshot to a key, one would bind only release and not press.

> I'd like to change the api so that the user may only ...

I'm strictly against any api change that is not necessary.


> bind a release _and_ a press command to a key (either one may be noop). Then unbinding will always unbind _both_. Does this make any sense to you?

Sorry, but I don't see the need for this strict api change. I think it's a very bad api design if the user has to specify both a command for press and release events only because it makes the programming part easier.

I can understand that this might be frustrating because you put already work in this PR, but I'd like to keep hlwm's api 1. backwards compatible and 2. easy to use.  And requiring two commands for each keybind violates both points.

madroach:
> What hassle? Do you mean the extra data structure? Or the implementation effort?

I'm thinking about the additional complexity added.

> I don't see the need for this strict api change

Me neither. It can be done without a breaking change:

```
 keybind [--release] KEY COMMAND [ARGS ...]
           Adds a key binding. When KEY is pressed, the internal COMMAND (with
           its ARGS) is executed. A key binding is a (possibly empty) list of
           modifiers (Mod1, Mod2, Mod3, Mod4, Mod5, Alt, Super, Control/Ctrl,
           Shift) and one key (see keysymdef.h for a list of keys). Modifiers
           and the key are concatenated with - or + as separator. If there is
           already a binding for this KEY, it will be overwritten. If --release is
           given, then the internal command is executed on key release.
           Examples:

           •   keybind Mod4+Ctrl+q quit

           •   keybind Mod1-i toggle always_show_frame

           •   keybind Mod1-Shift-space cycle_layout -1

       keyunbind KEY|-F|--all
           Removes the key binding for KEY. The syntax for KEY is defined in
           keybind. If -F or --all is given, then all key bindings will be
           removed.
```

The important part is that `keyunbind` will ungrab _both_: press _and_ release events. This introduces asymmetry between `keybind` and `keyunbind`. It is not exactly what I described above, since the user is not forced to _bind_ both press and release, but he is forced to _unbind_ both.

t-wissmann:
> > What hassle? Do you mean the extra data structure? Or the implementation effort?

> I'm thinking about the additional complexity added.

But this is only complexity on the implementation side. I prefer having a complicated implementation that is easy to use, instead of the other way round.

> The important part is that `keyunbind` will ungrab _both_: press _and_ release events. This introduces asymmetry between `keybind` and `keyunbind`.

Hm I understand. I don't really like the asymmetry, but I also don't see why you need it. Isn't it a perfectly fine Api if you have `Release` as a modifier, and if `keybind` / `keyunbind` both behave exactly as both you and I are expecting them to behave? (I can make an implementation suggestion based on or PR)

madroach:
The Asymmetry reflects the `XGrabKey` / `XUngrabKey` api, which doesn't differentiate between `KeyPress` and `KeyRelease`. Note that `KeyPress` and `KeyRelease` are not symmetric either since 6ea770f. That's why the api I'm proposing is not symmetric either.
What you are asking for is the ability to 'unbind' key presses and releasese seperately. Is this really necessary?

t-wissmann:
> The Asymmetry reflects the `XGrabKey` / `XUngrabKey` api, which doesn't differentiate between `KeyPress` and `KeyRelease`.

1. Note that `XGrabKey` and `XUngrabKey` have perfectly symmetric behaviour. 2. The user does not care how the x11 works internally; the user only interacts with hlwm. 3. at some (not so far in the future) I'd also like to try a wlroots backend for hlwm, why should 

> Note that `KeyPress` and `KeyRelease` are not symmetric either since [6ea770f](https://github.com/herbstluftwm/herbstluftwm/commit/6ea770f649cc9a69dd7edbb1e1bb7567a88d5a05). That's why the api I'm proposing is not symmetric either.

Yes, they are asymmetric because you've implemented them in an asymmetric way.

> What you are asking for is the ability to 'unbind' key presses and releasese seperately. Is this really necessary?

It's not so much about the necessity, but about a clear interface to the user. Assume you run:
```
hc keyunbind --all
hc keybind x spawn xterm
hc keybind Release-x spawn firefox   # (or hc keybind --release x spawn firefox)
hc keyunbind x
```
Would you still expect that the binding for firefox is active or not? If you say 'no', assume you run
```
hc keyunbind --all
hc keybind x spawn xterm
hc keybind Release-x spawn firefox
```
Would you expect that the binding for xterm is active or not?

Note that you and I have arguments on different levels: I first ask how the hlwm user should access the new features and then I'm looking for an implementation, whereas your approach is the other way round. that's of course fine for advanced users (and hlwm's users can be expected to have advanced knowledge), but I'd always go for the better user experience, whenever it can be implemented easily (which is the case here, and I'm about to show that it can be implemented easily by the way)

madroach:
Ok. You convinced me. Thank you for your patiolence. I'll probably try to implement it tomorrow.
Should I then keep ignoring modifiers for key releases or respect modifiers for key release (and add proper support for AnyModifier) ?

madroach:
I just had a quick look. Respecting modifiers is fine with me, but as I already wrote you, in that case I need support for AnyModifier. I'm afraid to implement that, the binding mechanism needs another refactoring. A reference count won't do the trick in that situation, because we need to check for **collisions with AnyModifier**, too. Not just exact matches. I think scanning the binding list on each new (un)binding request will be necessary to see whether there (already / still) is any exactly matching binding or AnyModifier binding. And then ungrab and/or grab the newly (un)bound combo only as necessary. I believe the code would be even simpler than the reference count and not incur any noticable performance penalty either.

t-wissmann:
> I just had a quick look. Respecting modifiers is fine with me, but as I already wrote you, in that case I need support for AnyModifier. I'm afraid to implement that, the binding mechanism needs another refactoring. A reference count won't do the trick in that situation, because we need to check for **collisions with AnyModifier**, too. Not just exact matches.

I don't see the issue. Can you please describe it in simple terms? Simple terms mean: 1. what keys do you press/release, 2. what happens 3. what you would expect hlwm to do instead

madroach:
The behaviour I want is this: The panel is shown only while the `$Mod` key is down. This needs a `XGrabKey(display, $Mod, AnyModifier, …)`.

t-wissmann:
> The behaviour I want is this: The panel is shown only while the `$Mod` key is down. This needs a `XGrabKey(display, $Mod, AnyModifier, …)`.

1. If you want explicit support of 'any modifier', then hlwm should introduce a `Any` modifier ;-)
2. I've just tried my patch (PR #1446), and showing the panel only while Mod1 is pressed is accomplished by ```bash
hc keybind Alt_L emit_hook togglehidepanel
hc keybind Release-Alt-Alt_L emit_hook togglehidepanel
```

madroach:
That's exactly what I used as example in my commit description 0769df2.
But this does _not_ accomplish what I asked for. Try pressing in that order: 'Mod', 'Ctrl'. Then releasing them in the same order: 'Mod', 'Ctrl'. Now the behaviour is inverted and the panel will _hide_ when 'Mod' is down.

That's what I need the 'Any' modifier for. I already implemented it two or three day ago, but then dropped it and made it implied by the 'Release' modifier.

t-wissmann:
> That's exactly what I used as example in my commit description [0769df2](https://github.com/herbstluftwm/herbstluftwm/commit/0769df2e0da43d1537caa0428b2b18f036f76fd9). But this does _not_ accomplish what I asked for.

> Try pressing in that order: 'Mod', 'Ctrl'. Then releasing them in the same order: 'Mod', 'Ctrl'. Now the behaviour is inverted and the panel will _hide_ when 'Mod' is down.

This is the description I was asking for. This thread becomes also frustrating for me, because even for simple questions from my side, it takes two or three comments until I get the actual behaviour that you want to achieve. Do you remember that I asked you for three items? 1. what you do, 2. what hlwm does, 3. what you want hlwm to do. … and I didn't get a clear answer from you. I think I now understand:

1. run `hc keybind Alt_L cmd_a` and `hc keybind Release-Alt_L cmd_b`, then press Alt, then press Ctrl, then release Alt.
2. hlwm runs `cmd_a` but not `cmd_b`
3. hlwm should have run also `cmd_b` because the "corresponding" keybinding was activated.

And I think it is a reasonable request if you want `cmd_b` to be executed also. It is just so frustrating and time-consuming for me, because you mix up (a) expected behaviour and (b) possible implementation methods)

PR #1447, opened 2022-02-10
Do not silence error channel of panel.sh main loop

t-wissmann:
I can't remember why I activated the silencing of stderr in
dd6436ad451ba351 but now it is causing trouble because it hides the
error message of the textwidth utility.


t-wissmann:
> Looking at this from the perspective of a future lurker, i suggest mentioning `panel.sh` in the title, or it will soon be confusing.

Good catch, thanks!

blarz (approved):
Makes sense!

PR #1449, opened 2022-02-16
Improve portability of testsuite

madroach:
these changes help a lot of tests to pass on OpenBSD.

There are still many tests left, which fail on OpenBSD.

Most of them fail when asserting a return code of non-zero.

Some fail because they expect a very specific error message over which herbstluftwm has no control. For example OpenBSD `getopt` has slightly different error messages.

This is work in progress, but the changes proposed here already help a lot.


t-wissmann (dismissed):
Thanks! This PR looks very good! I've made a couple of simply additions and approved the PR:

I've added some comments in the source code for a reader of the source in a few years in the future :-)

Also, I've moved the `sleep` from the loop body to the loop condition such that the loop terminates if the sleep is killed and thus has non-zero exit status.

If you're fine with the changes I made, then you can simply remove the 'wip' label from the PR and then your PR will be merged automatically

madroach:
I made your comment more specific. Otherwise I'm fine with your changes.

t-wissmann (approved):

PR #1457, opened 2022-03-06
CSS-inspired theme engine for client decorations

### LINKED ISSUE #1433: CSS Style theming of frames and client decorations
Problem: The theming capabilities are quite restricted, and often it is required to adjust the source code for changing nuances of the behaviour, for example:

* Changing the drawing of inner/outer borders in a tabbed client is quite fixed. One can change the colors and width, but not where the lines are drawn
* `smart_window_surroundings` and `smart_frame_surroundings` hide gaps and borders, and sometimes, it might be desirable to hide only one of them but not the other, https://github.com/herbstluftwm/herbstluftwm/pull/1432/files
* It is impossible right now to change the border color of one particular client. If one would implement this with the current theme objects, what would this mean after all? Would one specify a different `color` for a client, or a `DecorationScheme` object (one set of decoration attributes) object or a `DecTriple` (3 sets, namely for `active`/`normal`/`urgent`)? Any such choice would make it either very rigid or lead to an explosion of attributes.

Proposed solution: A css-style theme options, similarly to what [rofi](https://github.com/davatorium/rofi) does. The `theme.` as it is now would still be a kind of default theme that can then be overwritten with css fine tuning.

t-wissmann:
We apply the CSS engine to the theming of clients. So instead of the
explicit client decoration drawing, we have a generic CSS-based widget
render and the existing theme attributes generate a stylesheet.
There is no documentation yet, but two example themes.

This closes #1433.


t-wissmann:
It's slowly progressing
css-tabs-blue-buttons

Issue #1448, opened 2022-02-12
CI shutdown is unreliable


t-wissmann:
Sometimes, we get CI failures because hlwm takes too long to shut down after the big test case:
```
2022-02-11T17:31:27.4937618Z [gw0] [ 92%] ERROR tests/test_completion.py::test_completable_commands[False] 
2022-02-11T17:31:27.4938179Z 
2022-02-11T17:31:27.4938435Z ==================================== ERRORS ====================================
2022-02-11T17:31:27.4938803Z ____________ ERROR at teardown of test_completable_commands[False] _____________
2022-02-11T17:31:27.4939671Z [gw0] linux -- Python 3.8.10 /__w/herbstluftwm/herbstluftwm/.tox/py38/bin/python
2022-02-11T17:31:27.4940026Z 
2022-02-11T17:31:27.4940352Z hlwm_spawner = <function hlwm_spawner.<locals>.spawn at 0x7ff6a07c0d30>
2022-02-11T17:31:27.4941047Z request = <SubRequest 'hlwm_process' for <Function test_completable_commands[False]>>
2022-02-11T17:31:27.4941482Z xvfb = <conftest.MultiscreenDisplay object at 0x7ff6a0761520>
2022-02-11T17:31:27.4941715Z 
2022-02-11T17:31:27.4941826Z     @pytest.fixture()
2022-02-11T17:31:27.4942118Z     def hlwm_process(hlwm_spawner, request, xvfb):
2022-02-11T17:31:27.4942421Z         parameters = {
2022-02-11T17:31:27.4942883Z             'transparency': True,
2022-02-11T17:31:27.4943137Z         }
2022-02-11T17:31:27.4943538Z         if hasattr(request, 'param'):
2022-02-11T17:31:27.4944208Z             # the param must contain a dictionary possibly
2022-02-11T17:31:27.4944538Z             # updating the default parameters
2022-02-11T17:31:27.4944958Z             parameters.update(request.param)
2022-02-11T17:31:27.4945344Z     
2022-02-11T17:31:27.4945744Z         """Set up hlwm and also check that it shuts down gently afterwards"""
2022-02-11T17:31:27.4946309Z         additional_commandline = ['--no-tag-import']
2022-02-11T17:31:27.4946817Z         if not parameters['transparency']:
2022-02-11T17:31:27.4947352Z             additional_commandline += ['--no-transparency']
[...]

PR #1458, opened 2022-03-07
Test suite: ping before shutdown

t-wissmann:
Right before terminating herbstluftwm in the test suite, verify that
herbstclient is still working. Maybe, this helps with solving #1448.


blarz (approved):
Let's see if it helps

PR #1459, opened 2022-03-12
Fix: Remove client from tab bars before unmanage

t-wissmann:
Before unmanaging and deleting a client, remove it from all tab bars of
the remaining clients. This fixes a crash triggered by the new test case
included in this change.


blarz (approved):
Looks good to me!

PR #1464, opened 2022-03-29
Extend 'smart_frame_surroundings' for more granular control

gravndal:
It now accepts 'off', 'hide_gaps', and 'hide_all'.

In addition to what was previously possible, this change enables only hiding frame gaps when the frame tree consists solely of the one root frame.

This change is not fully backwards compatible, see MIGRATION.


t-wissmann:
Thanks! (sorry again for messing up the original https://github.com/herbstluftwm/herbstluftwm/pull/1432)

t-wissmann (approved):
Looks good, thanks!

PR #1466, opened 2022-03-30
Update mailinglist links

The-Compiler:
See https://listi.jpberlin.de/pipermail/hlwm/2022-March/000342.html


t-wissmann (approved):
Thanks a lot for taking care of the mailing list!! :tada:

PR #1479, opened 2022-05-08
Move XKeyGrabber out of KeyManager

t-wissmann:
This makes KeyManager independent of X and avoids some usages of the
global variable `g_display`.

In the MouseManager, instead of removing the masks that we ignore, we
now just filter for the masks we are interested in.


blarz (approved):
Looks good to me!

t-wissmann:
Thanks!

PR #1480, opened 2022-05-09
New setting 'show_frame_decorations'

t-wissmann:
This new settings gives a more fine-grained control when frame decorations
are shown. The new setting strictly generalizes the always_show_frame
setting, which now only exists for backwards-compatibility.

The feature was requested in #1183, where it was reported that many
people find the initial frame decorations of a single big frame ugly
when starting hlwm for the first time. To address this, it maybe makes
sense to hide it per default. Thus, the setting is set to
`focused_if_multiple` in the in the default autostart.

Also, changing this default in settings.h should prevent new users
without an autostart from being confused by the red root frame (see
e.g. #599).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,8 @@ Current git version
     following new values: 'off', 'hide_all', and 'hide_gaps'. Setting it to
     'hide_gaps' will only hide frame gaps when applicable, 'hide_all' and 'off'
     mirror the old behaviour with regards to 'true' and 'false'.
+  * New setting 'show_frame_decorations' that control, when frame decorations
+    are shown. This should be moved instead of the old 'always_show_frame'
   * New frame attribute 'content_geometry'
   * New monitor attribute 'content_geometry'
 
--- a/NEWS
+++ b/NEWS
@@ -12,8 +12,8 @@ Current git version
     following new values: 'off', 'hide_all', and 'hide_gaps'. Setting it to
     'hide_gaps' will only hide frame gaps when applicable, 'hide_all' and 'off'
     mirror the old behaviour with regards to 'true' and 'false'.
-  * New setting 'show_frame_decorations' that control, when frame decorations
-    are shown. This should be moved instead of the old 'always_show_frame'
+  * New setting 'show_frame_decorations' that controls when frame decorations
+    are shown. This should be used instead of the old 'always_show_frame'
   * New frame attribute 'content_geometry'
   * New monitor attribute 'content_geometry'
 
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -32,7 +32,8 @@ template<>
 Finite<ShowFrameDecorations>::ValueList Finite<ShowFrameDecorations>::values = ValueListPlain {
     { ShowFrameDecorations::all, "all" },
     { ShowFrameDecorations::focused, "focused" },
-    { ShowFrameDecorations::focused_if_ambiguous, "focused_if_ambiguous" },
+    { ShowFrameDecorations::focused_if_multiple, "focused_if_multiple" },
+    { ShowFrameDecorations::if_multiple, "if_multiple" },
     { ShowFrameDecorations::nonempty, "nonempty" },
     { ShowFrameDecorations::none, "none" },
 };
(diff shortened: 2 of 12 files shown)


t-wissmann:
The name `focused_if_ambiguous` is a bit clumsy, maybe someone has a better suggestion than I.

tiosgz:
> 'nonempty' shows frame decorations of frames that have client windows

I wonder if it makes sense to add 'empty' for completeness? But at that point it would get even more complicated and a `theme.`-like system would be better imo; therefore, perhaps, not.

(I don’t need it, i’m just bikeshedding.)

t-wissmann:
> > 'nonempty' shows frame decorations of frames that have client windows
> 
> I wonder if it makes sense to add 'empty' for completeness?

I won't implement useless features just for completeness of terminology ;)

oredaze (commented):
Should I have commented here instead of at https://github.com/herbstluftwm/herbstluftwm/issues/1183 ?
Sorry, first time being asked to do a review.
...
Edit: So, if the following 4 cases are possible, then you can go ahead with the changes:

1. All frames are visible, always. (present behaviour)
2. All frames are visible, if number of frames are > 1. (new behaviour we want)
3. Only focused frames are visible, always. (present behaviour)
4. Only focused frames are visible, if number of frames are > 1. (new behaviour we want)

t-wissmann:
> Should I have commented here instead of at #1183 ? Sorry, first time being asked to do a review.

Either is fine for me, so lets continue here. (I've copied some quotes over from the issue discussion). Also, you don't need to go through the code. It is enough that you check that this PR implements the functionality that you suggested.

> The wording on this seems a bit off. To me "decorations" sounds like all decorations. I thought we were talking only about frame background, because the frame border disappearance is already a present functionality (with smart_frame_surroundings).

Yes, this is about all decorations: background and border. In this way, this feature is also usable for those who don't have `smart_frame_surroundings` activated (like me). Hiding only the background and still drawing the border would have been more complicated to implemented, and it would have more conflicts with compositing managers (like picom) when it comes to shadow drawing.

> So, if the following 4 cases are possible, then you can go ahead with the changes:
> 
> 1. All frames are visible, always. (present behaviour)
> 2. All frames are visible, if number of frames are > 1. (new behaviour we want)
> 3. Only focused frames are visible, always. (present behaviour)
> 4. Only focused frames are visible, if number of frames are > 1. (new behaviour we want)

> Are you sure the new changes make N.2 possible? I don't see how, based on the wording.

Good catch! I didn't think about this, but I will implement it.

> Also I don't see the point of "none" and "nonempty" even existing.

Over the years, there were requests to hide the frame decorations alltogether. In principle, one does not need frame decorations at all when 1. having frame_gap, 2. disabling window_gap, 3. always removing empty frames.

oredaze:
> Over the years, there were requests to hide the frame decorations alltogether.

I don't understand these people, but hey... I have heard you should develop for people, not for yourself. I guess there are odd usecases for everything.

Seems like we have sorted everything out.
As for the variable names, maybe these will give some ideas:
2. all_if_ambiguous / all_if_multiple / all_smart / all_adaptive
4. focused_if_ambiguous / focused_if_multiple / smart_focused / adaptive_focused

t-wissmann:
> > Over the years, there were requests to hide the frame decorations alltogether.
> 
> I don't understand these people, but hey... I have heard you should develop for people, not for yourself. I guess there are odd usecases for everything.

With above described gaps, it makes perfect sense to me without any decorations :-)

> Seems like we have sorted everything out. As for the variable names, maybe these will give some ideas: 2. all_if_ambiguous / all_if_multiple / all_smart / all_adaptive 4. focused_if_ambiguous / focused_if_multiple / smart_focused / adaptive_focused

Alright, I've renamed `focused_if_ambiguous` to `focused_if_multiple` and added the new `if_multiple`.

By the way, I'd avoid names like `smart`/`adaptive`/`auto`/`hybrid`/`mixture`/`strategic` etc because they give no clue what they actually mean, except that there is some `if ..else..` going on.

oredaze:
Great. 
Thank you for the tip as well.

> With above described gaps, it makes perfect sense to me without any decorations :-)

I guess I am the person that likes to have more than 1 empty frame. I have bound to keys some sequences of splits that create "layouts" of sorts. And then there is firefox, which must not have any frames or borders around it, else you won't be able to flick click it's tabs. That's some of the ways I use herbstluft. :)
Thank you for your time.

NEWS
```diff
@@ -12,6 +12,8 @@ Current git version
     following new values: 'off', 'hide_all', and 'hide_gaps'. Setting it to
     'hide_gaps' will only hide frame gaps when applicable, 'hide_all' and 'off'
     mirror the old behaviour with regards to 'true' and 'false'.
+  * New setting 'show_frame_decorations' that control, when frame decorations
```
blarz, line 15: controls?

NEWS
```diff
@@ -12,6 +12,8 @@ Current git version
...
+    are shown. This should be moved instead of the old 'always_show_frame'
```
blarz, line 16: should be used?

src/settings.cpp
```diff
@@ -124,15 +136,25 @@ Settings::Settings()
...
+    for (auto i : {&gapless_grid,
          &tabbed_max,
          &smart_window_surroundings,
          &raise_on_focus_temporarily}) {
         i->changed().connect(&all_monitors_apply_layout);
     }
+    show_frame_decorations.changed().connect(&all_monitors_apply_layout);
     smart_frame_surroundings.changed().connect(&all_monitors_apply_layout);
     wmname.changed().connect([]() { Ewmh::get().updateWmName(); });
+    // connect deprectaed attribute to new settings:
```
blarz, line 148: typo deprecated

blarz (commented):
Some typos, otherwise looks good to me!

NEWS
oredaze: `control,` should be `controls`. And yes `moved` doesn't make sense. I think you meant `used`.

t-wissmann:
Thanks! I've fixed the typos.

oredaze (approved):
Great. Thank you. 👍

blarz (approved):

PR #1481, opened 2022-05-17
Introduce %{ } format for sprintf

t-wissmann:
In foreach loops and other meta commands, one often needs to first
compose an attribute path and then look up its value. The present change
simplifies this to one step:

    sprintf S '%{%c.count}' tags echo S

will print the value of `tags.count`.


t-wissmann:
This is still work in progress, because I still need to
- [ ] add more tests
- [ ] add documentation (NEWS, doc/herbstluftwm.txt)
- [ ] possibly simplify existing scripts

PR #1493, opened 2022-06-28
Add 'at_end' attribute for tag order

t-wissmann:
This adds a new attribute that keeps tags near the end of the tag list.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,7 @@ Current git version
   * New setting 'show_frame_decorations' that controls when frame decorations
     are shown. This should be used instead of the old 'always_show_frame'
   * New frame attribute 'content_geometry'
+  * New tag attribute 'at_end'
   * New monitor attribute 'content_geometry'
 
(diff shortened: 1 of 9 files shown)


t-wissmann:
Things that are missing
- [ ] NEWS entry
- [ ] test cases

t-wissmann:
Thanks! I've fixed the minor issues.

I'm not sure how specific this feature is. I think it can be usable for all how dynamically add and remove tags while having something like the scratchpad or q3-style terminal.

blarz:
> Thanks! I've fixed the minor issues.
> 
> I'm not sure how specific this feature is. I think it can be usable for all how dynamically add and remove tags while having something like the scratchpad or q3-style terminal.

Thank you for the fixes!

I see, I guess it is more a personal opinion, as I don't use those two scripts at all. But for everybody that does, this makes sense

blarz (approved):

PR #1498, opened 2022-07-10
Add --bidirectional arg for keybind

t-wissmann:
If a keybind for a combo is added, the corresponding release bind is
removed automatically. So if e.g. some script binds 'Alt-x' and another
script binds 'Release-Alt-x', then not both keybindings are active.
Instead, the first Alt-x is removed when 'Release-Alt-x' is bound.

In the rare occassion where both press and release are bound, this new
'--bidirectional' needs to be used.

### DIFF
--- a/MIGRATION
+++ b/MIGRATION
@@ -15,6 +15,9 @@ now incompatible with the 'toggle' command. Use 'cycle_values' instead:
 Instead of 'always_show_frame', the new 'show_frame_decorations' setting should
 be used.
 
+In the 'keybind' command, if both the key press and the key release are bound,
+this need to be explicitly specified by passing '--bidirectional' to keybind.
+
 0.9.2 to 0.9.3
 --------------
 The command 'cycle_value' now expects an attribute path instead of a settings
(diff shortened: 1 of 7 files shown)


t-wissmann:
@The-Compiler I once mentioned this to you on IRC. What do think about this change?

The-Compiler:
I'm still not a fan of the idea - if one script binds press and the other binds release, I'd expect exactly that to happen. One of the bindings silently being removed by a different binding (from a totally different place perhaps) sounds unintuitive and like a nightmare to track down to me.

bew:
I agree with @The-Compiler, this looks very much unintuitive.. 
Do you have a spcific reason you'd like to have this behavior?

t-wissmann:
My use-case is only the one described above: I want to avoid that users accidentally have both press and release bound at the same time.

But if the two of you would rather not have this behaviour, I'm considering to just close this PR (without merging)

t-wissmann:
Since nobody seems to want it, lets close this PR

PR #1500, opened 2022-07-11
feat: add show_frame_decorations if_empty to show frame placeholders

mgutz:
Added `if_empty` option to `show_frame_directions`. This is useful to show a placeholder for frames which have no children. See the transparent dark regions in the screenshot.

frame-decorations-if-empty

### DIFF
--- a/src/framedecoration.cpp
+++ b/src/framedecoration.cpp
@@ -154,6 +154,9 @@ void FrameDecoration::updateVisibility(const FrameDecorationData& data, bool isF
     case ShowFrameDecorations::nonempty:
         show = data.hasClients;
         break;
+    case ShowFrameDecorations::if_empty:
+        show = !data.hasClients;
+        break;
     case ShowFrameDecorations::focused:
         show = data.hasClients || isFocused;
         break;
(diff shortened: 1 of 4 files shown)


t-wissmann:
Looks good, thanks! It's very helpful that you directly extended the test suite, too!
Is there any particular reason why you replaced the single quotes with double quotes in the python strings?

mgutz:
@t-wissmann Apologies. That was probably VS Code reformatting code. I ran `make flake8` before committing.  I thought that would enforce your style.

t-wissmann:
> Apologies. That was probably VS Code reformatting code. I ran `make flake8` before committing. I thought that would enforce your style.

No worries! I've just reverted the change in quote-style. Both `"` and `'` are totally fine, and so flake8 didn't complain. Here, I'd like to stick to the original single quotes, simply because it makes the diff smaller, so later it will be easier to understand the semantical change of the present commit/PR.

t-wissmann:
I've also removed the screenshot from the commit message, such that the git log does not depend on images from github.

Here's the extracted screenshot:

>  See the transparent dark regions in the screenshot.

frame-decorations-if-empty

t-wissmann (approved):
Thanks!

PR #1504, opened 2022-07-12
Make ewmhnotify behave as expected

t-wissmann:
The purpose of disabling `ewmhnotify` is to set the hlwm-fullscreen
state independent of the ewmh fullscreen state (either to have one
enabled and the other disabled, or vice versa). To make this more easy
to use, this de-coupling should happen in both directions.

The main purpose seems to make firefox think that it is in fullscreen
state (such that it hides the gui) without firefix covering the whole
monitor (#1488).

### DIFF
--- a/doc/herbstluftwm.txt
+++ b/doc/herbstluftwm.txt
@@ -1140,10 +1140,9 @@ Each 'CONSEQUENCE' consists of a 'NAME'='VALUE' pair. Valid 'NAMES' are:
     a boolean; it defaults to *on*.
 
 +ewmhnotify+::
-    sets whether hlwm should let the client know about EMWH changes (currently
-    only the fullscreen state). If this is set, applications do not change to
-    their fullscreen-mode while still being fullscreen. 'VALUE' is a boolean,
-    it defaults to *on*.
+    sets whether hlwm should let the client know about the EMWH fullscreen
+    state. If this is set, applications do not change to their fullscreen-mode
+    while still being fullscreen. 'VALUE' is a boolean, it defaults to *on*.
 
 +fullscreen+::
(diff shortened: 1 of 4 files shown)


t-wissmann:
I still need to
  - [ ] update NEWS
  - [ ] update MIGRATION
  - [ ] add test cases

t-wissmann:
@patrick96, @LucasTavaresA: do you have any opinion on this?

LucasTavaresA:
> @patrick96, @LucasTavaresA: do you have any opinion on this?

as i said before this would be good as a rule that specifically makes a window break those rules, changing the existing behavior does not look like a good idea

doc/herbstluftwm.txt
```diff
@@ -1140,10 +1140,9 @@ Each 'CONSEQUENCE' consists of a 'NAME'='VALUE' pair. Valid 'NAMES' are:
     a boolean; it defaults to *on*.
 
 +ewmhnotify+::
-    sets whether hlwm should let the client know about EMWH changes (currently
-    only the fullscreen state). If this is set, applications do not change to
-    their fullscreen-mode while still being fullscreen. 'VALUE' is a boolean,
-    it defaults to *on*.
+    sets whether hlwm should let the client know about the EMWH fullscreen
```
blarz, line 1143: EWMH

ypnos:
Great feature, and LGTM apart from the typo blarz pointed out.

t-wissmann:
Thanks, ypnos! I'll probably update the property names as mentioned above before merging this PR. I put it low on my priority list because the two people requesting it (in irc and in the github discussions) didn't react at all so far.

Issue #1296, opened 2021-04-27
Use an X-independent IPC protocol


tiosgz:
Let’s say I am hyperactive to annoying these days, yes… (& now I can also include naïve in the list)

Due to #79, communication between hlwm & hc should not use Xorg. I suppose UNIX socket might be the way to go (but when I suppose, I’m wrong).

I am opening a separate issue for this, because

- it is more related to de-X-ing than to Wayland itself & I don’t want to spam inboxes of the people who are watching there
- there’s no *urgent* need for this, it matters to me & it seems like an interesting opportunity to get across the border between only reading C/++ & writing C/++. That is, unless you have something against this idea in general or it being implemented by someone who has at most a few hundreds of written C behind him (which will require more work from your side as well before it can actually get into hlwm), I would like to try & code this. --> I want to make sure that I don’t implement it using a tool that you don’t want.

t-wissmann:
To be honest, I'd keep the X-specific IPC-protocol as it is, unless there are reasons against it. (There might be reasons, I'm only saying the Wayland protocol is no reason for not using X in the X11-case).

In general, implementing a protocol might be a wrong place to start writing C/C++ code. For the protocol, I've put much effort into making it compatible. Changing this now to using UNIX sockets can be done, but one has to be very careful and make decisions very consciously (e.g. to ensure that hlwm still works when doing a system upgrade: then you have a new herbstclient implementation but still a old hlwm server running)

tiosgz:
Probably it doesn’t make sense at this moment, since Wayland is far in the future… but as I was talking about the IPC & then looking at it, it just seemed to me like it will have to be done – earlier or later. One protocol for both makes more sense to me than two separate.

TIL that if the protocol is changed, one or two versions (depending on how fast you expect people to update) will probably have to support both ways. Good.

---

<details><summary>Now I’m getting more childish again, sorry. You can ignore me & wait if I get my sanity back soon.</summary>

It doesn’t matter to you, it doesn’t matter to hlwm, it doesn’t matter to anyone (if I exaggregate my emotions, I’m saying this with tears, *not with anger*). This is one thing that I wanted to try, that matters to me & that I can’t do in another language, plus it isn’t bound to a specific library like Gtk – the perfect conditions for me… Ne’er gonna learn it, *sowieso*. Stuck with ruby just because I can code in ruby & so any way to learn another language is too boring or too hard. Nevermind.

<details><summary>Is this more sane?</summary>

[...]

PR #1506, opened 2022-07-30
Add binary pipe interface to herbstclient

t-wissmann:
This implements a binary 'interactive' mode in herbstclient.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,12 @@ herbstluftwm NEWS -- History of user-visible changes
 See the link:MIGRATION[] file for changes that introduced incompatibility to
 older versions.
 
+Current git version
+-------------------
+
+  * New herbstclient mode *--binary-pipe* which allows to send multiple
+    commands through a single herbstclient instance.
+
 Release 0.9.5 on 2022-07-30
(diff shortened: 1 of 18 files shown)


t-wissmann:
There are still some race conditions in the ewmh tests, but e.g. the run time of `test_generate_completable_commands` reduces from 2 minutes to 5 seconds.

kriansa:
This is great! :tada: 

It will probably handle most usecases for https://github.com/herbstluftwm/herbstluftwm/issues/1296. I can also consider including a hlwm IPC on [wmcompanion](https://github.com/kriansa/wmcompanion) without having to rely on xcffib and X11.

Q: Beside commands, are you considering supporting hooks receival through the same "channel"? I know `hc --idle` exists, but then we would have to manage 2 processes in order to listen for hooks as well.

Thanks for the awesome work!

t-wissmann:
Thanks a lot for the feedback @kriansa. This is very motivating for me :-)


Yes, I was wondering whether it would help to send hooks through the same channel. Now that you mention it, I will definitely consider it (not yet part of this PR but afterwards).

PR #1524, opened 2022-09-26
Re-enable using double hyphens before rule options

### LINKED ISSUE #1487: Rule flags must allow a -- prefix
The man page and earler hlwm versions allowed the args to `rule` to be prefixed with dashes, e.g. `--once`, but currently, this does not work:
```
hc rule --once floating=on floatplacement=center
rule: No operator in given arg: once
```

cwolfe3:
Previously, double hyphens were stripped in the tokenizeArg function which is only called after checking if the argument is "prepend", "once", "printlabel", "not", "!", or "fixedwidth". This failed to match on "--once", "--prepend", etc..., so the argument is eventually passed to tokenizeArg. But tokenizeArg expects all of its arguments to contain a = or ~ (everything else was supposed to be handled already). This resulted in an error.

Now, the stripping of hyphens is done at the beginning of rule parsing, and after manually moving the loop pointer for "not" and "!".

This closes #1487

### DIFF
--- a/src/rulemanager.cpp
+++ b/src/rulemanager.cpp
@@ -32,6 +32,7 @@ int RuleManager::parseRule(Input input, Output output, Rule& rule, bool& prepend
 
     for (auto argIter = input.begin(); argIter != input.end(); argIter++) {
         auto arg = *argIter;
+        arg = stripDoubleHyphens(arg);
 
         // Whether this argument is negated (only applies to conditions)
         bool negated = false;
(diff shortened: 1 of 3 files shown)


cwolfe3:
It looks like I used std:: redundantly. Sorry about that; I'm not used to C++. I can add a new commit if needed.
Thanks for looking at my first ever pull request!

t-wissmann:
> It looks like I used std:: redundantly. Sorry about that; I'm not used to C++. I can add a new commit if needed. Thanks for looking at my first ever pull request!

No worries, this `std::...` policy is specific to hlwm's codebase and not general C++, so it is absolutely fine not to know it :-). I've removed one `std::` and adjusted test cases for the restored hyphen behaviour.

t-wissmann (approved):
Thanks!

PR #1537, opened 2022-12-15
Add 'decorated' rule consequence

pjones:
This adds the ability to change a client's decoration attribute from a rule consequence.

Currently a WIP to let others know what I'm up to.  I still need to write a test and documentation for this.


pjones:
@t-wissmann I would like to add documentation and tests to go along with this PR but I'm running into some troubles.

When `a2x` runs `patch-manpage-xml.py`, it turns around and runs `xsltproc`.  Despite having `xsltproc` in `PATH`, the python script fails with:

```
FileNotFoundError: [Errno 2] No such file or directory: 'xsltproc'
```

I can't figure out why `PATH` isn't being respected.  I don't know if CMake is doing something weird or if it's `a2x`.

Any suggestions?

t-wissmann:
Hi, thanks for your effort to also update documentation & tests :-)

I don't have a guess where the PATH might have been broken. Maybe we can find the issue via printf-debugging. Where is xsltproc? (`which xsltproc`). We can check whethe this directory is listed in the PATH of cmake and of the python script. To this end, can you add additional print-statements according to the following diff?
```diff
diff --git doc/CMakeLists.txt doc/CMakeLists.txt
index e30becbe..3e4f7750 100644
--- doc/CMakeLists.txt
+++ doc/CMakeLists.txt
@@ -81,6 +81,7 @@ function(gen_manpage sourcefile man_nr)
         "XMLLINT = '${CMAKE_CURRENT_SOURCE_DIR}/patch-manpage-xml.py'")
         add_custom_command(
             OUTPUT ${dst}
+            COMMAND echo "PATH in CMAKE: $ENV{PATH}"
             COMMAND ${ASCIIDOC_A2X}
                     --verbose
                     --conf-file=${custom_asciidoc_cfg}
diff --git doc/patch-manpage-xml.py doc/patch-manpage-xml.py
index 51394fe7..02fb7f41 100755
--- doc/patch-manpage-xml.py
+++ doc/patch-manpage-xml.py
@@ -17,6 +17,7 @@ for arg in sys.argv[1:]:
         filename = arg
         break

+print(f"PATH in the python script: \"{os.environ['PATH']}\"", file=sys.stderr)
 xsltproc = 'xsltproc'

 doc_directory = os.path.dirname(__file__)
```
After adding these statements, you should see their output when running `make clean && make` in the `doc/` subfolder of your `build` directory.

lag00n (approved):
LGTM, working as intended.

pjones:
@t-wissmann I made a tiny bit of progress.

It looks like if you use a configuration file for `a2x` then the default is to not pass on the environment to child processes.  In order to do that you have to add `ENV = None` which is odd.  So the CMake change that allows me to move forward is:

```diff
diff --git a/doc/CMakeLists.txt b/doc/CMakeLists.txt
index e30becbe..a95bf8fd 100644
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -78,7 +78,7 @@ function(gen_manpage sourcefile man_nr)
         # our script patching the man page xml instead of running xmllint
         set(custom_asciidoc_cfg "${CMAKE_CURRENT_BINARY_DIR}/a2x-patched-dont-install.conf")
         file(WRITE ${custom_asciidoc_cfg}
-        "XMLLINT = '${CMAKE_CURRENT_SOURCE_DIR}/patch-manpage-xml.py'")
+        "ENV = None\nXMLLINT = '${CMAKE_CURRENT_SOURCE_DIR}/patch-manpage-xml.py'")
         add_custom_command(
             OUTPUT ${dst}
             COMMAND ${ASCIIDOC_A2X}
```

However, once that is working I get this error:

```
error : Unknown IO error
warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl"
```

This appears to be coming from `xsltproc` because it can't deal with the fact that the above URL redirects to an HTTPS URL.

I feel like I'm missing a package that provides the XSL files that `xsltproc` is trying to load.  Any suggestions?

t-wissmann:
> This appears to be coming from `xsltproc` because it can't deal with the fact that the above URL redirects to an HTTPS URL.
> 
> I feel like I'm missing a package that provides the XSL files that `xsltproc` is trying to load. Any suggestions?

Here on archlinux, many xml/xsl files come in the package `docbook-xsl` which is a dependency of asciidoc.

But I suspect that it's not worth the effort to make asciidoc work on your system. For the changes in this PR, it's probably the easiest to disable the documentation on your machine. You can do so by running
```bash
cmake -D WITH_DOCUMENTATION=no ..
```
from your build directory. As soon as you push changes to `doc/herbstluftwm.txt` (and NEWS), I can check that the produced html/man documentation looks alright :-)

Issue #976, opened 2020-07-29
[feature request] sticky client window attribute


Suavesito-Olimpiada:
First of all, I love this project and I've used it since 2016, thanks for the hard work and amazing features!

Now to the feature. I like to see youtube videos using straw-viewer and mpv. When I'm doing multitasking I find myself moving around the mpv window (it is always in floating mode in the bottom-right corner) across the different tags that I'm jumping around. I would like to be able to *fix in position* a _floating windows_ across the different tags, without having focus in it. That way I would be able to fix a (in my case) video, or maybe a document across tags that I'm using.

IDK how it would work if there are different monitors, maybe it can be fixed in a monitor, independent of the tags that are in there. 

Thanks and stay safe. :smile:

t-wissmann:
> First of all, I love this project and I've used it since 2016, thanks for the hard work and amazing features!

Thanks!

The feature you are asking for is usually called "sticky" windows (e.g. the [ewmh standard](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46035376814976) calls them sticky). I think, 'sticky' windows make a lot of sense, so I hope hlwm will have it some day :)

Suavesito-Olimpiada:
Thanks for the fast response. :)

I first heard of them as sticky in bspwm reddit, nos that you tell so.

I would like to contribute to this as it is a feature that will save me a lot of time (hear a lot of music in YT while working). However I'm not familiar with the codebase, but with guidance -I think- I could handle the job. Is there a way to talk or get any recomendation in how to begin the work? (Sorry if my english is slippery)

t-wissmann:
Do you have experience with C++ programming? The implementation roadmap can roughly be the following:
* add a boolean attribute `sticky` to the `Client` object
* whenever the tag on a monitor changes, move all sticky clients from the old to the new tag (Maybe this will mess up the stacking order, but we shouldn't care too much for the moment)
* add documentation (`doc/herbstluftwm.txt`, `NEWS`)
* add test cases (a 'standard' case, and the corner cases, e.g. when the tags of two monitors are swapped; does it work for tiled clients when going from a tiling tag to a floating tag, or vice versa?)

Suavesito-Olimpiada:
I do have experience with C++ programming. Thanks for the way through, I'll begin working on it. :D

bklaase:
[...]

PR #1542, opened 2023-02-17
Add sticky window/client support

fabian-thomas:
This PR implements sticky windows/clients as discussed in #976.

Anything I use this feature for works without issues, but I'll test a few days/weeks for bugs.

Clients always stay on the monitor they are on. This is how bspwm handles sticky windows too. When tags are swapped between two monitors, the sticky windows are swapped too, so that they stay on the correct monitor. Sticky windows stay focused when switching tags, but do not gain focus if they were previously not foucsed. Rule support is added too and works.

### DIFF
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -100,9 +100,6 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
         updateEwmhState();
         hook_emit({"fullscreen", fullscreen_() ? "on" : "off", WindowID(window_).str()});
     });
-    sticky_.changed().connect([this] {
-        this->updateEwmhState();
-    });
     minimized_.changed().connect([this]() {
         static long long minimizedTick = 0;
         minimizedLastChange_ = minimizedTick++;
@@ -140,7 +137,7 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
                 "whether this client covers all other "
                 "windows and panels on its monitor.");
     sticky_.setDoc(
-                "whether this client is sticky. "
+                "whether this client is pinned to the monitor. "
                 "This means that the client stays on its monitor, even when "
                 "the monitor switches to another tag.");
     minimized_.setDoc(
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -101,7 +101,7 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
         hook_emit({"fullscreen", fullscreen_() ? "on" : "off", WindowID(window_).str()});
     });
     sticky_.changed().connect([this] {
-            // TODO
+        this->updateEwmhState();
     });
     minimized_.changed().connect([this]() {
         static long long minimizedTick = 0;
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -40,6 +40,7 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
     , urgent_(this, "urgent", false)
     , floating_(this,  "floating", false)
     , fullscreen_(this,  "fullscreen", false)
+    , sticky_(this,  "sticky", false)
     , minimized_(this,  "minimized", false)
     , floating_effectively_(this,  "floating_effectively", false)
     , title_(this,  "title", "")
@@ -75,6 +76,7 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
     ewmhnotify_.setWritable();
     ewmhrequests_.setWritable();
     fullscreen_.setWritable();
+    sticky_.setWritable();
     pseudotile_.setWritable();
     sizehints_floating_.setWritable();
     sizehints_tiling_.setWritable();
@@ -97,6 +99,9 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
     fullscreen_.changed().connect([this] {
         updateEwmhState();
         hook_emit({"fullscreen", fullscreen_() ? "on" : "off", WindowID(window_).str()});
+    });
+    sticky_.changed().connect([this] {
+            // TODO
     });
     minimized_.changed().connect([this]() {
         static long long minimizedTick = 0;
@@ -134,6 +139,9 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
     fullscreen_.setDoc(
                 "whether this client covers all other "
                 "windows and panels on its monitor.");
+    sticky_.setDoc(
+                "whether this client is sticky, "
+                "i.e., stays on the focused monitor.");
     minimized_.setDoc(
                 "whether this client is minimized (also called "
                 "iconified).");
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -140,8 +140,9 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
                 "whether this client covers all other "
                 "windows and panels on its monitor.");
     sticky_.setDoc(
-                "whether this client is sticky, "
-                "i.e., stays on the focused monitor.");
+                "whether this client is sticky. "
+                "This means that the client stays on its monitor, even when "
+                "the monitor switches to another tag.");
     minimized_.setDoc(
                 "whether this client is minimized (also called "
                 "iconified).");
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,7 @@ Current git version
     commands through a single herbstclient instance.
   * CSS-inspired theme engine (new attributes 'theme.name',
     'theme.style_override')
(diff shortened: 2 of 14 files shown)


t-wissmann:
> This PR implements sticky windows/clients as discussed in #976.
> Anything I use this feature for works without issues, but I'll test a few days/weeks for bugs.

Nice, thanks!

> Clients always stay on the monitor they are on. This is how bspwm handles sticky windows too. When tags are swapped between two monitors, the sticky windows are swapped too, so that they stay on the correct monitor. Sticky windows stay focused when switching tags, but do not gain focus if they were previously not foucsed. Rule support is added too and works.

I expect that switching the between tags moves the sticky floating clients always up in the stacking order. And sticky tiling clients are always moved to the focused frame right? But I think that's a reasonable price to pay. (Most of the time, sticky clients are floating and supposed to be on the top of the stacking order).

> * [ ]  add note to manpage/documentation

Here, you only need to add two things:

  * Mention the new 'sticky' rule in the rules section of the man page
  * Extend the description in `sticky_.setDoc()` a bit: "whether this client is sticky. This means that the client stays on its monitor, even when the monitor switches to another tag"

> * [ ]  add test cases

Beside the obvious things to test, it might be worth to test that setting 'sticky' and 'switchtag' in the rules works well (and in particular does not crash). If there are any issues here, it might be helpful to set the 'sticky' attribute as late as possible when applying the ClientChanges.

> * [ ]  review changed hook todo. Do we need to notify the xserver? Is sticky an attribute supported by xorg?

```C++
    sticky_.changed().connect([this] {
            // TODO
    });
```
I think you only need to call `updateEwmhState();` here because the value of `sticky` only has an effect when you switch to another tag.
In the Ewmh module, it would be good if `NET_WM_STATE` then contains the flag `_NET_WM_STATE_STICKY` for sticky clients (see [the ewmh doc](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm45962883349936)). But it's also fine to leave this for a later PR.

> * [ ]  fix pipelines if they fail (?)

Yes, this would be great :) This can be tricky sometimes, so feel free to ask in case of any trouble. Currently, there is one bug in the python-xlib in the CI, which non-deterministically
causes failure with some `RRMode` (RandR) exception. So don't be worried by
that, please :-)

fabian-thomas:
> I expect that switching the between tags moves the sticky floating clients always up in the stacking order. And sticky tiling clients are always moved to the focused frame right? But I think that's a reasonable price to pay. (Most of the time, sticky clients are floating and supposed to be on the top of the stacking order).

I'm not entirely sure about the code, but that is what I observed when testing it 😅. I copied some code from the merge_tag function. I especially don't understand the part about removeSlice and insertSlice. Is that needed? What does it do?

> Beside the obvious things to test, it might be worth to test that setting 'sticky' and 'switchtag' in the rules works well (and in particular does not crash). If there are any issues here, it might be helpful to set the 'sticky' attribute as late as possible when applying the ClientChanges.

I added a test case for `sticky` + `switchtag` and it seems to work without problems.

> Yes, this would be great :) This can be tricky sometimes, so feel free to ask in case of any trouble. Currently, there is one bug in the python-xlib in the CI, which non-deterministically causes failure with some `RRMode` (RandR) exception. So don't be worried by that, please :-)

The new tests pass locally. I'm not sure about the other cases. Especially `tests/test_rules.py::test_floatplacement_smart_create_many` crashed for me all the time. This may be bacause of what you described or because I use python3.10. Maybe the pipeline will tell us more.

t-wissmann:
Thanks!

> > I expect that switching the between tags moves the sticky floating clients always up in the stacking order. And sticky tiling clients are always moved to the focused frame right? But I think that's a reasonable price to pay. (Most of the time, sticky clients are floating and supposed to be on the top of the stacking order).
> 
> I'm not entirely sure about the code, but that is what I observed when testing it sweat_smile. I copied some code from the merge_tag function. I especially don't understand the part about removeSlice and insertSlice. Is that needed? What does it do?

The 'slices' are the entries in the stack. Every tag has it's own stack, so when moving windows to other tags, this also needs to be reflected in the stacks.

> > Beside the obvious things to test, it might be worth to test that setting 'sticky' and 'switchtag' in the rules works well (and in particular does not crash). If there are any issues here, it might be helpful to set the 'sticky' attribute as late as possible when applying the ClientChanges.
> 
> I added a test case for `sticky` + `switchtag` and it seems to work without problems.

Thanks!

> > Yes, this would be great :) This can be tricky sometimes, so feel free to ask in case of any trouble. Currently, there is one bug in the python-xlib in the CI, which non-deterministically causes failure with some `RRMode` (RandR) exception. So don't be worried by that, please :-)
> 
> The new tests pass locally. I'm not sure about the other cases. Especially `tests/test_rules.py::test_floatplacement_smart_create_many` crashed for me all the time. This may be bacause of what you described or because I use python3.10. Maybe the pipeline will tell us more.

Yes this was the bug in python-xlib. I worked around this by pinning python-xlib to an older version.

I'm now testing your PR locally. So far it works great. The only thing I noticed is that panel's don't detect hlwm's sticky state for some reason. I don't know yet what the issue is, because according to `xprop` the information is there. But I will look into it.

fabian-thomas:
I pushed a fix for an issue I noticed. When using sticky windows, the tag_status command returned wrong information.

Are there wrapper functions for what I added to support sticky windows? It seems weird to me that I needed to write a function that moves clients from one tag to another.

t-wissmann:
Hi, sorry for letting you wait!

> I pushed a fix for an issue I noticed. When using sticky windows, the tag_status command returned wrong information.

Thanks!
 
> Are there wrapper functions for what I added to support sticky windows? It seems weird to me that I needed to write a function that moves clients from one tag to another.

I've never needed this before because I never moved multiple windows around before. There is `TagManager::moveClient`, but I think it's not really appropriate here because it also re-layouts the involved monitor.

By the way, I have removed the ewmh-part of sticky because I've now realized that EWMH's 'sticky' property means something different (which only has an effect if you have a large virtual desktop through which the window manager scrolls).
The EWMH-option for the present sticky-feature does not have an own name. Instead the [`_NET_WM_DESKTOP`](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm45317634125936) property of the window is set to `0xFFFFFFFF` if the window is on all desktops. But this can be done in a separate PR together with test cases.

t-wissmann:
The only thing I'm wondering now is whether 'sticky' is a good name. bspwm and i3 call it 'sticky' and hlwm will probably never implement this virtual desktop scrolling, so hopefully we will never have a terminology clash :)

I'm sorry for giving you wrong pointers regarding EWMH!

fabian-thomas:
> The only thing I'm wondering now is whether 'sticky' is a good name. bspwm and i3 call it 'sticky' and hlwm will probably never implement this virtual desktop scrolling, so hopefully we will never have a terminology clash :)
> 
> I'm sorry for giving you wrong pointers regarding EWMH!

Hi, probably 'pinned' is a better name for the feature. But I would be in favor of just using 'sticky' in line with bspwm and i3. But the docs should probably mention 'pinned' in some way so that people find it if they are not familiar with the other term.

Are there any other things to do before merging this into main? I've used the changes for the last few weeks since I implemented it and never had an issue.

src/client.cpp
```diff
@@ -98,6 +100,9 @@ Client::Client(Window window, bool visible_already, ClientManager& cm)
         updateEwmhState();
         hook_emit({"fullscreen", fullscreen_() ? "on" : "off", WindowID(window_).str()});
     });
+    sticky_.changed().connect([this] {
```
fabian-thomas, line 103: this addition here can be removed, right?

t-wissmann: Indeed, yes!

t-wissmann:
> Hi, probably 'pinned' is a better name for the feature. But I would be in favor of just using 'sticky' in line with bspwm and i3. But the docs should probably mention 'pinned' in some way so that people find it if they are not familiar with the other term.

Makes sense, then lets stick to 'sticky'.

> Are there any other things to do before merging this into main? I've used the changes for the last few weeks since I implemented it and never had an issue.

I think the PR looks good! I've just changed some minor things (removed the sticky_.changed() callback, mention 'pinned' in the doc, and changed 'sticky state' -> 'sticky attribute' in the rules documentation). If that's fine with you, I'd say the PR is ready to be merged :)

fabian-thomas:
Yep, LGTM. :)

t-wissmann (approved):
Thanks for your work!

PR #1544, opened 2023-03-09
Stick to old python-xlib in CI until bugfix

t-wissmann:
In python-xlib 0.32, the bug https://github.com/python-xlib/python-xlib/pull/242 was introduced which caused `test_floatplacement_smart_create_many` to fail deterministically. There already is a pull request in `python-xlib` with a fix, but it is not merged yet. So for the time being, let's simply use the old python-xlib version.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,15 +24,14 @@ skipsdist = true
 # repository anyway)
 commands =
     pip install ./python/
-    # pin python-xlib to version 0.31 until the issue
-    # https://github.com/python-xlib/python-xlib/pull/242
-    # is fixed.
-    pip install --force-reinstall -v "python-xlib==0.31"
     {envpython} -m pytest {posargs}
 deps =
     ewmh
     pytest-xdist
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -24,6 +24,10 @@ skipsdist = true
 # repository anyway)
 commands =
     pip install ./python/
+    # pin python-xlib to version 0.31 until the issue
+    # https://github.com/python-xlib/python-xlib/pull/242
+    # is fixed.
+    pip install --force-reinstall -v "python-xlib==0.31"
```
blarz, line 30: Is it also possible to require the version in the `deps` line directly?

```
deps =
    [...]
    python-xlib==0.31
```

t-wissmann: Indeed, this seems to work too! I didn't dare to try it because I couldn't find an example online.

blarz: Nice :partying_face:

blarz (approved):

PR #1550, opened 2023-04-18
Move layout algorithms to own module

t-wissmann:
This change encapsulates the layout algorithms in their own module. The first benefit is that the long layout.cpp becomes shorter. The other benefit is that it will be easier to implement new layout algorithms (which is requested quite frequently).


t-wissmann:
This is still work in progress, I just wanted to see the result of the testsuite :-)

Issue #1003, opened 2020-09-11
Feature Request: Additional Layouts


rutrum:
There are currently 4 layouts: vertical, horizontal, max, and grid.  I think that there could be a lot variety in auto tiling within a frame.  Window manager dwm has many of these.  Two that stand out:
* Finonacci: First window takes 50% of the screen vertically, the next 25% horizontally, and so on.  See https://dwm.suckless.org/patches/fibonacci/
* Equal sized grid: Similar to grid.  When the current grid reaches 5 windows for example, the fifth is twice as large as the others.  This equally sized grid would scale all five to be about the same area on the screen.  That is, the bottom row might be shorter, and would be split in half for the last 2 of the 5 windows.

How feasible is adding additional layouts?

The-Compiler:
For the "equal sized grid" one, it sounds like you want the `gapless_grid` setting - from the manpage:

> This setting affects the size of the last client in a frame that is arranged by grid layout. If set, then the last client always fills the gap within this frame. If unset, then the last client has the same size as all other clients in this frame.

rutrum:
![This image](https://user-images.githubusercontent.com/23003873/92994855-276a3900-f4cc-11ea-9c1b-c284c7eab384.png) describes what I meant by "equally sized grid".

t-wissmann:
I understand what you want, and it sounds very reasonable. Since this is essentially the same as `grid`, I think it makes more sense to extend `gapless_grid` by a third option `stretched` than implementing a new layout. What do you think?

tiosgz:
There’s also a difference in height — but it can as well be a mistake…

t-wissmann:
> There’s also a difference in height — but it can as well be a mistake…

I assume that's just due to the drawing.

t-wissmann:
Do you prefer this as an additional layout (i.e. do you want it side by side with the current grid) or would should this be a setting configuring the behaviour of the existing grid layout?

noctuid:
[Here](https://github.com/HeavyRain266/hlwm-utils) are some more example layouts someone has implemented with manual tiling.

Ideally, there would be some user-configurable way of adding new automatic layouts like in other window managers. Bspwm allows an external rules script where you can query the current layout and then echo the split direction, percentage/ratio, and other rules. This as-is would probably not be a great fit for herbstluftwm since there is real automatic tiling that can resize/reposition windows other than the current one, but I imagine something similar (outputting w/h percentages and positions for windows) could be useful.

[...]

PR #1556, opened 2023-06-15
Add masterstack layout

fabian-thomas:
This PR adds the master stack layout as discussed in #1003.

TODOs:
- add docs
- check todo in code
- maybe add options for:
  - position of master node (left, right), currently its always right
  - which node should be focused in the stack when going right from master, currently the bottom node is focused
  - split ratio (currently its 0.6)

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ doxygen-html/
 # ignore various other IDE config dirs
 .vscode/
 .idea/
+
+.direnv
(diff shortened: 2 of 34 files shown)


fabian-thomas:
Didn't answer last year for some reason, sorry.

> I found the bottom one surprising when testing your code. Does the middle one make sense?

The reason I used the bottom one at that time was that it is the most recent node. Later I switched that to the top node because thats easier for your eyes. Middle would also be a valid choice, but I feel like then it's hard to see which one is the middle one (e.g. when there are 4 nodes in the stack). Best case would probably be remembering the node like it would be for the manual case, but I guess that complicates the code quite a bit.

I have used the current version of the code for the past year, so I would say functionality wise it's solid already. Maybe I can find some time to finish the PR if people are interested in adding the layout upstream. Otherwise I'll just keep using it like it currently is.

PR #1577, opened 2024-02-26
Restore non-transparent frame by removing mask

### LINKED ISSUE #1576: Previously transparent frames are rendered with incorrect size when `frame_bg_transparent=off`
#### Steps to reproduce:
```
Xephyr :1 -screen 1920x1080 -resizeable &
export DISPLAY=:1
herbstluftwm &
herbstclient set frame_bg_transparent off 
herbstclient split right 0.5 
xclock &
sleep 1
herbstclient resize left +0.05
herbstclient focus right
killall xclock
herbstclient remove
```
#### Observed behaviour:
There is only one frame (`herbstclient dump` prints `(clients max:0)`), but it is only filled about half.
#### Expected behaviour:
The frame should be filled fully.

Storchschnabel:
This fixes incorrect rendering of resized, once-transparent focused frames with `frame_bg_transparent off`.

`window_make_intransparent` only took into account the _current_ size. Properly removing the mask allows for later resizing.
Fixes #1576.

### DIFF
--- a/src/framedecoration.cpp
+++ b/src/framedecoration.cpp
@@ -131,7 +131,7 @@ void FrameDecoration::render(const FrameDecorationData& data, bool isFocused) {
         window_cut_rect_holes(xcon, window, rect.width, rect.height, holes);
         window_transparent = true;
     } else if (window_transparent) {
-        window_make_intransparent(xcon, window, rect.width, rect.height);
+        window_make_intransparent(xcon, window);
         window_transparent = false;
     }
     if (isFocused) {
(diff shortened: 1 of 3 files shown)


Storchschnabel:
@t-wissmann are you going to merge this? It works for me without issues.

t-wissmann:
Hi @Storchschnabel, sorry!
I think my original plan was to implement your issue description as a testcase and then take a look at your PR...  But then I simply forgot this PR :-(

Anyway, your changes look good, thanks a lot for the excellent description and the bugfix! I don't know why we had implemented it in such a difficult manner before. Also, I've implemented your issue description as a test case in #1584.

t-wissmann (approved):
Thanks :-)

PR #1584, opened 2024-06-07
Reproduce issue #1576 as a test case

t-wissmann:
This reproduces issue #1576 as a test case. The issue is that the frame background masks were not correctly reset; the issue was reported and fixed by @Storchschnabel.

Since the issue can be reproduced rather simply, this commit adds it as a test case.

### DIFF
--- a/tests/test_frame_decorations.py
+++ b/tests/test_frame_decorations.py
@@ -44,7 +44,7 @@ def test_frame_bg_transparency_area_bug(hlwm, x11, mouse):
     hlwm.call(['set', 'frame_bg_transparent', 'off'])
     hlwm.call(['set', 'always_show_frame', 'on'])
     # put two frames side by side:
-    hlwm.call(['split', 'right', 0.5 ])
+    hlwm.call(['split', 'right', 0.5])
     # put the mouse cursor on the right-hand frame
     screen_size = hlwm.attr.monitors[0].geometry()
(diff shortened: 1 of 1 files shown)


t-wissmann:
Let us double check that the test case fails for the old master and runs through after 'Update branch' :-)

t-wissmann:
Strangly, the test case does not fail in the CI:
```
tests/test_frame_decorations.py::test_frame_bg_transparency_area_bug 
[gw2] [  3%] PASSED tests/test_ewmh.py::test_net_current_desktop_after_tag_index_change[0-2-2] 
tests/test_ewmh.py::test_net_current_desktop_after_tag_index_change[0-2-3] 
[gw3] [  3%] PASSED tests/test_frame_decorations.py::test_frame_bg_transparency_area_bug 
```

But it fails locally for me:
```
        # now, the formerly left hand frame should span
        # the whole screen. So lets see which window is under the
        # mouse cursor:
        window = x11.get_window_under_cursor()
        # it should not be the root window
>       assert window != x11.root
E       assert <Window 0x000003eb> != <Window 0x000003eb>
E        +  where <Window 0x000003eb> = <conftest.X11 object at 0x748645ec3f10>.root
```

I don't know where the different behaviour comes from. Maybe the window shape extensions are not taken into account by the X version in the CI... But as long as the test still passes with the freshly merged PR #1577, I think it still makes sense to include the new test.

PR #1588, opened 2024-07-31
Expand 'smart_window_surroundings' functionality

### LINKED ISSUE #1530: option to make `smart_window_surroundings` only hide if there is only one window on a tag
I often find myself in the situation that I have frames with just one window because I want different sizes for them (think of a small terminal under your text editor). The only way to resize windows independently is to use frames.

The problem is, that I find it unacceptable to have just one window on my monitor AND window borders as they serve no purpose. This is what the `smart_window_surroundings` solves, although it also hides window borders in a frame with just one window.

Keeping the frame borders always on or making them look the same as the active window border is not an option since then you would get double borders, once for the frame and a second time for a window in case you have more than one window in a farme.

I was able to produce the desired effect in a fairly simple bash script, but IMO this is so fundamental, it should be an option in the WM itself.

This could be implemented much like the recent `show_frame_decorations` option with multiple values indicating different desired behaviours. Something like
- `off` for  the current `smart_window_surroundings off`
- `tag` for when the borders should be hidden when there is only one window in the current tag
- `frame` for when the borders should be hidden when there is only one window in a frame (the current `smart_window_surroundings on`)

The bash script I was talking about:
```bash
#!/usr/bin/env bash

herbstclient watch tags.focus.client_count

herbstclient --idle attribute_changed tags.focus.client_count "\d*" 1 | while read event; do
  herbstclient set smart_window_surroundings on
  # echo "toggeling window surroundings off"
done &

herbstclient --idle attribute_changed tags.focus.client_count "\d*" "[^1]" | while read event; do
  herbstclient set smart_window_surroundings off
  # echo "toggeling window surroundings on"
done &
```

bileslav:
Similarly to the changes in `smart_frame_surroundings`, this change makes
`smart_window_surroundings` more fine-grain.

This closes #1530

### DIFF
--- a/MIGRATION
+++ b/MIGRATION
@@ -15,14 +15,14 @@ put
 
 in your autostart.
 
-The command 'smart_window_surroundings' is no longer a boolean, and as such it is
+The setting 'smart_window_surroundings' is no longer a boolean, and as such it is
 now incompatible with the 'toggle' command. Use 'cycle_values' instead:
 
     hc cycle_value settings.smart_window_surroundings off one_window_and_frame one_window
 
 0.9.4 to 0.9.5
 --------------
-The command 'smart_frame_surroundings' is no longer a boolean, and as such it is
+The setting 'smart_frame_surroundings' is no longer a boolean, and as such it is
 now incompatible with the 'toggle' command. Use 'cycle_values' instead:
 
     hc cycle_value settings.smart_frame_surroundings off hide_gaps hide_all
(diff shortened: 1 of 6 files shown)


t-wissmann:
Thanks! I've rephrased the documentation and the condition a bit. Please check that the changes are fine to be merged under your name.

bileslav:
@t-wissmann, these are great changes; I think it's ready for merging. Thank you very much!

t-wissmann (approved):
Thanks! I've also added test cases in a separate PR #1593, so the present PR is ready to be merged :-)

PR #1589, opened 2024-08-01
CI: Remove custom docker images and unify main build jobs

t-wissmann:
There is quite some redundancy in the two main build jobs. For easier
maintenance, they are merged into one job running on a matrix strategy.

While at it, I have removed the need for our custom docker images from
https://hub.docker.com/r/hlwm/ci and have switched to github's standard
ubuntu system and to a standard ubuntu docker image for the ancient 32 bit
build job. This simplifies the whole ci process (and its maintenance).

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -46,7 +46,7 @@ jobs:
       # Checks out a copy of your repository on the ubuntu-latest machine
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: install dependencies
+      - name: Install dependencies
         uses: awalsh128/cache-apt-pkgs-action@latest
         with:
           packages: ccache clang clang-tidy curl cmake g++ git iwyu lcov libx11-dev libxext-dev libxft-dev libfreetype-dev libxinerama-dev libxrandr-dev libxfixes-dev ninja-build pkg-config python tox wget x11-xserver-utils xdotool xterm xvfb xserver-xephyr
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -52,7 +52,7 @@ jobs:
           packages: ccache clang clang-tidy curl cmake g++ git iwyu lcov libx11-dev libxext-dev libxft-dev libfreetype-dev libxinerama-dev libxrandr-dev libxfixes-dev ninja-build pkg-config python tox wget x11-xserver-utils xdotool xterm xvfb xserver-xephyr
           version: 1.0
 
-      - uses: actions/cache@v1
+      - uses: actions/cache@v3
         name: Cache ~/.ccache
         with:
           path: ~/.ccache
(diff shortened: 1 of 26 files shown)


.github/workflows/hlwm.yml
```diff
@@ -171,7 +163,7 @@ jobs:
           CFLAGS: -m32
         run: |
           cd herbstluftwm-*/
-          find ~/.ccache | wc -l
+          test -f ~/.ccache && find ~/.ccache | wc -l
```
blarz, line 166: Shouldn't this be `-d`?

src/cssname.h
```diff
@@ -50,7 +50,6 @@ class CssName {
         index_ = static_cast<size_t>(builtin);
     }
     CssName(const std::string& name);
-    CssName(const CssName& other) = default;
```
blarz, line 53: What happened here?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py38
+envlist = py310
 skipsdist = true
```
blarz, line 3: We could update the flake8 version while here

src/cssname.h
t-wissmann: I saw it in one of the changes made in your PR and then one of the compiler errors suggested to remove the default constructor here (and fortunately, we do not seem to need it).

tox.ini
t-wissmann: Can you suggest  version? I just made this change to mke the CI pass again.

.github/workflows/hlwm.yml
t-wissmann: Yes, thanks!

tox.ini
blarz: 7.0.0 (the latest) works locally here

But you need to move the comments in the ignore section *above* the rule numbers.

```
[testenv:flake8]
deps =
    flake8 == 3.8.3
commands = flake8 .

[flake8]
ignore =
    # line too long
    E501,
    # incompatible with pep8.. "Line break occurred before a binary operator"
    W503,
```

src/cssname.h
blarz: Oh I see, then let's kill it :)

blarz:
@t-wissmann what is the current state of this? Should somebody take over?

We also discussed that maybe removing ancient gcc support (seeing that we first should fix brand new gcc support :smile:)

t-wissmann:
@blarz Thanks for the reminder! The CI is in a really undesirable state at the moment. I tried to make the present PR #1589 working, but it might be more effort, because the current ubuntu image has a much stricter gcc. So to get the other waiting PRs merged, I've created the PR #1614 that still uses hlwm's docker images and does as little as necessary to let the CI pass again.

t-wissmann:
Fixed by #1655

PR #1600, opened 2024-09-20
python: Use pathlib instead of os.path

The-Compiler:
Use [pathlib](https://docs.python.org/3/library/pathlib.html) instead of `os.path`, which exposes file paths as objects, similar to what `tmpdir` does.

Also replace `tmpdir` by the [more modern `tmp_path`](https://docs.pytest.org/en/stable/how-to/tmp_path.html) which uses the pathlib API instead of the (somewhat pytest specific) `py.path.local` one.

Got a bit bigger than I thought - let me know if you'd rather not have the churn or only have the `tmpdir` -> `tmp_path` part, no hard feelings in that case.

### DIFF
--- a/tests/test_herbstclient.py
+++ b/tests/test_herbstclient.py
@@ -7,10 +7,10 @@ import sys
 import struct
 import contextlib
 from herbstluftwm import Herbstluftwm
-from conftest import PROCESS_SHUTDOWN_TIME, HcIdle
+from conftest import BINDIR, PROCESS_SHUTDOWN_TIME, HcIdle
 from Xlib import X, Xatom
 
-HC_PATH = os.path.join(os.path.abspath(os.environ['PWD']), 'herbstclient')
+HC_PATH = BINDIR / 'herbstclient'
 
 
 @pytest.mark.parametrize('argument', ['version', '--idle'])
--- a/tests/test_python_binds.py
+++ b/tests/test_python_binds.py
@@ -1,17 +1,17 @@
 import pytest
-import os.path
 import subprocess
 import sys
 import os
 import conftest
+import pathlib
 from herbstluftwm.types import Rectangle
 
 
 def test_example(hlwm):
     # test the example.py shipped with the bindings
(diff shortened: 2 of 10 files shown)


tests/test_python_binds.py
```diff
@@ -1,17 +1,17 @@
...
 from herbstluftwm.types import Rectangle
 
 
 def test_example(hlwm):
     # test the example.py shipped with the bindings
-    example_py = os.path.join(os.path.dirname(__file__), '..', 'python', 'example.py')
+    example_py = pathlib.Path(__file__).parents[1] / 'python' / 'example.py'
     # make 'herbstclient' binary available in the PATH
-    os.environ['PATH'] = conftest.BINDIR + ':' + os.environ['PATH']
+    os.environ['PATH'] = str(conftest.BINDIR) + os.pathsep + os.environ['PATH']
```
ypnos, line 14:
```suggestion
    os.environ['PATH'] = f"{conftest.BINDIR}{os.pathsep}{os.environ['PATH']}"
```

ypnos (dismissed):
The typical import behavior for pathlib is "from pathlib import Path" and use Path() directly in the code.

But this might also be a question of preference.

tests/test_herbstclient.py
```diff
@@ -7,10 +7,10 @@
 import struct
 import contextlib
 from herbstluftwm import Herbstluftwm
-from conftest import PROCESS_SHUTDOWN_TIME, HcIdle
+from conftest import BINDIR, PROCESS_SHUTDOWN_TIME, HcIdle
 from Xlib import X, Xatom
 
-HC_PATH = os.path.join(os.path.abspath(os.environ['PWD']), 'herbstclient')
```
blarz, line 13: import os can be removed now

blarz:
Call me the PR necromancer :sweat_smile:

t-wissmann:
Thanks, @The-Compiler! Also, @ypnos made a suggestion which sounds reasonable to me :-)

blarz (approved):
Thank you for this! Sorry for the delay :sweat_smile:

PR #1601, opened 2024-09-21
Make tag selection strategy for new monitors configurable

The-Compiler:
Add a new `monitors.tag_selection_strategy` attribute. Depending on what it is set to, when a new monitor is added, HLWM will either:

- Pick the first tag that's currently not shown (default and existing behavior)
- Prefer the first tag that's empty (no clients) instead
- Enforce that only an empty tag is shown


t-wissmann (approved):
Sorry for letting you wait so long..

Everything is perfect: Removal of old functions, Updated Documentation, multiple test cases :-)

Issue #1612, opened 2025-02-19
Failure to build with G++ version 15.0


sham1:
Currently on Fedora Rawhide and 42, building Herbstluftwm 0.9.5 fails due to the following error within `xconnection.cpp`:

```
/builddir/build/BUILD/herbstluftwm-0.9.5-build/herbstluftwm-0.9.5/src/xconnection.cpp: In function ‘std::string iso_8859_1_to_utf8(const char*)’:
/builddir/build/BUILD/herbstluftwm-0.9.5-build/herbstluftwm-0.9.5/src/xconnection.cpp:295:9: error: ‘uint8_t’ was not declared in this scope
  295 |         uint8_t ch = source[i];
      |         ^~~~~~~
/builddir/build/BUILD/herbstluftwm-0.9.5-build/herbstluftwm-0.9.5/src/xconnection.cpp:15:1: note: ‘uint8_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
   14 | #include "globals.h"
  +++ |+#include <cstdint>
   15 | 
/builddir/build/BUILD/herbstluftwm-0.9.5-build/herbstluftwm-0.9.5/src/xconnection.cpp:296:13: error: ‘ch’ was not declared in this scope
  296 |         if (ch < 0x80) {
      |             ^~
gmake[2]: *** [src/CMakeFiles/herbstluftwm.dir/build.make:894: src/CMakeFiles/herbstluftwm.dir/xconnection.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[2]: Leaving directory '/builddir/build/BUILD/herbstluftwm-0.9.5-build/herbstluftwm-0.9.5/redhat-linux-build'
gmake[1]: *** [CMakeFiles/Makefile2:228: src/CMakeFiles/herbstluftwm.dir/all] Error 2
gmake[1]: Leaving directory '/builddir/build/BUILD/herbstluftwm-0.9.5-build/herbstluftwm-0.9.5/redhat-linux-build'
gmake: *** [Makefile:139: all] Error 2
```

Simple fix is to add `cstdint` onto the file, as suggested by the error message. I'm just wondering if it should be included within `xconnection.cpp` or some kind of a header file instead.

PR #1613, opened 2025-03-10
Add #include <stdint.h> to fix building with gcc 15

kostadinsh:
With gcc 15, the C++ Standard Library no longer includes other headers that were internally used by the library. In herbstluftwm's case the missing header is `<stdint.h>`

Downstream Gentoo bug: https://bugs.gentoo.org/937529

Closes: #1612

### DIFF
--- a/src/xconnection.cpp
+++ b/src/xconnection.cpp
@@ -6,6 +6,7 @@
 #include <X11/Xutil.h>
 #include <X11/extensions/Xrender.h>
 #include <fcntl.h>
+#include <stdint.h>
 #include <unistd.h>
 #include <climits>
 #include <cstring>


The-Compiler (approved):

blarz:
Thank you for the fix @kostadinsh :rocket: 

The CI was broken for a while until last night. Now that @t-wissmann fixed it, can you check the output. The include what you use linter is complaining about the new include, but it just needs to be moved a few lines up and then the fix looks good to me!

kostadinsh:
My git clone got messed up somehow, pushed the change sorting the header alphabetically.

t-wissmann:
@mergifyio refresh

PR #1614, opened 2025-03-12
CI: remove old 32 bit job

t-wissmann:
It does not seem to be possible to let github actions pass data into old docker containers. So let us disable the old 32 bit build for the time being.
Also, this commit updates the version of the github action modules and installs imagemagick and librsvg2-bin in the website job to make the CI pass again.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -286,6 +286,6 @@ jobs:
         run: |
           builddir=$BUILD ci/mktar.sh
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
           path: herbstluftwm*.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/hlwm.yml
```diff
@@ -132,71 +132,71 @@ jobs:
...
-      - name: Install
-        env:
-          DESTDIR: ${{ github.workspace }}/install-herbstluftwm/
-        run: |
-          cd herbstluftwm-*/build/
-          mkdir -v -p $DESTDIR
-          ninja -v install
-          # check that the man page has been installed:
-          find $DESTDIR -name 'herbstluftwm.1' -printf '%P\n' | grep . # grep for the right exit-status
+  # build-old-32bit:
```
blarz, line 135: Should we remove this instead of commenting it out?

blarz: Thank you!

blarz (approved):
Let's remove the blocking job and fix newer gcc versions first :)

PR #1618, opened 2025-05-10
Prevent mouse position restore on monitor change

codefaux:
First, I believe I've accomplished this with the following goals in mind;
- Optional, default behavior completely unchanged
- Minimal footprint, implemented properly (I believe) and conforming to existing scheme/style
- Clear intent, no unintended impacts

I really dislike when things move my mouse. I expect things to be where I left them in general, especially on my computer. In looking for an option to resolve this, I found (a few) others discussing similar feelings, so I decided to try my hand at implementing the feature instead of seeking another WM.

There is only one time when HerbstluftWM calls `XWarpPointer` -- during `monitor_focus_by_index`

The existing choice given by configuration options present when activating a monitor (ie via keybind) is "Do you want your mouse moved where it was before you switched away, or recentered?"

My intent is that the only behavioral change is to avoid moving the mouse upon changing focus between monitors, and only if the user desires it. This gives a third option; "Should HerbstluftWM move your mouse at all?"


t-wissmann:
Hi,

it makes perfect sense to add such an option! I also like the minimal footprint! A few comments:

* Every boolean setting should have a positive name (so one should not use something like `not` in settings names). In your case: something along `monitors_move_cursor` which is `true` per default.
* Instead of a setting, it would be even easier to have an attribute `monitors.move_cursor` instead. Code-wise, this amounts to
```
    Attribute_<bool> move_cursor;
```
in `monitormanager.h` and
```
    , move_cursor  (this, "move_cursor", true)
```
In the initializer list of `MonitorManager::MonitorManager()` in `monitormanager.cpp`.
You can use this boolean then via `g_monitors->move_cursor()` in `monitor_focus_by_idx()`.

* Please also mention the new attribute in the `NEWS` file
* Consider adding a small test case. This requires a python-function in `test/test_monitor.py` and one needs to extend `def mouse()` in `conftest.py` by a function that reports the mouse location (internally calling `xdotool getmouselocation`).

codefaux:
> Hi,
> 
> it makes perfect sense to add such an option! I also like the minimal footprint! A few comments:
> 
>     * Every boolean setting should have a positive name (so one should not use something like `not` in settings names). In your case: something along `monitors_move_cursor` which is `true` per default.
> 
>     * Instead of a setting, it would be even easier to have an attribute `monitors.move_cursor` instead. Code-wise, this amounts to
> 
> 
> ```
>     Attribute_<bool> move_cursor;
> ```
> 
> in `monitormanager.h` and
> 
> ```
>     , move_cursor  (this, "move_cursor", true)
> ```
> 
> In the initializer list of `MonitorManager::MonitorManager()` in `monitormanager.cpp`. You can use this boolean then via `g_monitors->move_cursor()` in `monitor_focus_by_idx()`.
> 
>     * Please also mention the new attribute in the `NEWS` file
> 
>     * Consider adding a small test case. This requires a python-function in `test/test_monitor.py` and one needs to extend `def mouse()` in `conftest.py` by a function that reports the mouse location (internally calling `xdotool getmouselocation`).

Thanks for the positivity. I'm glad it makes sense to more ears than just mine.

Respectfully;
- I agree with the positive-naming convention overall, however I couldn't come up with anything better at the time. Naming it `monitors_move_cursor` implies that the monitors are moving things and that doesn't feel better. I'd prefer to have 'mouse' first since it associates well with `mouse_recenter_gap` when you're tab-completing or reading documentation. Maybe `mouse_pos_restore` default to true? I really feel like `no_move_mouse` is both concise and autological; "no move mouse true|on|1|enabled" means it won't move mouse, but I'm not stuck to it.
- From a user POV rather than a developer; When I add/remove monitors, will I need to update this setting again if it's a monitors.* attribute? I don't think most users will know this, and it would increase the buden on documenting the feature. Both ambiguity and workload sound negative. Also from my understanding, this would be per-monitor? The only other per-monitor settings are geometry-/tag-related, so it feels very out of place. I disagree increasingly with this suggestion.

- In which way is it "even easier" to change to a different thing which seems less predictable/expected? Isn't "do/don't move my mouse" more of a global WM-level "setting" than a monitor-level attribute, similar to "move my mouse to center when the gap is X amount or more"? __What__ makes it easier?

- I feel it makes more sense to have the setting which overrides `mouse_recenter_gap` in the same place _as_ `mouse_recenter_gap`. If this setting should be a monitor.* attribute, why isn't `mouse_recenter_gap` also a monitor.* attribute?

I had wondered if I should put a change note someplace, but I was less confident I was meant to. I'll update NEWS with the next changes, after discussion above.

I don't understand writing tests. I've never approached them, as a concept. This will be my first attempt. I'll certainly hold off on this until further discussion, but I'm willing to do the footwork of learning it.

Let me know what you think of my feedback, and thank you again.

t-wissmann:
Hey, sorry that I couldn't reply yet. Feel free to remind me if I don't react within one week.
(Just a quick thought: It also makes sense to put the setting under `mouse.`, e.g. `mouse.follows_monitor_focus`).

codefaux:
> Hey, sorry that I couldn't reply yet. Feel free to remind me if I don't react within one week. 

No worries; since I wrote it, I also already am running it, thus my impetus is greatly reduced, lol

>(Just a quick thought: It also makes sense to put the setting under `mouse.`, e.g. `mouse.follows_monitor_focus`).

_That_ I like. It reads well, it's more clear than my original name. Also when looking for mouse.* things in [herbstluftwm(1)](https://herbstluftwm.org/herbstluftwm.html) I just found 'focus_follows_mouse' -- 'mouse_follows_monitor_focus` could be another option, but that again goes back to set vs set_attr and I'm a but unclear.

Off topic; What's the effective difference to use a `set_attr` style value instead of a `set` style value? I don't have a single `set_attr` in my autostart. I wouldn't think to look there, either. Am I missing significant functionality? I guess there are no other things to `set_attr` in `mouse` domain? I've only been using hlwm on my daily driver for like a week, forgive my ignorance.

I'll write it up under `mouse.follows_monitor_focus` but I would be interested in understanding the reasoning, especially if it's the only thing there.

ETA weekend or sooner, but I understand there's no hurry.

codefaux:
Actually scratch that, I can't integrate that into `mouse`.

I don't see that it would be an easy enough change for me to tackle, and I'm not confident I understand the scope of making it.

There doesn't appear to be a `mouse` class to extend with attributes as is present in Monitor and Settings.

I tried looking into the `set_attr` command's autocomplete probe to understand why/how it populates and derive where `mouse` comes from in that context, but it's a bit more complex than I can dig out.

I'll rename what I've got to mouse_follows_monitor_focus. It's a bit long, but it's exactly the same number of characters in the option suggested, with one substitution. I'll keep it as a Setting attribute -- for now at least.

Let me know what you'd like changed when you have time, no pressure.

t-wissmann:
Hey, sorry for the delay. The PR's description is used as the commit message, so I'm about to update it. I quote the original PR description below:

> Hi there, thanks for your time. hlwm is great, and I find myself using it on increasingly more systems.
> 
> First, I believe I've accomplished this with the following goals in mind;
> 
> * Optional, default behavior completely unchanged
> * Minimal footprint, implemented properly (I believe) and conforming to existing scheme/style
> * Clear intent, no unintended impacts
> 
> I really dislike when things move my mouse. I expect things to be where I left them in general, especially on my computer. In looking for an option to resolve this, I found (a few) others discussing similar feelings, so I decided to try my hand at implementing the feature instead of seeking another WM.
> 
> There is only one time when HerbstluftWM calls `XWarpPointer` -- during `monitor_focus_by_index`
> 
> The existing choice given by configuration options present when activating a monitor (ie via keybind) is "Do you want your mouse moved where it was before you switched away, or recentered?"
> 
> My intent is that the only behavioral change is to avoid moving the mouse upon changing focus between monitors, and only if the user desires it. This gives a third option; "Should HerbstluftWM move your mouse at all?"

t-wissmann:
> Actually scratch that, I can't integrate that into mouse.

I have chatted with @blarz about this. The best option is to fix minor issues in this PR, merge it, and then I can move the setting to `mouse.` in a separate PR.

t-wissmann (dismissed):
LGTM,

Co-Approved by @blarz

t-wissmann (approved):
LGTM,

Co-Approved by @blarz

PR #1621, opened 2025-06-12
Add tab reordering in max layout with shift command

0x17de:
This change adds a new `max_tab_reorder` setting that allows users to reorder tabs within max layout frames using $mod+shift+direction hotkeys.

Tested this extensively in real-world usage, since last week - no issues encountered.


0x17de:
Hey @t-wissmann , thanks for your update! Sounds good - happy to close this MR as this resolves my request

t-wissmann:
Thanks! I'm glad that this solves your use-case :-)

Issue #1629, opened 2025-08-03
pavucontrol spams _NET_FRAME_EXTENTS and WM_NORMAL_HINTS


t-wissmann:
When starting `pavucontrol` in hlwm, an endless queue of `PropertyNotify` events is triggered regarding the atom `_NET_FRAME_EXTENTS` and `WM_NORMAL_HINTS`. There are more than thousand events within seconds, causing both hlwm and all applications to freeze and become unusable.

PR #1634, opened 2025-09-21
Only redraw on new size hints if geometry changes

t-wissmann:
If a client changes its size hints via the property WM_NORMAL_HINTS, then we should only redraw the client if the new hints would change the size of the client.

This fixes bug #1629 in which pavucontrol endlessly sending new WM_NORMAL_HINTS property notify events, causing hlwm to redraw for no reason but causing even more property notify events.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,6 +18,8 @@ Current git version
     window surroundings when there is only one window and one frame in a tag,
     'one_window' and 'off' mirror the old behaviour with regards to 'true' and
     'false'.
+  * Bug fix: Only redraw after new WM_NORMAL_HINTS if it would affect the
+    clients geometry.
 
 Release 0.9.5 on 2022-07-30
 ---------------------------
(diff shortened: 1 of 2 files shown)


t-wissmann:
@The-Compiler You were also affected by this issue, weren't you? Can you check whether the branch `lazyhints` helps?

In my cause, `pavucontrol` still sends multiple property notify events per second, but at least hlwm  ignores them instead of freezing..

blarz:
Interesting! Ok, I can't reproduce the behaviour of pavucontrol locally, but I don't see any regression with the patch and if it fixes your problem, let's go!

blarz (approved):
LGTM

t-wissmann:
Thanks for checking the changes :-)

I still need to understand mergify here, because it makes no sense what mergify does: The content of this PR was checked by the CI and now mergify runs the CI *again* before merging (which is the very same commit because the base branch was up to date)

PR #1642, opened 2026-01-01
INSTALL: clarify how to build without doc

### LINKED ISSUE #1636: [Compile error] ASCIIDOC_A2X-NOTFOUND: command not found
I just checked out the git sources and got it to compile to 87%, well, 88%:

    [ 87%] Generating hlwm-objects-gen.txt
    [ 88%] Built target gen_object_asciidoc [ 89%] Generating herbstclient.1
    /bin/sh: line 1:     ASCIIDOC_A2X-NOTFOUND: command not found

Not sure what this is. Probably some documentation? I actually don't need any documentation so perhaps that
could be the default; but either way, I think it should not fail on this environment variable. I am more interested
in trying out herbstluftwm, than in reading documentation; I read documentation online anyway, so the oldschool
man pages I don't usually read. (Herbstluftwm has a nifty tutorial, but now I am stuck because the tutorial is
only helpful if I can compile herbstluftwm.)

Anyway just reporting this here.

t-wissmann:
Make it more explicit how to build herbstluftwm without the documentation. This closes #1636.

### DIFF
--- a/.github/workflows/hlwm.yml
+++ b/.github/workflows/hlwm.yml
@@ -13,6 +13,8 @@ on:
     paths-ignore:
     - '.mergify.yml'
   pull_request:
+    branches:
+    - master
     paths-ignore:
     - '.mergify.yml'
 
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,5 +1,20 @@
+merge_queue:
+  max_parallel_checks: 1
+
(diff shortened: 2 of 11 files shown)


blarz (approved):
LGTM, making the cmake call explicit makes it clearer

PR #1643, opened 2026-01-01
Remove --locked from the documentation

t-wissmann:
The --locked option can cause a frozen state if the autostart does not run `unlock`. The advantages of --locked does not outweight the risks, so --locked is removed from the documentation. But it is still present for backwards compatibility and experienced users.

The original intention was to avoid flickers when a autostart gradually sets up the theme. With the new CSS-style themes, this is only one step, so the flickering is avoided in the long run anyway.


blarz (approved):
Good riddance to `--locked`.

PR #1645, opened 2026-01-01
Only update _NET_FRAME_EXTENTS if necessary

### LINKED ISSUE #1644: Only update _NET_FRAME_EXTENTS if necessary
The theme engine currently updates the `_NET_FRAME_EXTENTS` property quite often, even when the focus changes. To reproduce, run

```
xprop -spy
```
from a terminal and click an arbitrary window. Then change focus or resize (other) clients. Then, `xprop` repeatedly reports that the property is updated: 
```
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
_NET_FRAME_EXTENTS(CARDINAL) = 5, 5, 32, 5
...
```

t-wissmann:
Whenever decorations are redrawn, we possibly need to update the content of the
_NET_FRAME_EXTENTS X11 property. Formerly, the property was always updated
whenever some window on the monitor was redrawn. The present change remembers
the last value of the frame extents in the class Decoration and only updates
the X11 property if one of the values changes.

To reproduce the issue and the fix, run

    xprop -spy

from a terminal and select any window. It reports all property changes,
in particular those to _NET_FRAME_EXTENTS.

This closes #1644.

### DIFF
--- a/ci/diff-json-doc.py
+++ b/ci/diff-json-doc.py
@@ -3,6 +3,7 @@
 import re
 import subprocess
 import os
+import pathlib
 import sys
 import argparse
 import json
@@ -20,8 +21,8 @@ class GitDir:
         tmp_dir = self.dirpath
         full_cmd = [
             'git',
-            '--git-dir=' + os.path.join(tmp_dir, '.git'),
-            '--work-tree=' + tmp_dir
+            f'--git-dir={tmp_dir / ".git"}',
+            f'--work-tree={tmp_dir}',
         ] + list(cmd)
         print(':: ' + ' '.join(full_cmd), file=sys.stderr)
         return subprocess.run(full_cmd, check=check)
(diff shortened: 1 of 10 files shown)


blarz (approved):
Verified locally, no more unnecessary events are emitted!

Issue #1543, opened 2023-02-21
Client border not cleared when urgent client is focused


fabian-thomas:
In horizontal layout, when a client is set to urgent, it gets its urgent border color. If we now jump to this client, it is focused, so the border color should change to the active color. This does not happen for me. It stays with the urgent border color.

dexgs:
I am seeing this as well (git master). It seems to happen irrespective of layout.

PR #1649, opened 2026-01-09
Add missing comparison to operator== in DecorationParameters

gravndal:
Fixes window decorations not updating immediately outside tabbed max layout frames.

Closes https://github.com/herbstluftwm/herbstluftwm/issues/1543.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -42,13 +42,15 @@ passenv =
 
 [testenv:flake8]
 deps =
-    flake8 == 3.8.3
+    flake8 == 7.1.2
 commands = flake8 .
 
 [flake8]
 ignore =
-    E501, # line too long
-    W503, # incompatible with pep8.. "Line break occurred before a binary operator"
+    # line too long
+    E501,
+    # incompatible with pep8.. "Line break occurred before a binary operator"
+    W503,
 
 [pytest]
 markers =


dexgs:
Thanks. This change fixes the issue for me.

dexgs:
I took a look myself, and the root cause seems to be that the `urgent_` field of `DecorationParameters` is not compared in its `==` operator implementation: https://github.com/herbstluftwm/herbstluftwm/blob/master/src/decoration.h#L61-L69
```cpp
    bool operator==(const DecorationParameters& other) const {
        return focused_ == other.focused_
                && fullscreen_ == other.fullscreen_
                && pseudotiled_ == other.pseudotiled_
                && floating_ == other.floating_
                && minimal_ == other.minimal_
                && tabs_ == other.tabs_
                && urgentTabs_ == other.urgentTabs_;
    }
```
The result is that `setParameters` returns early here: https://github.com/herbstluftwm/herbstluftwm/blob/master/src/decoration.cpp#L171-L175

`Client::urgencyAttributeChanged` already calls `Client::redraw`, which calls `DecorationParameters::setParameters`. `recomputeStyle` just calls `setParameters` with `force = true`.

Changing the `==` operator implementation so it also checks `urgent_ == other.urgent_` fixes the root problem.

t-wissmann (approved):
Perfect!

PR #1663, opened 2026-07-27
systemd.service and sd_notify() integration

Flowdalic:
I am modernizing my herbstluftwm + gnome-flashback setup. Recent GNOME versions have overhauled session management. GNOME has deprecated its legacy internal session manager and now delegates session component initialization to `systemd --user`.

Because of this, herbstluftwm needs to be started a systemd user unit, that can be later selected by the gnome-flashback-herbstluftwm session (not part of this PR). Therefore, it seems sensible for herbstluftwm to ship a `herbstluftwm.service`. This is done in the first commit. The second commit adds `sd_notify()` integration, which is not strictly necessary, but nice to have. Feel free to pick neither, the first, or all commits in this PR.

### DIFF
--- a/share/herbstluftwm.service
+++ b/share/herbstluftwm.service
@@ -8,3 +8,4 @@ After=graphical-session-pre.target
 ExecStart=herbstluftwm
 ExecReload=herbstluftwm reload
 Restart=on-failure
+Type=notify
--- a/share/herbstluftwm.service
+++ b/share/herbstluftwm.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=A manual tiling window manager for X11
+Documentation=man:herbstluftwm(1)
+PartOf=graphical-session.target
+After=graphical-session-pre.target
+
+[Service]
+ExecStart=herbstluftwm
+ExecReload=herbstluftwm reload
+Restart=on-failure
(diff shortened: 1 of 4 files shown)


t-wissmann:
Hi Flo,
thanks! I was a bit reluctant to pull in new dependencies, so sorry for letting you wait! but I think it makes sense in the long run.

Pull Requests are squashed to one commit, so all the description of your changes should be moved to the pull request description (which will be used as the commit message). I haven't tested it yet and I'm not sure how to do this. (Is there a way to test this in the test suite?)

Can you also add an entry to the `NEWS` file?

share/herbstluftwm.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=A manual tiling window manager for X11
+Documentation=man:herbstluftwm(1)
+PartOf=graphical-session.target
+After=graphical-session-pre.target
+
+[Service]
+ExecStart=herbstluftwm
+ExecReload=herbstluftwm reload
```
t-wissmann, line 9: Do you mean `herbstclient reload`? But anyway, I'm not sure whether this is really what `systemd` understands as reloading, because `herbstclient reload` simply runs `~/.config/herbstluftwm/autostart`

Issue #1527, opened 2022-10-01
Placing float windows in the middle of its parent


kriansa:
Is it possible to create a rule where a selected few client windows should be placed in the center of their parent window instead of the center of the monitor? 

I'm having one of those first world problems that when using a super wide monitor, few popups that are placed in `center` are becoming annoying when their parent is on the side of the screen. I was thinking about playing with `floating_geometry` but I don't know how to get the coords of the parent window. Is it even possible?

Thanks!

t-wissmann:
Hi, I think it is only possible by implementing it directly in hlwm's source code (something like `floatplacement=parentcenter`)

kriansa:
I see. I'll try to take a stab at it this weekend and see how it goes. I was just reading the code and it doesn't seem complicated, although I'm not experienced with X or Xlib. I was looking at `client.cpp` class and I couldn't find a way to easily get the "parent window" (if that concept even exists), would you help me pointing out where should I look to get that?

t-wissmann:
Implementing it is not so easy as one may think. You need to add a `Client* transientFor = nullptr;` attribute in `client.h`. Then, in `Client::init_from_X()`, you need to read the [WM_TRANSIENT_FOR property](https://tronche.com/gui/x/icccm/sec-4.html) and save the result in `transientFor`. Then, in the implementation of `parentcenter`, one can place it in the center of `transientFor->last_size_`. (the above description is just a sketch)

PR #1670, opened 2026-09-07
Keep transient windows above the window they are transient for

SebastienGllmt:
This PR makes hlwm enforce the EWMH stacking rule for transient windows at every place where the stacking order of a window changes

Sorry this PR description is a bit long (even though the code itself is not that long) because although the code change isn't that big, it impacts end-user behavior in a few different ways I wanted to lay out so the impact of these changes is more clear (and it doesn't really make sense to have these as separate PR, since you need all of them bundled to have coherent behavior)

## Bug

For a window with a `WM_TRANSIENT_FOR` hint, it's possible for the main window to permanently hide its own dialog (which is doubly bad in the case the modal dialog disables the parent window until the dialog is handled)

More specifically, this happens when the parent window is:
1. the target of a `raise` command (`raise_on_click`, `raise_on_focus`, or re-raising caused by a re-layout)
2. moved into the fullscreen layer (`fullscreen` on)
3. moved into the focus layer (`raise_on_focus_temporarily`, or the focused client whenever a fullscreen client exists on the tag)
5. moved into the floating layer or back into the tiling layer (`floating` attribute set or cleared)

## Background

Recall:
1. `Client::raise()` is `Stack::raiseSlice` of the client's own slice.
2. `Client::lower()` is `Stack::lowerSlice` of the client's own slice
3. `Monitor::applyLayout` moves only the client's own slice into the fullscreen / focus layer on every re-layout 
5. `HSTag::applyClientState` (the `floating` attribute) moves only the client's own slice into the floating or the tiling layer

The `WM_TRANSIENT_FOR` hint is read once (and only once) in `ClientManager::applyDefaultRules` (since #725) to float the window, and then never checked again

## Why the window manager

There are two different sources that force this to be a window manager concern:
1. X's EWMH says "Windows that are transient for another window should be kept above this window." ([source](https://specifications.freedesktop.org/wm/latest-single/#STACKINGORDER))
2. [ICCCM](https://tronche.com/gui/x/icccm/sec-4.html) 4.1.2.6: a transient window is "a pop-up on behalf of" the named window, so a dialog behind that window has failed at its purpose (and a modal one leaves the application unusable)

The base for this PR follows the sketch in #1527, except that the `Window` is stored rather than a `Client*` (so that a parent that is managed later, or unmanaged first, needs no bookkeeping)

## Fix

This fix is split into multiple commits (each containing tests):

#### [b2cecbd](https://github.com/herbstluftwm/herbstluftwm/commit/b2cecbdbacc15fa3f497f463f44822b87e71fec6): fix fullscreened float handling

Although strictly speaking a separate issue, this PR also makes a that a floating window that is fullscreen is now stacked in the fullscreen layer (like a tiled one) both because
1. it's a related type of issue
2. it's needed for the proper fix of the main issue to work (since dialogs are auto-floated by #725 and so would otherwise be invisible to that loop)

More specifically, we now compute the layer over all clients of the tag (`HSTag::foreachClient`)

#### [52e53f8](https://github.com/herbstluftwm/herbstluftwm/commit/52e53f84716f5939262df3bf36ebae627fadf6b1): fix `raise` handling

raising a window now also raises its dialogs, so a raise (click, focus, re-layout) can no longer put a window on top of its own dialog

Concretely,
1. `Client` remembers the hint (`transientFor_`, read at manage time and on `PropertyNotify` of `WM_TRANSIENT_FOR`, including its deletion)
3. `Client::raise()` has the new behavior:
	1. raises its own slice
	2. raises every client on the same tag whose hint names it, recursively (lowest transient first so the transients keep their relative order)

Note:
1. Transients on another tag are left alone.
2. Every existing raise path goes through `Client::raise()` (so fortunately no need to change callers)

#### [bf2f3ee](https://github.com/herbstluftwm/herbstluftwm/commit/bf2f3eef234876567fb93996ab22f14ee6b450c7): : fix fullscreen and focus layer handling

Apply the same fix in two other places where hlwm reorders the stack:
1. a window entering the fullscreen layer
2. a window entering the focus layer

Concretely,
1. Make the new `raise()` behavior also changed with `Monitor::applyLayout` (by moving it into `Client::withTransients()`)
2. Make `Monitor::applyLayout` fills the fullscreen layer and the focus layer in that order
3. A client entering the fullscreen layer below a transient that is already there (a fullscreen dialog) re-inserts the transient above itself

#### [eb806e2](https://github.com/herbstluftwm/herbstluftwm/commit/eb806e2e): fix `floating` attribute handling

Apply the same fix in one other places where hlwm reorders the stack: setting floating on a parent inserts it on top of the floating layer, above its already-floating dialog (and clearing it does the same in the tiling layer)

Fix: a window that is set floating (or tiled again) now enters the new layer below its dialogs that are already there

Concretely,
1. `Client::raiseIntoLayer()` puts a client on top of a layer with its transients above it (a transient already in the layer is re-inserted above the client, a transient not in the layer is brought along only when asked for)
2. The fullscreen and focus layers bring the transients along; the floating and tiling layers don't (a tiled dialog of a floated window stays tiled)

#### [867ec99](https://github.com/herbstluftwm/herbstluftwm/commit/867ec996807b0c14802867a105d4c5599bf98970): fix `lower` handling

lowering a dialog now also lowers the window it belongs to, so a `lower` can no longer put a dialog below its own window

Concretely,
1. `Client::lower()` lowers its own slice, then the client it is transient for (recursively up the `WM_TRANSIENT_FOR` chain on the same tag), each to the bottom
2. lowering a window that has dialogs needs no extra move (its dialogs are above it already, and stay there)

### Which windows to raise

Note: when it comes to raising parents vs dialogs, EWMH constrains one case only ("Windows that are transient for another window should be kept above this window", [source](https://specifications.freedesktop.org/wm/latest-single/#STACKINGORDER)), and raising a dialog alone never violates it.

However, bringing the parent up too is a policy WMs disagree on:
1. mutter always raises the root ancestor ([`meta_window_raise`](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/core/window.c), reason given: "it's weird to see windows from other apps stacked between a child and parent window of the currently active app")
2. openbox only for modal dialogs ([`restack_windows`](https://github.com/danakj/openbox/blob/master/openbox/stacking.c))

hlwm
1. has no application-group notion
2. does not track `_NET_WM_STATE_MODAL`
4. has `raise` as a command on one named client (also behind `raise_on_click`/`raise_on_focus`)

therefore, making hlwm move a second, unnamed window would be a new policy rather than a bug fix (so I left it out of this PR. Can be handle separately if people want this behavior)

### Handling of `lower`

`lower` on a dialog is the mirror image of `raise` on its window: the named window moves as asked, and the invariant then requires the window it is transient for to end up below it

Options considered:
1. lower the dialog alone (behavior before this PR): the dialog sits below its own window until that window is next raised. This is exactly the state EWMH says to avoid and the state the rest of this PR removes, so exempting `lower` would be inconsistent (and "the user named that window" is the same argument that doesn't hold for `raise` on the parent)
2. lower the dialog, then the window it is transient for (recursively): what openbox does ([`restack_windows`](https://github.com/danakj/openbox/blob/master/openbox/stacking.c), "lower our parents after us, so they go below us"). Needs only the existing `Stack::lowerSlice`
3. stop the dialog directly above its window: what mutter does (its stacking constraints keep a transient above its parent at any position, so a lowered dialog stays just above its parent, [`stack.c`](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/core/stack.c)). Needs positional insertion in `Stack`, and makes `lower` a no-op whenever the parent is already on top

I picked option 2 as
1. it felt the more coherent with what this PR is trying to address: most applications can't be used while a dialog is up, so lowering just the dialog is almost always just getting the user into a bad/confusing state
2. it fits hlwm's `Stack`: a slice is raised to the top or lowered to the bottom of its layers, and relations are kept by moving a group in order (as `HSTag::onGlobalFloatingChange` already does, and as the `raise` side of this PR does)

Note: lowering a window that has dialogs is unchanged: the dialogs are above it already, so nothing else needs to move

### Invariant enforcement location in the code

The relation is enforced at the places where the stack order changes rather than modelled inside `Stack`:
1. `Stack` orders `Slice`s and knows nothing about clients (a `Slice` is also a frame or an unmanaged window); `WM_TRANSIENT_FOR` is a client property, so `Stack` would need a callback into the client manager on every reorder
2. those places are few and known: `raiseSlice` is only called from `Client::raise`, and `sliceAddLayer` from `Monitor::applyLayout` (fullscreen, focus) and `HSTag` (the `floating` attribute, both directions). The remaining `sliceAddLayer` callers keep the invariant already: the tag-wide `floating` toggle inserts at the bottom in stack order, and a newly managed client has no transients yet
3. each of those places has the client at hand, so the fix is a call to `withTransients()`/`raiseIntoLayer()` there, instead of a new invariant inside `Stack` that every existing caller would have to be audited against anyway

## Repro on 0.9.6 / master

```
herbstclient floating on
# open a window P, a dialog D transient for P (e.g. any GTK app's message box), and a third window O
herbstclient stack        # Floating-Layer: O, D, P (top to bottom)
herbstclient raise $P
herbstclient stack        # Floating-Layer: P, O, D on master — D, P, O with the fix

# the layer moves (third commit):
herbstclient jumpto $O
herbstclient set_attr clients.$P.fullscreen true
herbstclient stack        # Fullscreen-Layer: P alone on master (D stays below in the Floating-Layer) — D, P with the fix

# lower (fifth commit), starting again from O, D, P:
herbstclient lower $D
herbstclient stack        # Floating-Layer: O, P, D on master (D under its own window) — O, D, P with the fix
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -18,9 +18,11 @@ Current git version
   * Keep transient windows (dialogs with a WM_TRANSIENT_FOR hint) above the
     window they belong to: raising a client now also raises its transient
     clients, and a client in the fullscreen layer or in the focus layer
-    (raise_on_focus_temporarily) takes its transient clients with it, so a
-    dialog is no longer covered by its own application after the application
-    window is raised, focused, clicked or made fullscreen.
+    (raise_on_focus_temporarily) takes its transient clients with it, and a
+    client that is set floating (or tiled again) enters the new layer below
+    its transients that are there, so a dialog is no longer covered by its
+    own application after the application window is raised, focused,
+    clicked, made fullscreen or set floating.
 
 Release 0.9.6 on 2026-04-03
 ---------------------------
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,10 @@ Current git version
     unmapped (e.g. minimized) client changes. Previously, the next unmap by
     the application itself (e.g. hiding to the system tray) was ignored and
     the window stayed managed although it was not shown anymore.
+  * Put a floating client (the 'floating' attribute set) that is fullscreen
+    in the fullscreen layer, like a tiled fullscreen client. Previously it
+    was resized to the monitor but stayed in the floating layer, below any
+    floating window raised after it.
 
 Release 0.9.6 on 2026-04-03
 ---------------------------
--- a/NEWS
+++ b/NEWS
@@ -17,8 +17,10 @@ Current git version
     floating window raised after it.
   * Keep transient windows (dialogs with a WM_TRANSIENT_FOR hint) above the
     window they belong to: raising a client now also raises its transient
-    clients, so a dialog is no longer covered by its own application after
-    the application window is raised, focused or clicked.
+    clients, and a client in the fullscreen layer or in the focus layer
+    (raise_on_focus_temporarily) takes its transient clients with it, so a
+    dialog is no longer covered by its own application after the application
+    window is raised, focused, clicked or made fullscreen.
 
 Release 0.9.6 on 2026-04-03
 ---------------------------
--- a/NEWS
+++ b/NEWS
@@ -20,9 +20,11 @@ Current git version
     clients, and a client in the fullscreen layer or in the focus layer
     (raise_on_focus_temporarily) takes its transient clients with it, and a
     client that is set floating (or tiled again) enters the new layer below
-    its transients that are there, so a dialog is no longer covered by its
+    its transients that are there, and lowering a client also lowers the
+    client it is transient for, so a dialog is no longer covered by its
     own application after the application window is raised, focused,
-    clicked, made fullscreen or set floating.
+    clicked, made fullscreen or set floating, or after the dialog is
+    lowered.
 
 Release 0.9.6 on 2026-04-03
 ---------------------------
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,10 @@ Current git version
     in the fullscreen layer, like a tiled fullscreen client. Previously it
     was resized to the monitor but stayed in the floating layer, below any
     floating window raised after it.
+  * Keep transient windows (dialogs with a WM_TRANSIENT_FOR hint) above the
+    window they belong to: raising a client now also raises its transient
+    clients, so a dialog is no longer covered by its own application after
+    the application window is raised, focused or clicked.
 
 Release 0.9.6 on 2026-04-03
 ---------------------------
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -335,6 +335,26 @@ void Client::collectWithTransients(vector<Client*>& result, std::set<Client*>& v
 }
 
 /**
+ * @brief put this client on top of the given layer of its tag's stack, and
+ * its transient windows above it. A transient that is already in the layer
+ * is re-inserted above the client; a transient that is not in the layer is
+ * inserted only if bringTransients is set (e.g. the fullscreen layer takes
+ * the dialogs of a fullscreen window with it, whereas the floating layer
+ * leaves a tiled dialog of a floated window where it is).
+ */
+void Client::raiseIntoLayer(HSLayer layer, bool bringTransients) {
+    for (Client* client : withTransients()) {
+        bool inLayer = client->slice->layers.count(layer) != 0;
+        if (client == this || inLayer || bringTransients) {
+            if (inLayer) {
+                tag()->stack->sliceRemoveLayer(client->slice, layer);
+            }
+            tag()->stack->sliceAddLayer(client->slice, layer);
+        }
+    }
+}
+
+/**
  * @brief raise this client and keep its transient windows above it: after
  * the client itself, every client on the same tag whose WM_TRANSIENT_FOR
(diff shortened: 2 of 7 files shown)


t-wissmann:
Hey, thanks for your work! It may take a bit of time for me to read through it. A few requests upfront:
* Is it possible to split the PR into smaller pieces? E.g. by first introducing the `transientFor_` member and maintaining it?
* `    Window      transientFor_ = None;`, this should rather be a `Link_<Client>`. You need to give this link a name (in the `Client` constructor init-list: `  , transientFor_(*this, "transient_for")` ). Then, you can use the `transient_for` attribute in the test cases).
* With the `Link_` (or a pointer in general), one needs to reset it to `NULL` whenever the respective client gets unmanaged. The respective function is `void Client::forgetOtherClient(Client* otherClient)`, so it needs to be extended by something like `if (transientFor_() == otherClient) { transientFor_ = NULL; }`. (`Link_` has  suitable overloaded `=`)

PR #1671, opened 2026-09-21
Fix out-of-bounds completion argument access

SebastienGllmt:
Completion can theoretically be used to find out whether a command accepts another argument

In this case, `herbstclient complete 3 merge_tag` would ideally return `HERBST_NO_PARAMETER_EXPECTED`, but instead it segfaults due to an out-of-bound error

This PR fixes this issue


t-wissmann (approved):
Makes perfect sense!

Thanks for fixing it!

